PR #71, opened 2025-11-24
 Makefile: add golangci-lint version check:

huhong-web:
If the golangci-lint you are using was built with a Go version lower than the current project's code, golangci-lint will be stuck and use huge memory.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,7 +20,7 @@ GO_BUILD_STATIC_WITH_VERSION := $(GO_BUILD_STATIC) \
 all: gen-deps gen sync build
 
 gen-deps:
-	# maybe need to install libbpf-devel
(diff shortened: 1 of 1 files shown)


huhong-web:
已收到

Makefile
```diff
@@ -55,6 +56,13 @@ fmt-rewrite-rules:
 	gofmt -w -r 'interface{} -> any' $(shell find . -type f -name '*.go' -not -path "./vendor/*")
 
 golangci-lint:
+	@command -v golangci-lint >/dev/null 2>&1 || \
+	{ echo "golangci-lint not found, go install github.com/golangci/golangci-lint/cmd/golangci-lint@1.62.2"; }
+	@lint_go_version=$$(golangci-lint --version | grep "built with go" | awk '{print $$7}'); \
```
hao022: why not
golangci-lint version --format short

PR #113, opened 2026-01-19
fix(utils/executil): fix issue110 get uncorrect container name 

huhong-web:
fix issue110 where container name was resolved as the host hostname instead of the container's own hostname.


huhong-web:
已经修改

PR #183, opened 2026-05-26
fix: trim spaces from KubeletClientCertPath parts and include err in panic

huhong-web:
Splitting on ',' without trimming caused a leading-space path when the config value used ', ' as separator, breaking cert-based kubelet client creation. Also surface the underlying error in the panic message so failures are easier to diagnose.


huhong-web:
已收到

PR #232, opened 2026-06-16
fix: guard unsafe type assertions to prevent runtime panics

Redamency:
## Summary

Two unchecked type assertions in the codebase could trigger a runtime panic if the underlying value type ever changes (e.g. due to refactoring of dependencies or unexpected cache contents). Replace them with the comma-ok pattern.

### 1. `internal/request/rawhttp.go` - `connectUnix`

`net.Dial(\"unix\", ...)` typically returns a `*net.UnixConn`, but the code asserts this without checking. If the assertion ever fails, the program panics and the dialed connection leaks.

**Fix**: verify the assertion, close the connection on mismatch, and return a descriptive error.

### 2. `internal/storage/localfile/localfile.go` - `newFileWriter`

The value loaded from `sync.Map` is asserted to `io.Writer` without checking. Although today the only writer added is a `*filerotate.FileRotator` (which implements `io.Writer`), an unchecked assertion is fragile.

**Fix**: use the comma-ok pattern and return `nil` on mismatch.

## Why this matters

Both sites are on hot paths: `connectUnix` is called for every security-agent request, and `newFileWriter` is called for every record written to local file storage. A panic here would crash the whole process. The fix is small and defensive.

## Testing

No behavioral change for the success path. Builds cleanly.

### DIFF
--- a/internal/request/rawhttp.go
+++ b/internal/request/rawhttp.go
@@ -18,6 +18,7 @@ import (
 	"bufio"
 	"bytes"
 	"encoding/json"
+	"fmt"
 	"io"
 	"net"
 	"net/http"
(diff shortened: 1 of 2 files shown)


Redamency:
Closing this PR — on review, both changes are over-defensive and one (localfile) could mask issues by returning nil where the caller would dereference it. Will open a more focused PR addressing a real bug.

PR #246, opened 2026-06-18
fix(iotracing): skip reset diskstat samples

pkurunner:
Diskstats counters can go backwards after a device reset/recreate. My first patch clamped each field separately, but that can still mix two counter epochs in one sample.

This version treats any backwards field as a reset and drops that interval by returning zero metrics. The next sample is then compared against the new counter epoch, so we avoid both uint64 underflow and misleading partial deltas.

Testing: gofmt -w core/autotracing/iotracing.go

Not run locally: GOOS=linux GOARCH=amd64 go test -c ./core/autotracing is blocked by missing generated capnp files in my local snapshot.


pkurunner:
Checked the failing VM jobs after updating this. ubuntu20 stops in `test_dropwatch_ratelimit.sh`; ubuntu24 builds `iotracing` and passes `test_iotracing.sh`, then fails the final log check on `cpusys: signal: killed` during shutdown.

I do not have permission to rerun failed jobs from the fork, but the current failures do not point to this diskstats change.

PR #249, opened 2026-06-18
fix(events): reserve watch clients atomically

pkurunner:
Two `/v1/events/watch` requests can both pass the active-client check before either increments the counter, so `max_clients` is not a hard cap under concurrency.

This reserves the slot with CAS and releases the same counter on exit. I added a small concurrent test around that path.

Testing:
- `gofmt -w cmd/huatuo-bamai/handlers/events.go cmd/huatuo-bamai/handlers/events_test.go`
- Not run: `go test ./cmd/huatuo-bamai/handlers` on Windows is blocked by Linux-only `github.com/prometheus/procfs/sysfs` build constraints.
- Not run: `GOOS=linux GOARCH=amd64 go test -c ./cmd/huatuo-bamai/handlers` is blocked locally by missing generated `internal/toolstream/transport` capnp types on main.

### DIFF
--- a/cmd/huatuo-bamai/handlers/events.go
+++ b/cmd/huatuo-bamai/handlers/events.go
@@ -66,6 +66,22 @@ func NewEventsHandler(maxClients, keepAliveIntervalSecs int) *EventsHandler {
 	return h
 }
 
+func (h *EventsHandler) acquireClient() bool {
+	for {
+		active := h.activeClients.Load()
+		if int(active) >= h.maxClients {
+			return false
+		}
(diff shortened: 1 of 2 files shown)


pkurunner:
CI note: Build/Lint, PR name lint, and Ubuntu 22.04 VM passed. Ubuntu 20.04 failed in `integration/test_dropwatch_ratelimit.sh` with `events=0` and `rate-limit warnings=0`; that path is unrelated to the `/v1/events/watch` client counter change here.

PR #252, opened 2026-06-20
fix(flamegraph): use root name index

pkurunner:
`LevelsToTree` was reading the root label from `Values[0]`, but that slot is the start offset. Child nodes already read names from `NameOffest`; the root should do the same, otherwise a root whose name index is not zero gets labeled incorrectly.

Testing:
- `gofmt -w internal/flamegraph/flamegraph.go internal/flamegraph/flamegraph_test.go`
- `GOCACHE=.gocache-flamegraph go test ./internal/flamegraph`


pkurunner:
CI note: PR name lint and Build/Lint passed. Ubuntu 22.04 VM failed before tests while installing `capnproto`: `apt-get` could not acquire `/var/lib/dpkg/lock-frontend` because `unattended-upgr` was holding it, so this looks unrelated to the flamegraph change.

PR #339, opened 2026-07-07
fix(pod): prevent panics from bare type assertions and nil type validation

wc4440222:
## What does this PR do?

Fixes two panic-causing bare type assertions in `internal/pod/`.

### 1. `LabelHostNamespace()` — bare type assertion panic

**Problem:** `Container.LabelHostNamespace()` uses a bare type assertion `c.Labels[labelHostNamespace].(string)`. If the label is missing, `nil`, or not a string, this causes a runtime panic. This function is called in:
- `pkg/metric/data.go` — metric label generation
- `pkg/tracing/document_writer.go` — tracing document generation
- `internal/matcher/container_matcher.go` — container field extraction

A container without the `HostNamespace` label would crash the entire metric/tracing pipeline.

**Fix:** Replace with comma-ok pattern. Returns empty string on nil Container, nil Labels map, missing key, or wrong type.

### 2. `RegisterContainerLifeResources()` — Elem() panic on non-pointer

**Problem:** The validation logic `anythingType.Kind() != reflect.Pointer && anythingType.Elem().Kind() != reflect.Struct` has a short-circuit evaluation flaw. If `anythingType` is a non-pointer type (e.g., `reflect.TypeOf(42)`), the first condition is true but `anythingType.Elem()` panics because non-pointer types don't have an `Elem()` method. Passing `nil` also panics.

**Fix:** Split into three separate checks:
1. Check for nil first
2. Check `Kind() != reflect.Pointer` 
3. Only then check `Elem().Kind() != reflect.Struct`

### Testing

Added `container_safety_test.go` with 8 test cases:
- `TestLabelHostNamespace_MissingLabel` — label key not present
- `TestLabelHostNamespace_NilLabels` — Labels map is nil
- `TestLabelHostNamespace_NilReceiver` — Container is nil
- `TestLabelHostNamespace_WrongType` — label value is int, not string
- `TestLabelHostNamespace_ValidLabel` — normal case still works
- `TestRegisterContainerLifeResources_NilType` — nil type returns error
- `TestRegisterContainerLifeResources_NonPointerType` — int type returns error
- `TestRegisterContainerLifeResources_PointerToNonStruct` — *int returns error

```
$ go test -count=1 -v ./internal/pod/ -run 'TestLabelHostNamespace|TestRegisterContainerLifeResources'
=== RUN   TestLabelHostNamespace_MissingLabel
--- PASS: TestLabelHostNamespace_MissingLabel (0.00s)
=== RUN   TestLabelHostNamespace_NilLabels
--- PASS: TestLabelHostNamespace_NilLabels (0.00s)
=== RUN   TestLabelHostNamespace_NilReceiver
--- PASS: TestLabelHostNamespace_NilReceiver (0.00s)
=== RUN   TestLabelHostNamespace_WrongType
--- PASS: TestLabelHostNamespace_WrongType (0.00s)
=== RUN   TestLabelHostNamespace_ValidLabel
--- PASS: TestLabelHostNamespace_ValidLabel (0.00s)
=== RUN   TestRegisterContainerLifeResources_NilType
--- PASS: TestRegisterContainerLifeResources_NilType (0.00s)
=== RUN   TestRegisterContainerLifeResources_NonPointerType
--- PASS: TestRegisterContainerLifeResources_NonPointerType (0.00s)
=== RUN   TestRegisterContainerLifeResources_PointerToNonStruct
--- PASS: TestRegisterContainerLifeResources_PointerToNonStruct (0.00s)
PASS
```

## Checklist

- [x] Code follows the project's style guidelines (gofumpt + gofmt)
- [x] Self-reviewed the code
- [x] Added unit tests for all fix scenarios
- [x] No breaking API changes (returns empty string instead of panic)
- [x] DCO signed (`Signed-off-by`)

Signed-off-by: wc4440222 <199748891+wc4440222@users.noreply.github.com>

### DIFF
--- a/internal/pod/container.go
+++ b/internal/pod/container.go
@@ -17,12 +17,11 @@ package pod
 import (
 	"errors"
 	"fmt"
+	"huatuo-bamai/internal/log"
 	"regexp"
 	"sync"
 	"syscall"
 	"time"
-
-	"huatuo-bamai/internal/log"
 )
 
 // containerIDRegexp matches a 12-64 character hex container ID.
@@ -66,9 +65,18 @@ func (c *Container) LifeResources(key string) any {
 	return c.lifeResources[key]
 }
 
-// LabelHostNamespace returns namespace label
+// LabelHostNamespace returns the HostNamespace label.
+// Returns empty string if the label is missing or not a string,
+// preventing a runtime panic from a bare type assertion.
 func (c *Container) LabelHostNamespace() string {
-	return c.Labels[labelHostNamespace].(string)
+	if c == nil || c.Labels == nil {
+		return ""
+	}
+	v, ok := c.Labels[labelHostNamespace].(string)
+	if !ok {
+		return ""
+	}
+	return v
 }
 
 func (c *Container) InitPidOrInitnsPid() int {
--- a/internal/pod/container_life_resources.go
+++ b/internal/pod/container_life_resources.go
@@ -32,7 +32,13 @@ var lifeResourcesTpl sync.Map
 //	data := map[string]int{"acct": 0, "usage": 0}
 //	RegisterContainerLifeResources("cpu", reflect.TypeOf(data))
 func RegisterContainerLifeResources(key string, anythingType reflect.Type) error {
-	if anythingType.Kind() != reflect.Pointer && anythingType.Elem().Kind() != reflect.Struct {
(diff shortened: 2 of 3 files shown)


hao022:
@wc4440222 这些都是内部函数，不需要强校验，现在没有遇到为nil的情况。

PR #344, opened 2026-07-08
fix(tracing): protect sync.Map bare type assertions and replace AllocTaskID panic

wc4440222:
## Problem

`pkg/tracing/task.go` has two categories of safety issues:

### 1. Panic in `AllocTaskID()`
```go
num, err := rand.Int(rand.Reader, charsetLength)
if err != nil {
    panic("Failed to generate random number")
}
```
`crypto/rand` errors are extremely rare (system entropy starvation), but panicking crashes the entire agent process. This should return an error instead.

### 2. Bare type assertions on `sync.Map` values
Five functions do `value.(*task)` without the `, ok` safety check:
- `tasksGarbageCollect` — line ~105
- `RunningTaskCount` — line ~221
- `ListTasks` — line ~232
- `Result` — line ~263
- `StopTask` — line ~281

If corrupt data somehow enters `taskLifeTmpCache`, these assertions will panic at runtime, crashing the agent. Defensive `ok` checks are essential for long-running production daemons.

## Fix

1. **`AllocTaskID()`**: Changed signature from `string` to `(string, error)`, replacing `panic` with `return "", fmt.Errorf(...)`.
2. **All `sync.Map` assertions**: Added `, ok` checks with appropriate error handling (log+skip for iteration, `StatusNotExist` error for `Result`, `fmt.Errorf` for `StopTask`).
3. **Updated callers**:
   - `NewTask()` in `task.go`: handles the error, logs it, returns empty string
   - `NewPipeline()` in `internal/profiler/aggregator/pipeline.go`: handles the error with an inline func
   - `TestAllocTaskID` in `task_test.go`: updated for new signature

## Testing

- All existing tests pass (except one pre-existing flaky timeout test unrelated to our changes)
- `go vet ./pkg/tracing/` passes
- `make import-fmt` applied

### DIFF
--- a/internal/pod/container.go
+++ b/internal/pod/container.go
@@ -17,11 +17,12 @@ package pod
 import (
 	"errors"
 	"fmt"
-	"huatuo-bamai/internal/log"
 	"regexp"
 	"sync"
 	"syscall"
 	"time"
+
+	"huatuo-bamai/internal/log"
 )
 
 // containerIDRegexp matches a 12-64 character hex container ID.
@@ -32,9 +33,9 @@ var (
 	containers = map[string]*Container{}
 
 	// updated
-	lastUpdatedAt = time.Now()
-	updatedStep   = 5 * time.Second
-	containersMu  sync.RWMutex
+	lastUpdatedAt     = time.Now()
+	updatedStep       = 5 * time.Second
+	containersMapLock sync.RWMutex
 )
 
 // Container object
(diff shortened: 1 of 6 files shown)


hao022:
@wc4440222 修复了一些问题和冲突，代码已经合入：

https://github.com/ccfos/huatuo/commit/e347a544fdd88544a4f1e8de2114af4b855112d2
https://github.com/ccfos/huatuo/commit/11f76845490655a31e43f61bce2d9223cca80a44

PR #343, opened 2026-07-08
feat(net_tx_latency): add TCP send-path latency tracing

123123213weqw:
## What

Adds `net_tx_latency`, the send-path counterpart to the existing `net_rx_latency` tracer — TCP **TX (send-path) latency** tracing.

## Why

huatuo already traces latency on the receive path (`net_rx_latency`), but the send path was a blind spot. This fills that gap so we can tell whether a slow TCP send is stalled in the stack (`tcp_sendmsg → qdisc`) or in the qdisc/driver/NIC (`net_dev_queue → net_dev_xmit`).

Refs ccfos/huatuo#323.

## How

Splits the TX path into two stages, each with its own threshold; an event fires only when a stage exceeds its threshold:

| stage | segment | default threshold |
|---|---|---|
| `TX_STAGE_SENDMSG` | `tcp_sendmsg → net_dev_queue` | 50 ms |
| `TX_STAGE_NIC` | `net_dev_queue → net_dev_xmit` | 1 ms |

Unlike RX (which diffs `skb->tstamp` set on receive), TX has no pre-existing timestamp, so the BPF program anchors timestamps itself and correlates the stages with two LRU hash maps:

- `tx_sock_start` (key = `sock *`) — captured at `kprobe/tcp_sendmsg`
- `tx_skb_start` (key = `sk_buff *`) — captured at `tracepoint/net/net_dev_queue`

The sender's `pid`/`comm` are carried sock→skb so NIC-stage events (which run in softirq context) still report the originating process. The `perf_event_t` layout intentionally mirrors `net_rx_latency`, so userspace reuses `NetTracingData` for storage and Grafana consistency.

## Changes

- **new** `bpf/net_tx_latency.c` — eBPF: 3 probes (`tcp_sendmsg`, `net_dev_queue`, `net_dev_xmit`), 2 hash maps, threshold-gated event submission.
- **new** `core/events/net_tx_latency.go` — userspace decode + filtering (host-netns exclusion, QoS blacklist); reuses `NetTracingData`.
- **mod** `core/events/config.go` — `NetTxLatency` config struct (`Sendmsg2Qdisc`, `Qdisc2Nic`, `ExcludedHostNetnamespace`, `ExcludedContainerQos`).
- **mod** `huatuo-bamai.conf` — `[EventTracing.NetTxLatency]` section + docs.
- **mod** `build/docker/grafana/dashboards/autotracing-event.json` — `net_tx_latency` stat panel (id 200).
[...]

PR #346, opened 2026-07-09
feat(net_rx_latency): support IPv6 alongside IPv4

### LINKED ISSUE #324: [功能建议]: 收包延迟检测支持 IPv6（兼容 IPv4）
## 背景

当前 huatuo 的网络收包延迟检测仅支持 IPv4，越来越多的生产环境已启用 IPv6，需要同步支持。

## 需求描述

扩展收包延迟检测功能以支持 IPv6 协议：

1. **perf_event_t 结构体扩展**：增加 IPv6 地址字段（128 bit）及地址族标记（AF_INET / AF_INET6）
2. **BPF 程序增强**：增加 IPv6 头部解析分支，正确提取源/目的 IPv6 地址
3. **用户态 Go 代码适配**：同步解析和处理 IPv6 地址信息

## 技术要求

- 确保现有 IPv4 功能不受影响，保持完全向后兼容
- IPv4 和 IPv6 共用同一事件流水线，通过地址族标记区分

## 验收标准

- [ ] IPv6 收包延迟可正常采集
- [ ] 现有 IPv4 功能不受影响
- [ ] 事件结构中包含正确的地址族信息

## 意向参与贡献

- [ ] 我有意向参与具体功能的开发实现并将代码贡献回到上游社区

123123213weqw:
## What

Extends `net_rx_latency` (TCP **receive**-path latency) to support **IPv6** alongside IPv4, with full backward compatibility.

Closes ccfos/huatuo#324.

> **Stacked on #343.** This branch is based on `feat/net-tx-latency` (#343) because the RX/TX event struct (`perf_event_t` / `netRcvPerfEvent`) is **shared** (TX aliases it). The diff therefore includes #343's commit. Merge #343 first and this PR collapses to the IPv6-only delta; or review commits `fc29724` (this PR) on top of `f2cde54` (#343).

## Why

huatuo's RX latency tracer was IPv4-only (gated on `ETH_P_IP`/`AF_INET`, addresses stored as `u32`). Production increasingly runs IPv6, so RX latency for v6 traffic was a blind spot.

## How

**One event struct, family-tagged.** `perf_event_t` gains an `addr_family` byte (`AF_INET`/`AF_INET6`) and the two address fields grow `u32 → 16 bytes` (IPv4 = low 4 bytes in network order; IPv6 = all 16). Userspace formats per family. JSON output keeps the existing `tcp_saddr`/`tcp_daddr` **string** fields (now IPv4 *or* IPv6) and adds one new `address_family` → backward compatible for storage/Grafana/IPv4 tests.

**Conditional IPv6 TCP-layer probe.** IPv4's TCP stage hooks `kprobe/tcp_v4_rcv`; the IPv6 analog is `kprobe/tcp_v6_rcv`, which only exists with `CONFIG_IPV6`. Since `Attach()` is all-or-nothing, attaching it unconditionally would fail the whole RX tracer on IPv4-only kernels. Mirror `core/metrics/iolatency_tracing.go`: switch to `EventPipeByName` + `AttachWithOptions`, appending `tcp_v6_rcv` only when `bpf.HasKprobeFunction("tcp_v6_rcv")`. The two tracepoint stages (`netif_receive_skb`, `skb_copy_datagram_iovec`) are protocol-agnostic, so IPv6 latency is collected on every kernel (2 of 3 stages minimum; all 3 when `tcp_v6_rcv` exists).

## Changes

- **`bpf/net_rx_latency.c`** — new `perf_event_t` layout; `skb_is_ipv4_tcp()` → `skb_tcp_family()` (returns family); `submit_rxlat_event()` branches on family to parse `iphdr`/`ipv6hdr`; new `kprobe/tcp_v6_rcv` probe; `tcp_v4_rcv` passes `AF_INET`.
- **`bpf/net_tx_latency.c`** — struct synced to the identical new layout; `submit_txlat_event()` sets `addr_family=AF_INET` and writes the 4-byte IPv4 addr into the 16-byte field. TX stays IPv4-only (IPv6 *parsing* for TX is a separate follow-up).
- **`core/events/net_rx_latency.go`** — `netRcvPerfEvent` new layout; `NetTracingData.AddressFamily`; shared `addrs()` helper (used by RX **and** TX) + AF consts; decode loop uses it; attach switched to `EventPipeByName` + conditional `AttachWithOptions`.
- **`core/events/net_tx_latency.go`** — decode uses `pd.addrs()`, sets `AddressFamily`.
- **`internal/utils/netutil/netutil.go`** — adds `Inetv6Ntop`.
- **`integration/test_net_rx_latency_ipv6.sh`** + **`integration/testdata/test_net_rx_latency_ipv6_user.c`** (new) — IPv6 integration test (ULA veth addrs, IPv6 curl, asserts `tcp_saddr`/`tcp_daddr` + `address_family=="ipv6"`), mirroring the IPv4 RX test.

## Verification

- ✅ `make all` green on the build host — all BPF probes (including the new IPv6 path: `ipv6hdr`, `__builtin_memcpy` addr copy, `tcp_v6_rcv`) and all 6 Go binaries compile.
- ✅ Struct-layout match empirically confirmed: Go `unsafe.Sizeof(netRcvPerfEvent{}) == 120` bytes == the BPF `perf_event_t` size — so no silent byte-mismatch corruption.
- ✅ `gofmt` clean.
- ⏳ **Integration tests need a root run** (no passwordless sudo in my build env): `sudo make integration` exercises the new IPv6 test plus the existing IPv4 RX and TX tests (backward compat). The IPv6 test `skip`s cleanly if the host lacks IPv6.

## Out of scope

- TX IPv6 *parsing* (TX is struct-synced here but stays IPv4-only) — separate follow-up.
- Prometheus metrics (still deferred from #343).


123123213weqw:
@huatuo-robot ## Reply to: [P1] Missing Apache 2.0 Copyright Header

Thank you for catching this. You were right — the new `bpf/net_tx_latency.c`
(the file that starts with `// go:build ignore`) shipped without the Apache
2.0 copyright header, while the other new files in this PR
(`core/events/net_tx_latency.go`, `integration/test_net_tx_latency.sh`, and
`integration/test_net_rx_latency_ipv6.sh`) all carry it correctly.

### What changed
- Added the standard Apache 2.0 license header to `bpf/net_tx_latency.c`,
  placed immediately before the existing `// go:build ignore` build-constraint
  line, exactly as suggested.
- The header text matches the maintainer's suggestion and is byte-for-byte
  identical to the header already present in `core/events/net_tx_latency.go`,
  keeping the PR internally consistent.

### Scope kept tight
Only the one file you flagged was modified. The header was inserted above the
`// go:build ignore` line, which is left untouched, so the file remains
excluded from the Go package build as before. No code or behavior changed.

### Verification
- `gcc -fsyntax-only -x c bpf/net_tx_latency.c`: the preprocessor lexes
  cleanly through the new comment header and the build-constraint line and
  only stops at the expected `#include "vmlinux.h"` (the BPF/vmlinux headers
  are not present in this sandbox), confirming the added comment block
  introduces no syntax errors.
- Confirmed the 13-line header is identical to the one in
  `core/events/net_tx_latency.go` via `diff`.
- Confirmed `// go:build ignore` is still present and correctly separated by
  a blank line.
- Note: the Go toolchain (`go`, `golangci-lint`, `gofumpt`) is not installed
  in this worktree, so `make check` could not be executed here. The change is
  a comment-only addition to a `// go:build ignore` file, so it is excluded
  from `import-fmt`/`golangci-lint` and does not affect the Go build. CI will
  run the full `make check` on push.

The diff is a single, additive change to one file.

123123213weqw:
@huatuo-robot Thanks for catching this. The two C test helpers were indeed missing the Apache 2.0 copyright header.

## What changed

I added the standard project Apache 2.0 header to both new C helper files that the integration scripts compile:

- `integration/testdata/test_net_rx_latency_ipv6_user.c` — compiled by `test_net_rx_latency_ipv6.sh`
- `integration/testdata/test_net_tx_latency_sink.c` — compiled by `test_net_tx_latency.sh`

These are exactly the C helpers the review referenced.

## Format choice

I matched the header byte-for-byte to the format already used in `bpf/net_tx_latency.c` and `core/events/net_tx_latency.go` (the `// Copyright 2026 The HuaTuo Authors` Apache-2.0 block), which is the established C/Go convention in this tree. It is placed immediately before the first `#include`, with a single blank line separating it from the code, mirroring the other C sources.

## Verification

- Both files still compile cleanly with the exact flags the integration scripts use:
  `cc -O2 -Wall -Wextra` — no warnings, no errors.
- Confirmed the new headers are identical to the canonical header in `bpf/net_tx_latency.c` via `diff`.
- The change is comment-only: `git diff` shows 14 added lines per file and no other modifications, so there is no behavior change.

Thanks again for the review.

123123213weqw:
@huatuo-robot ## Re: [P1] Missing Apache 2.0 Copyright Header

Thank you for catching this. The new file in question is
`integration/testdata/test_net_rx_latency_user.c` (the IPv4 helper added
in commit `a408c293`). It was the only net latency user-space C helper
still missing the header — its IPv6 sibling and the TX sink helper already
carried it after `c26e3f39`.

### What changed
Prepended the standard project Apache 2.0 copyright header to
`integration/testdata/test_net_rx_latency_user.c`, copied byte-for-byte
from `test_net_rx_latency_ipv6_user.c` (the same format also used in
`bpf/net_tx_latency.c` and `core/events/net_tx_latency.go`). The change is
comment-only: the existing code below the header is untouched.

### Verification
- Confirmed the first 14 lines of `test_net_rx_latency_user.c` now match
  `test_net_rx_latency_ipv6_user.c` exactly (`diff` of those lines is empty).
- Scanned all net latency user-space C helpers
  (`test_net_rx_latency_user.c`, `test_net_rx_latency_ipv6_user.c`,
  `test_net_tx_latency_sink.c`): all three now start with the
  `Copyright 2026 The HuaTuo Authors` header.
- Compiled `test_net_rx_latency_user.c` with the exact flags used by the
  integration script (`integration/test_net_rx_latency.sh:84`):
  `cc -O2 -Wall -Wextra` — builds cleanly with no warnings. Re-compiled the
  TX sink sibling as a regression control with the same result.

### Scope note
I limited the change to the file referenced by the review. Two unrelated,
pre-existing `.bpf.c` test fixtures (`test_bpf_dbg.bpf.c`,
`test_minimal.bpf.c`) do not carry this header, but they are outside the
net RX latency work in this PR, so I left them untouched to keep the change
minimal and focused. Happy to address those separately if you'd like a
follow-up.

123123213weqw:
@huatuo-robot # Reply: IPv6 TCP stage reported as `RX_STAGE_TCPV4`

Thanks for the catch — agreed this was misleading. Implemented your first
suggested option (protocol-agnostic stage), since `address_family` already
carries v4/v6 and it keeps the three-stage model intact (no duplicated
threshold).

## What changed

**Stage constant — `RX_STAGE_TCPV4` → `RX_STAGE_TCP`**
- `bpf/net_rx_latency.c`: renamed the enum value; both `tcp_v4_rcv_prog` and
  `tcp_v6_rcv_prog` now emit `RX_STAGE_TCP` (the `AF_INET`/`AF_INET6`
  argument still distinguishes the protocol on the wire and in userspace).
- `core/events/net_rx_latency.go`: `latStageNames[1]` is now `"RX_STAGE_TCP"`,
  so both IPv4 and IPv6 events surface a correct, family-neutral stage name.

**Threshold renamed for consistency — `rxlat_thresh_tcpv4` → `rxlat_thresh_tcp`**
This is the same class of v4-specific naming at the same two call sites: the
threshold is shared by both kprobes (including the IPv6 one), so I renamed the
BPF `volatile const` and its Go loader map key to match. This also aligns with
the already protocol-agnostic `cfg.NetRxLatency.Driver2TCP` config field and the
existing `TX_STAGE_*` naming. I deliberately did **not** add a separate
`RX_STAGE_TCPV6`/`rxlat_thresh_tcpv6`, because it would force duplicating the
threshold and split one concept into two.

**Documentation kept in sync**
- `docs/key-feature/instant-observability_en.md` and `_zh.md`: the documented
  `lat_stage` value is updated to `RX_STAGE_TCP`.
- `build/docker/grafana/dashboards/autotracing-event.json`: the panel text now
  lists `RX_STAGE_TCP`.

**Regression test** (`core/events/net_rx_latency_test.go`)
- `TestLatStageNamesProtocolAgnostic` asserts the TCP stage name is
  `RX_STAGE_TCP` and that no stage name embeds a `V4`/`V6` token.
- `TestLatStageNamesOrdering` pins `latStageNames` to the `enum rx_lat_stage`
  ordering (NETIF/TCP/USERCOPY) so a future C-enum change can't silently
  desync the Go slice and mislabel events or break the `latThresholds` lookup.

The kernel symbol names (`tcp_v4_rcv`, `tcp_v6_rcv`) are left untouched — those
are literal kernel functions, not stage identifiers.

## Verification

- `go build ./...` — passes (whole module).
- `go vet ./core/events/...` — clean.
- `go test ./core/... -count=1` — passes, including the two new tests.
- Mutation check: temporarily reverting the rename to `RX_STAGE_TCPV4` makes
  the new tests fail (and restoring it makes them pass again), confirming they
  genuinely guard the regression.
- BPF build reproduced with the project toolchain:
  `clang -Wall -O2 -g -target bpf -D__TARGET_ARCH_x86 -mcpu=v1 -c
  bpf/net_rx_latency.c -o net_rx_latency.o -I bpf/include` → compiles cleanly;
  the resulting object exposes the renamed `rxlat_thresh_tcp` const and contains
  no `tcpv4` identifier.
- `gofmt` applied to the changed Go files (they are gofmt-clean). Note: the
  repo's stricter `gofumpt`/`goimports`/`golangci-lint` (`make check`) were not
  available in this environment; the Go files are canonical `gofmt` output, and
  the import grouping matches the existing convention.

Left the changes in the worktree, uncommitted, for you to publish.

123123213weqw:
@huatuo-robot ## Reply to [P2] IPv6 extension headers not handled in `skb_tcp_family`

Thank you for the clear write-up and for confirming this is acceptable as a known v1 limitation.

I added the requested explanatory comment directly at the IPv6 `nexthdr` check in `skb_tcp_family()` (`bpf/net_rx_latency.c`):

```c
if (proto == bpf_ntohs(ETH_P_IPV6)) {
    struct ipv6hdr h;

    bpf_probe_read(&h, sizeof(h), skb_network_header(skb));
    // NOTE: nexthdr check does not walk extension header chains;
    // IPv6+EH TCP packets are caught by the tcp_v6_rcv
    // kprobe instead.
    return (h.nexthdr == IPPROTO_TCP) ? AF_INET6 : 0;
}
```

It captures exactly the two points you raised:
1. `nexthdr` only inspects the first next-header field and does not walk the extension header chain (Hop-by-Hop, Routing, Fragment, etc.), so IPv6+EH TCP packets return 0 and are silently skipped by the `netif_receive_skb` and `skb_copy_datagram_iovec` tracepoints (2 of 3 RX stages).
2. Those packets are still captured by the `tcp_v6_rcv` kprobe, since the kernel has already parsed the extension headers by the time that probe fires (and it is family-checked by symbol rather than via `skb_tcp_family`).

Wording note: I kept your suggested text verbatim, only wrapping the second line so the comment stays within the project's 80-column `clang-format` limit (the lines are indented two levels / 16 columns inside the `if` block, so the single-line form would have been 83 columns).

This is a comment-only change; there is no behavioral change and no new build dependency. I rebuilt `net_rx_latency.o` with the same `build/clang.sh` toolchain used by `make bpf-build` in CI and it compiled cleanly. (`clang-format` itself is not installed in this environment, so I verified the 80-column compliance manually.)

123123213weqw:
@huatuo-robot ## Reply to maintainer review

Hi, and thank you for the careful review.

### [P1] No bounds check on `pd.LatStage` before array index

Agreed — this was a real panic risk. `perf_event_t.lat_stage` is a `uint8`
straight from BPF perf data, and indexing `txStageNames` (2 entries) /
`latThresholds` directly with it would panic on any corrupt or unexpected
value, taking down the whole tracing goroutine.

What I changed:

- Added a small shared helper `lookupLatStage(stage, names, thresholds)`
  in `core/events/net_rx_latency.go` that resolves the stage name **and**
  the per-stage threshold only after confirming the index is in range for
  **both** slices, returning `ok=false` otherwise.
- `core/events/net_tx_latency.go` (the new code you flagged) now calls it
  and skips the event with `log.Warnf("net_tx_latency: unknown lat_stage %d,
  skipping", pd.LatStage)` + `continue`, exactly in the spirit of your
  suggested fix.
- Since you also noted the identical pre-existing pattern in
  `net_rx_latency.go`, I applied the same guard there too so the two code
  paths stay consistent (it's the same bug, same one-line fix, same package).
  I chose a shared helper rather than duplicating the `if` so the behavior is
  defined and tested in one place — happy to inline it instead if you prefer
  the literal form from your suggestion.

Verification added (`core/events/net_tx_latency_test.go`):

- `TestLookupLatStageBounds` — for both the TX and RX stage sets, every
  in-range stage resolves with `ok=true`, and out-of-range values (the first
  out-of-range index, one past it, and `0xFF`) return `ok=false` **without
  panicking**.
- `TestLookupLatStageThresholdMismatch` — guards the "fail closed" path when
  the thresholds slice is shorter than the names slice.
- `TestTxStageNamesOrdering` — pins `txStageNames` to the BPF enum ordering
  (parity with the existing RX ordering test), since the bounds check relies
  on the slice being correctly sized.

I also reproduced the original failure to confirm the fix is real: a direct
`txStageNames[255]` access panics with `index out of range [255] with length
2`, while `lookupLatStage(255, ...)` returns `ok=false` cleanly.

Checks run in the worktree (all green):
- `go build ./core/...`
- `go vet ./core/events/`
- `go test ./core/events/`
- `golangci-lint run ./core/events/... --timeout=5m --config .golangci.yaml`
- `gofumpt`/`goimports`/`gofmt` (the `make import-fmt` pipeline) report the
  changed files compliant.

Note: the full `make unit` target also runs `bpf-build` (clang/LLVM + kernel
headers) and `make test` additionally drives the integration/e2e shell suites
that need a live K8s cluster with `/etc/kubernetes/pki`; those were not run in
this worktree because the BPF toolchain and cluster are not available here.
The change itself is Go-only and confined to `core/events/`, so the BPF
object layout and behavior are untouched.

No changes were committed or pushed; the verified edits are left in the
worktree for publishing.

123123213weqw:
@huatuo-robot ## Reply to maintainer review

Hi, and thank you for the careful review.

### [P2] `filterTxByConfigAndResolveContainerID` returns empty containerID with ok=true

Good catch flagging this for verification — I traced the full downstream path
rather than assuming.

**Verification result: `tracing.Save` does handle an empty `containerID`
gracefully.**

`tracing.Save` → `documentWriter.saveRaw` → `newBaseDocument`
(`pkg/tracing/document_writer.go:118`) short-circuits an empty containerID:

```go
if req.ContainerID == "" {
    return &document, nil
}
```

It returns the document with **no error** and leaves the container-enrichment
fields (`ContainerID`, `ContainerHostname`, `ContainerType`, `ContainerQoS`,
…) empty, while preserving the latency payload. There is no nil-pointer deref
and no `"container not found"` failure, so the event is stored as an
un-attributed document rather than dropped or errored.

Because the verification passed, the conditional you suggested —
`return "", false` to skip — was **not** applied. I want to be explicit about
why, since the alternative of changing it would have been a real regression:

1. **The empty-containerID path is intentional and load-bearing.** It is not
   specific to net_tx/rx latency. `core/events/netdev_events.go:217` already
   calls `tracing.Save` with no `ContainerID` at all for host-level link-status
   events. And for net_tx/rx latency, when `ExcludedHostNetnamespace` is
   `false`, host-namespace events flow straight into the `ct == nil` branch
   (the host netns is not a normal container, so `pod.ContainerByNetInode`
   legitimately returns `nil, nil`). Returning `("", false)` there would
   **silently drop host network latency events** whenever host-netns exclusion
   is disabled.

2. **TX already matches RX.** `filterByConfigAndResolveContainerID` in
   `core/events/net_rx_latency.go` uses the identical
   `if ct == nil { return "", true }`. Changing only TX to `return "", false`
   would make the two diverge, which is the opposite of "matching RX
   behavior." If you'd prefer both loops to drop un-attributable events, I'm
   happy to do that as a follow-up — but it's a behavior change to RX
   (pre-existing, outside this PR's scope) and would also drop the host-netns
   case above, so I didn't make it unilaterally.

So I kept the resolver behavior and instead made the verification you asked for
**durable as a regression test**.

**What I added** (`pkg/tracing/store_test.go`):

- `TestNewBaseDocumentEmptyContainerID` — calls `newBaseDocument` (the core of
  `tracing.Save`) with `ContainerID: ""`, and asserts: no error is returned;
  the document is non-nil; all container-enrichment fields are empty; and the
  tracer name, run type, and latency payload (`TracerData`) are preserved, so
  the event is not silently lost. This is exactly the contract the net_tx/rx
  resolvers rely on when they yield `("", true)`.

I confirmed the test is a genuine guard rather than a no-op: temporarily
removing the `if req.ContainerID == ""` short-circuit in `newBaseDocument`
makes the test fail with `container  not found` (the lookup falls through to
`pod.ContainerByID("")` → not found → error), then restoring it makes the test
pass again.

Checks run in the worktree (all green):
- `go build ./pkg/tracing/... ./core/events/...`
- `go vet ./pkg/tracing/... ./core/events/...`
- `go test ./pkg/tracing/... ./core/events/...` (including the new test)
- `golangci-lint run ./pkg/tracing/... ./core/events/... --timeout=5m --config .golangci.yaml` → exit 0
- `gofumpt`, `goimports -local huatuo-bamai`, and `gofmt` all report
  `pkg/tracing/store_test.go` compliant (the `make import-fmt` pipeline).

Note: the full `make unit` target additionally runs `bpf-build` (clang/LLVM +
kernel headers) and `make test` drives integration/e2e shell suites that need
a live K8s cluster with `/etc/kubernetes/pki`; those were not run here because
the BPF toolchain and cluster are unavailable in this worktree. The change is
Go-only and confined to `pkg/tracing/`, so BPF object layout and behavior are
untouched.

No changes were committed or pushed; the verified edits are left in the
worktree for publishing.

123123213weqw:
Update on the integration-test status (the ⏳ in the PR description): I ran them with root on the build host (Linux 6.8) and they exposed two harness bugs, now fixed in this stack.

## Bugs found & fixed

1. `HUATUO_BAMAI_INTEGRATION_ARGS_STR` was never initialized by the harness, so `test_net_tx_latency.sh` / `test_net_rx_latency_ipv6.sh` died under `set -u` at the save/restore line before doing anything. Fixed by initializing it in `integration/env.sh` (mirrors the existing `HUATUO_BAMAI_E2E_ARGS_STR`). — landed in #343 (the base).
2. `integration_huatuo_bamai_start` defaults to `--disable-storage` when called with only a config writer, so the `[Storage.LocalFile]` events these tests assert on were never written. Fixed by passing `HUATUO_BAMAI_ARGS_INTEGRATION` to the helper so it takes its non-default branch. — in #343 (tx) and here (ipv6).

## Feature verified working

With those fixes, `net_rx_latency` (v4 + v6) and `net_tx_latency` emit real latency events and LocalFile storage writes them. Confirmed by running huatuo-bamai directly (no `--disable-storage`): ~1700 `net_rx_latency` events captured in 12s on background traffic, and the per-tracer event files (`events/net_rx_latency`, `events/net_tx_latency`) are produced.

## Remaining (not feature bugs)

- The veth-based tests are timing-sensitive on this host: they rely on a slow-recv server inducing >1ms RX latency, which doesn't always exceed the 1ms threshold under load, so the per-veth event count is sometimes 0 → flaky pass/fail. Making it deterministic likely needs `tc netem` to guarantee the latency rather than relying on recv delay.
- `test_net_tx_latency.sh` needs `ip netns add`, which this host rejects, so it SKIPs here (passes-by-skip). It should run in an env that allows netns creation.

Happy to follow up with the `tc netem` hardening if that's useful.

123123213weqw:
@hao022 note on the earlier `huatuo-robot` copyright-header findings, in case it comes up in review:

- `bpf/net_tx_latency.c`, `core/events/net_tx_latency.go`, and the two `integration/testdata/*.c` helpers now carry the Apache 2.0 header (added after the bot ran).
- `bpf/net_rx_latency.c` intentionally does **not** — it follows the upstream BPF C convention (`bpf/dropwatch.c`, `bpf/oom.c`, `bpf/native_cpu_profiler.c` all start with `#include "vmlinux.h"` and carry no copyright header). The bot appears to treat `.c` files uniformly regardless of this convention.

If you'd prefer the BPF C files to carry headers too, I'm happy to add one — just let me know which way you want the repo to go and I'll apply it consistently (here and in the sibling net-latency PRs).

hao022:
@123123213weqw 你好，感谢贡献代码。这个pr里的commit非常多，不容易审查。你可以先将 “feat(net_rx_latency): support IPv6 alongside IPv4” 这个特性单独提交上来。等这个过了，其他的特性再独立一个pr审查。这样能够提高合入的进度。

123123213weqw:
按 @hao022 在 #346 的建议拆分：IPv6 特性已单独重新提交为 #505（已变基到最新 main）。本 PR 关闭，其余特性后续分别独立提交。/ Split as requested: the IPv6 feature now lives standalone in #505 (rebased onto latest main). Closing; remaining features will be resubmitted as separate PRs.

PR #362, opened 2026-07-13
fix(dropwatch): correct packet metadata layout

messere1:
## Summary

- remove the stale Go-only `Type` field that shifts every packet metadata field after `drop_reason` by four bytes
- mirror the C struct's tail padding explicitly and add compile-time field-offset checks
- add a regression test that decodes a perf sample laid out like `struct packet_meta`

This restores correct `net_namespace_inode`, device name, and task command values, allowing container lookup by network namespace inode to work again.

## Testing

- `go test ./cmd/dropwatch`
- `make check`
- `make unit` *(WSL baseline limitation: `internal/cgroups/TestCgroupInterfaces/CpuUsage` expects cgroup v1 `cpuacct.stat`; the same test fails on unmodified upstream `main` in this environment)*

### DIFF
--- a/cmd/dropwatch/bpf_records_test.go
+++ b/cmd/dropwatch/bpf_records_test.go
@@ -1,0 +1,71 @@
+// Copyright 2026 The HuaTuo Authors.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package main
+
+import (
+	"bytes"
+	"encoding/binary"
+	"testing"
+
+	"huatuo-bamai/internal/utils/bytesutil"
+)
+
+// TestPacketMetaParse decodes a perf sample laid out exactly like struct
+// packet_meta in bpf/dropwatch.c and checks that every field lands where the
+// kernel wrote it. It fails if the Go mirror drifts from the C layout, e.g.
+// the stale 4-byte Type field that shifted net_inum, dev_name and comm.
+func TestPacketMetaParse(t *testing.T) {
+	buf := make([]byte, 96)
+
+	le := binary.LittleEndian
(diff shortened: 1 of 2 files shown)


cmd/dropwatch/bpf_records_test.go
```diff
@@ -0,0 +1,71 @@
...
+// the stale 4-byte Type field that shifted net_inum, dev_name and comm.
+func TestPacketMetaParse(t *testing.T) {
+	buf := make([]byte, 96)
+
+	le := binary.LittleEndian
+	le.PutUint64(buf[0:], 1111)        // ktime_ns
+	le.PutUint64(buf[8:], 2222)        // tgid_pid
+	le.PutUint64(buf[16:], 3333)       // net_cookie
+	le.PutUint64(buf[24:], 4444)       // kfree_skb_addr
+	le.PutUint64(buf[32:], 5555)       // memcg_css_addr
```
huhong-web: the number of test cases need to be meaningful, eg, change 1111 to a likely random ktime_ns and "1111" is used multimes.

cmd/dropwatch/bpf_records_test.go
```diff
@@ -0,0 +1,71 @@
...
+)
+
+// TestPacketMetaParse decodes a perf sample laid out exactly like struct
+// packet_meta in bpf/dropwatch.c and checks that every field lands where the
+// kernel wrote it. It fails if the Go mirror drifts from the C layout, e.g.
+// the stale 4-byte Type field that shifted net_inum, dev_name and comm.
+func TestPacketMetaParse(t *testing.T) {
+	buf := make([]byte, 96)
+
+	le := binary.LittleEndian
```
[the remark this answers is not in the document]
messere1: Fixed in e74f260b. The fixture now uses `binary.NativeEndian` for all integer fields, matching `binary.Read` and keeping the layout test endian-neutral. Verified with `go test ./cmd/dropwatch -run '^TestPacketMetaParse$' -count=1` and `make check`.

cmd/dropwatch/bpf_records_test.go
messere1: Addressed in 588d68b8. The fixture now uses deterministic, field-specific values: a realistic `ktime_ns`, an encoded TGID/PID, asymmetric cookie bytes, and representative kernel addresses. Named constants are reused by both the buffer writer and assertions, so the magic literals are no longer repeated. Verified with `go test ./cmd/dropwatch -run '^TestPacketMetaParse$' -count=1` and `make check`.

huhong-web:
great! thanks a lot.

PR #366, opened 2026-07-14
fix(job): synchronize terminal lifecycle

### LINKED ISSUE #363: [Bug]: internal/job manager races between Stop and monitorJob
### 这个问题是否已经存在？

### 部署环境

- HUATUO commit: `e347a544fdd88544a4f1e8de2114af4b855112d2` (current `upstream/main`)
- Go: `go1.24.13 linux/amd64`
- OS/kernel: `Linux 5.15.167.4-microsoft-standard-WSL2 x86_64`
- Module: `internal/job`

The issue is reproduced by the existing unit test and does not require a real cluster or agent deployment.

### 如何复现

1. Check out `e347a544fdd88544a4f1e8de2114af4b855112d2`.
2. Run:

```bash
go test -race ./internal/job -run '^TestManagerCreate$/^create_success$' -count=5
```

3. The race detector reports an unsynchronized read/write of `job.Status`:

```text
Read:
  internal/job.(*Manager).monitorJob.func1()
  internal/job/manager.go:296

Previous write:
  internal/job.(*Manager).updateJobStatus()
  internal/job/manager.go:234
  internal/job.(*Manager).Stop()
  internal/job/manager.go:154
```

The existing `create success` test follows a normal production path: `Create` starts `monitorJob`, then `Stop` closes the stop channel and transitions the job to `stopped`. The monitor goroutine concurrently reads the same mutable status in its deferred cleanup.

Expected: stopping and monitoring a job must not race, and a terminal `stopped` transition must not be overwritten.

Actual: `Stop/updateJobStatus` writes `job.Status` while `monitorJob` reads it without synchronization. Depending on interleaving, the deferred cleanup can also attempt a second stop/final transition, save a conflicting final state, or decrement the per-host job count more than once.

### 修复建议或方案

- Synchronize all mutable job lifecycle fields, either with a per-job mutex/state object or a manager-owned transition helper.
- Make terminal state transitions idempotent so only one path saves the final state, removes the in-memory job, and decrements `jobsByHost`.
- Ensure `monitorJob` cannot change a successfully stopped job to `failed` during deferred cleanup.
- Add a race regression test.

Acceptance command:

```bash
go test -race ./internal/job -run '^TestManagerCreate$/^create_success$' -count=10
```

### 其他

I searched the current open and closed issues for `internal/job` races and found no match.

Open PRs #299 and #306 address races in `pkg/tracing/task.go`; they do not touch this `internal/job` manager lifecycle race.

messere1:
## Summary

- serialize `Stop` and monitor transitions through per-job runtime state
- make terminal transitions idempotent so agent stop, persistence, cleanup, and host-count updates happen once
- return snapshots for in-memory reads and wait for monitor goroutines during shutdown
- add race regressions for concurrent stops and stopped-state overwrites

## Testing

- `make check`
- `go test -race ./internal/job -count=1`
- `go test -race ./internal/job -run '^TestManagerStop$' -count=20`
- `make unit` (`internal/job` passes; the full target only fails in `internal/cgroups/TestCgroupInterfaces/CpuUsage` because this WSL environment does not expose cgroup v1 `cpuacct.stat`)

Fixes #363

### DIFF
--- a/internal/job/manager.go
+++ b/internal/job/manager.go
@@ -47,6 +47,7 @@ type Manager struct {
 	storage    Store
 	nodeAgent  NodeAgent
 	stopChan   chan struct{}
+	stopOnce   sync.Once
 	config     ManagerConfig
 }
 
@@ -70,7 +71,24 @@ func newManagerWithStore(storage Store, nodeAgent NodeAgent, config ManagerConfi
 
 // Shutdown stops the manager and all background jobs
 func (m *Manager) Shutdown() {
-	close(m.stopChan)
+	var doneChans []chan struct{}
+	m.jobs.Range(func(_, value any) bool {
+		job := value.(*Job)
+		job.runtime.mu.Lock()
+		if job.runtime.doneChan != nil {
+			doneChans = append(doneChans, job.runtime.doneChan)
+		}
+		job.runtime.mu.Unlock()
+		return true
+	})
+
+	m.stopOnce.Do(func() {
+		close(m.stopChan)
+	})
+
+	for _, doneChan := range doneChans {
+		<-doneChan
+	}
 }
 
 func (m *Manager) Create(req CreateJobRequest) (*Job, error) {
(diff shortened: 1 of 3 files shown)


hao022:
@messere1 这个代码已经被重构（整体解决问题相比单点要好些）。可以参考最新的代码，哪里可以优化，直接提交pr即可。感谢pr

PR #368, opened 2026-07-14
fix(job): synchronize job lifecycle and protect bare type assertions

### LINKED ISSUE #363: [Bug]: internal/job manager races between Stop and monitorJob

wc4440222:
### Problem

`go test -race ./internal/job -run TestManagerCreate/create_success -count=5` reports a data race on `job.Status`:

- **Write**: `Stop()` → `updateJobStatus()` writes `job.Status` (manager.go:234)
- **Read**: `monitorJob` defer reads `job.Status` (manager.go:296) without synchronization

Depending on interleaving, the deferred cleanup can also:
- Overwrite a successfully stopped job to `failed`
- Save a conflicting final state
- Decrement `jobsByHost` more than once

### Fix

1. **Per-job mutex** (`sync.Mutex` on `Job` struct): protects `Status`, `LastUpdate`, `EndTime`, `Error` fields

2. **Idempotent terminal transitions**: `updateJobStatus` checks if the job is already in a terminal state before applying a new one. This prevents `monitorJob`'s defer from overwriting a `JobStatusStopped` set by `Stop()` with `JobStatusFailed`.

3. **Safe status reads**: `monitorJob`'s defer reads `job.Status` under lock before deciding to stop/fail the job. `StopAll` and `Delete` also read status under lock.

4. **7 bare type assertions fixed**: All `sync.Map.Load` results now use comma-ok pattern with nil checks:
   - `Create`: `jobCountVal.(int)`, `countVal.(int)`
   - `Stop`: `jobVal.(*Job)`
   - `Get`: `jobVal.(*Job)`
   - `List`: `value.(*Job)`
   - `StopAll`: `value.(*Job)`
   - `Delete`: `jobVal.(*Job)`
   - `updateJobStatus`: `countVal.(int)`

### Test

```
go test -race -count=10 ./internal/job/
```

All tests pass clean with `-race` detector. `TestMonitorJobDeferNoNilPanic` updated to expect `JobStatusStopped` (correct idempotent behavior) instead of `JobStatusFailed` (original buggy behavior). `stubJobStore` made thread-safe.

### Files Changed

- `internal/job/types.go` — add `mu sync.Mutex` to `Job` struct
- `internal/job/manager.go` — mutex protection + idempotent transitions + comma-ok assertions
- `internal/job/manager_test.go` — thread-safe stub + updated assertion

### DIFF
--- a/internal/job/manager.go
+++ b/internal/job/manager.go
@@ -69,6 +69,14 @@ func newManagerWithStore(storage Store, nodeAgent NodeAgent, config ManagerConfi
 }
 
 // Shutdown stops the manager and all background jobs
+
+// isTerminalStatus returns true if the job status is a terminal state
+// from which no further transition should occur.
+func isTerminalStatus(status JobStatus) bool {
+	return status == JobStatusCompleted || status == JobStatusStopped ||
+		status == JobStatusFailed || status == JobStatusTimeout
+}
+
 func (m *Manager) Shutdown() {
 	close(m.stopChan)
 }
@@ -80,8 +88,8 @@ func (m *Manager) Create(req CreateJobRequest) (*Job, error) {
 
 	jobCountVal, exists := m.jobsByHost.Load(req.Host)
 	if exists {
-		jobCount := jobCountVal.(int)
-		if jobCount >= m.config.MaxJobsPerHost {
+		jobCount, ok := jobCountVal.(int)
+		if ok && jobCount >= m.config.MaxJobsPerHost {
 			return nil, fmt.Errorf("maximum number of jobs reached for host %s", req.Host)
 		}
 	}
@@ -129,7 +137,9 @@ func (m *Manager) Create(req CreateJobRequest) (*Job, error) {
 
 	currentCount := 0
 	if countVal, exists := m.jobsByHost.Load(req.Host); exists {
-		currentCount = countVal.(int)
+		if n, ok := countVal.(int); ok {
+			currentCount = n
+		}
 	}
 	m.jobsByHost.Store(req.Host, currentCount+1)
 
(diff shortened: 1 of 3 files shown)


hao022:
@wc4440222 欢迎基于最新的代码进行优化。

PR #379, opened 2026-07-15
fix(autotracing): protect containersCPUIdle global map with mutex

### LINKED ISSUE #378: fix(autotracing): containersCPUIdle global map concurrent access causes fatal error
## Problem

In `core/autotracing/cpuidle.go`, the package-level global variable `containersCPUIdle` (type `map[string]*containerCPUInfo`) is accessed without any synchronization:

- **Written** by `updateContainersCPUIdle()` — inserts new entries and modifies existing entries
- **Read and deleted** by `detectCPUIdleContainer()` — iterates with `for id, container := range containersCPUIdle` and calls `delete(containersCPUIdle, id)` inside the loop

Both functions are called from `Start()` sequentially, but `containersCPUIdle` is a **package-level global variable**. If the tracing framework creates multiple `cpuIdleTracing` instances or restarts tracing (which is common in monitoring systems), two goroutines will access the same global map concurrently — one writing while another is iterating/deleting, causing:

```
fatal error: concurrent map writes
```
or
```
fatal error: concurrent map iteration and map write
```

This is a process-crashing bug, not just a data race.

## Fix

Add a `sync.Mutex` to protect all access to `containersCPUIdle`:

```go
var (
    containersCPUIdle   = make(containersCPUIdleMap)
    containersCPUIdleMu sync.Mutex
)
```

All reads, writes, and iterations must hold the mutex.

## Acceptance Criteria

- [ ] All access to `containersCPUIdle` protected by mutex
- [ ] `go build` passes
- [ ] No `concurrent map writes` possible under any goroutine schedule

wc4440222:
## Problem

In `core/autotracing/cpuidle.go`, the package-level global map `containersCPUIdle` is accessed without synchronization:

- **Written** by `updateContainersCPUIdle()` — inserts and modifies entries
- **Iterated and deleted** by `detectCPUIdleContainer()` — `for range` + `delete()` inside the loop

Since `containersCPUIdle` is a **package-level global variable**, if the tracing framework creates multiple `cpuIdleTracing` instances or restarts tracing, concurrent goroutines will access the same map, causing:

```
fatal error: concurrent map writes
```
or
```
fatal error: concurrent map iteration and map write
```

This is a process-crashing bug.

## Fix

Add a `sync.Mutex` to protect all access to `containersCPUIdle`:

```go
var (
    containersCPUIdle   = make(containersCPUIdleMap)
    containersCPUIdleMu sync.Mutex
)
```

Both `updateContainersCPUIdle` and `detectCPUIdleContainer` now acquire `containersCPUIdleMu.Lock()` before accessing the map.

Fixes #378

### DIFF
--- a/core/autotracing/cpuidle.go
+++ b/core/autotracing/cpuidle.go
@@ -23,6 +23,7 @@ import (
 	"path"
 	"runtime"
 	"strconv"
+	"sync"
 	"time"
 
 	"huatuo-bamai/internal/cgroups"
@@ -89,13 +90,19 @@ type cpuIdleThreshold struct {
 // containersCPUIdleMap is the container information
 type containersCPUIdleMap map[string]*containerCPUInfo
 
-var containersCPUIdle = make(containersCPUIdleMap)
+var (
+	containersCPUIdle   = make(containersCPUIdleMap)
+	containersCPUIdleMu sync.Mutex
+)
 
 func updateContainersCPUIdle(f *matcher.ContainerMatcher) error {
 	containers, err := pod.NormalContainers()
 	if err != nil {
 		return err
 	}
+
+	containersCPUIdleMu.Lock()
+	defer containersCPUIdleMu.Unlock()
 
 	for _, container := range containers {
 		if !f.Match(container) {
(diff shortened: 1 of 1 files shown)


hao022:
@wc4440222 https://github.com/ccfos/huatuo/pull/459 这个是类似的提交，但均没有并发访问的问题。

PR #389, opened 2026-07-18
feat(net_retransmit): trace TCP retransmits with 5-tuple + container context

123123213weqw:
## What

Adds a TCP retransmit tracer (kprobe `tcp_retransmit_skb`) emitting per-retransmit events with the 5-tuple (IPv4 + IPv6), socket state, device, and container / cgroup / netns context — the event-source foundation for the packet-loss ↔ retransmit network topology.

First slice of ccfos/huatuo#325 (continues the 7/12 claim). Topology graph / time-range query / Grafana viz are follow-ups.

## Why

#325 needs retransmit events correlated with dropwatch packet-loss events to build a topology of where loss / retransmits happen between nodes. huatuo had **no retransmit detection at all**, so this lands the event source first — the rest of the topology consumes it.

## How

- **BPF** `bpf/net_retransmit_skb.c` — kprobe on `tcp_retransmit_skb(struct sock *sk, struct sk_buff *skb, int segs)`; reads skb from PARM2. Family from `skb->protocol`; src/dst from the IP/IPv6 header (low-4-bytes-in-16 for v4, all 16 for v6), ports + seq from the TCP header — the same proven reads as `net_rx_latency`. Captures memory-cgroup CSS, netns cookie + inode, device name, comm/pid. Rate-limited.
  - kprobe was chosen over `tracepoint/sock/tcp_retransmit_skb` because the tracepoint is **absent** on the tested 6.8 kernel (`/sys/kernel/tracing/events` has no `tcp_retransmit_skb`), while the function is present in kallsyms. The function signature keeps `sk`=PARM1, `skb`=PARM2 across versions.
- **Go** `core/events/net_retransmit.go` — loads the object, attaches the single kprobe gated by `bpf.HasKprobeFunction("tcp_retransmit_skb")` (→ `types.ErrNotSupported` / inactive if absent), reads perf events, decodes, attributes to a container via CSS / net-cookie / net-inode (matching `dropwatch`), and saves a `NetRetransmitTracingData` event.
- **Schema** shares the address shape with net_rx/tx latency: 16-byte addresses keyed by `address_family` (`ipv4` / `ipv6`).

## Changes

- `bpf/net_retransmit_skb.c` (new) — kprobe + perf event struct.
- `core/events/net_retransmit.go` (new) — loader / attach / decode / save + container resolver.
- `core/events/net_retransmit_test.go` (new) — byte-layout decode tests; the 120-byte mirror is pinned by a compile-time size assert, and cases cover IPv4 / IPv6 address parsing and truncated input.

## Verification

On the `wzu` host (Linux 6.8; `tcp_retransmit_skb` present in kallsyms; BTF available):

- ✅ BPF compiles — `make all` builds `net_retransmit_skb.o` (CO-RE / BTF) and all 6 binaries.
- ✅ `go build ./...`, `go test ./core/events/...` (new decoder tests), `go vet`.
- ✅ `goimports` / `gofumpt` / `golangci-lint` (v1.62.2) clean.
- ⏳ Live retransmit generation not exercised here (needs `tc netem loss` on a veth); the probe loads / attaches by construction (HasKprobeFunction gate + object compiles + struct-layout pinned). Integration follow-up, same as the net_rx_latency IPv6 PR.

## Out of scope (follow-ups)

- 5-tuple → node topology graph + time-range query (the #325 viz piece).
- dropwatch ↔ retransmit correlation view.
- Grafana topology panel.
- Config (filter / throttle) — currently always-on when the tracer is enabled.

### DIFF
--- a/bpf/net_retransmit_skb.c
+++ b/bpf/net_retransmit_skb.c
@@ -1,0 +1,134 @@
+// go:build ignore
+
+#include "vmlinux.h"
+
+#include <bpf/bpf_core_read.h>
+#include <bpf/bpf_endian.h>
+#include <bpf/bpf_helpers.h>
+#include <bpf/bpf_tracing.h>
+
+#include "bpf_cgroup.h"
+#include "bpf_common.h"
+#include "bpf_net_namespace.h"
+#include "bpf_ratelimit.h"
+#include "bpf_sock.h"
+#include "vmlinux_net.h"
+
+// Cap the event rate so a retransmit storm cannot overwhelm userspace. Mirrors
+// net_rx_latency's limiter; tune later via config if needed.
+BPF_RATELIMIT(retrans_rate, 1, 100);
+
+// Addresses are 16-byte fields (IPv4 = low 4 bytes network order, IPv6 = all
+// 16) keyed by addr_family, exactly like net_rx_latency/net_tx_latency. Mirrored
+// byte-for-byte by core/events/net_retransmit.go netRetransmitPerfEvent.
+struct perf_event_t {
+	u64 ktime_ns;       // 8  — bpf_ktime_get_ns, for time-range correlation
+	u64 tgid_pid;       // 8
+	u64 memcg_css_addr; // 8  — container attribution (memory cgroup CSS)
+	u64 net_cookie;     // 8  — net namespace cookie (>= 5.14)
+	u64 pkt_len;        // 8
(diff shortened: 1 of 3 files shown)


123123213weqw:
整理我的开放 PR 队列，先关闭。分支保留，需要时重新打开或基于最新 main 重新提交。/ Cleaning up my open PR queue — closing for now. The branch is kept; will reopen or resubmit on latest main later.

PR #393, opened 2026-07-19
feat(net_rx_latency): add iptable rule-matching latency stage (#44)

123123213weqw:
## What
Adds an `RX_STAGE_IPTABLE` stage to `net_rx_latency` that measures `ipt_do_table()` self-duration (IPv4 iptables rule matching), surfacing slow rule evaluation on iptables-mode k8s (< v1.29) with complex rulesets. Addresses the iptable-granularity part of #44.

> **Stacked on #343 + #346.** Based on `feat/net-rx-latency-ipv6` (#346) because it reuses that tracer's `AttachWithOptions` + conditional-attach + `lookupLatStage` + ordering tests. The diff therefore includes #343/#346; merge #343 then #346, and this PR collapses to the iptable-only delta (review commit `a29851a` on top of `06d13bb`).

## Why
#44's netsendlat/netrecvlat main bodies shipped via #343/#346. The remaining ask is the **「更细粒度: iptable 转发时延」**: k8s < v1.29 defaults to iptables networking, and with complex rulesets `ipt_do_table()` (the rule-matching core) becomes a forward-path bottleneck. huatuo had no netfilter/iptable BPF at all.

## How
- **Function self-duration, not an skb-tstamp delta**: `kprobe/ipt_do_table` + `kretprobe/ipt_do_table` isolate rule-matching cost (a tstamp delta would conflate with NETIF-style receive delay).
- **entry↔ret correlation by `pid_tgid`** over `ipt_lat_map` (HASH, value=`{start_ns, skb}`), mirroring `memory_reclaim_events.c`'s softirq pattern — `ipt_do_table` runs in softirq with preemption disabled around the call, so pid_tgid is identical at entry/return. skb is stashed at entry because the ret probe can't read `PT_REGS_PARM1`.
- **Conditional attach** via `HasKprobeFunction("ipt_do_table")`: kernels without `CONFIG_IP_NF_IPTABLES` (or nftables-mode ≥ v1.29) skip both probes; `AttachWithOptions` keeps the rest of the tracer alive.
- **AF_INET/TCP only** (ipt_do_table is IPv4-only; ip6t_do_table / nft_do_chain out of scope).
- **Stage appended last** (index 3) so NETIF/TCP/USERCOPY indices stay stable.

## Verification
- ✅ `make all` green on the build host (24 BPF probes incl. the new ipt_do_table entry/ret + all 6 binaries).
- ✅ `go test ./core/events/... -run TestLatStage` green (ordering/bounds + new `TestLatStageIptableIsLastIndex`).
- ✅ gofmt clean.
- ⏳ Runtime needs a privileged iptables-mode k8s node (no passwordless sudo in my build env): real attach, measured deltas, reentrancy behavior pending.

## Known limitations
- **Reentrancy**: a reentrant call (user-defined `-j MYCHAIN` chain, TEE target) overwrites the outer entry; the outer sample is dropped — no false attribution, only a lost outer measurement.
- **Threshold semantics differ**: NETIF/TCP/USERCOPY are skb-tstamp deltas; IPTABLE is the function's own runtime. `lat_ms` is not directly comparable across stages (documented).
- **No userspace attribution** in softirq (`comm`/`pid` = `-`/0); pod attribution still works via `netns_inum`/`net_cookie`.

## Out of scope
`ip6t_do_table` (IPv6) and `nft_do_chain` (nftables, k8s ≥ v1.29 default) — separate follow-ups.


hao022:
@123123213weqw 建议这些 commit 分别pr；否则不好审查合并。

123123213weqw:
> @123123213weqw 建议这些 commit 分别pr；否则不好审查合并。

ok

123123213weqw:
整理我的开放 PR 队列，先关闭。分支保留，需要时重新打开或基于最新 main 重新提交。/ Cleaning up my open PR queue — closing for now. The branch is kept; will reopen or resubmit on latest main later.

PR #399, opened 2026-07-20
fix: group container processes by root

### LINKED ISSUE #397: [Bug]: container process grouping stops at the immediate parent
### 这个问题是否已经存在？

### 部署环境

HUATUO commit: `0ed6c0f974c0bd722ceae28254a1dc9c0403b193` (main)
Component: `internal/profiler/procutil` and Java profiler PID resolution
Go: `go1.24.13 linux/amd64`
OS: Ubuntu under WSL2

The grouping failure is deterministic and does not depend on kernel or cgroup features.

### 如何复现

1. Let the container process scan match a three-level process chain plus one independent process:

```text
PID 100, PPID 1
PID 101, PPID 100
PID 102, PPID 101
PID 200, PPID 1
```

2. Run the grouping logic currently embedded in `findProcessesInCgroups`. It only checks whether the immediate PPID is matched.

Actual result:

```text
map[100:[100 101] 101:[102] 200:[200]]
```

Expected result:

```text
map[100:[100 101 102] 200:[200]]
```

`ResolveJavaPids` consumes the map keys as profiler targets, so PID 101 can incorrectly become a second Java profiling target for the same matched process tree.

### 修复建议或方案

Build a PID-to-PPID map for all matched processes, then walk the complete matched ancestor chain for every process and group it under the highest matched ancestor. Detect parent cycles to avoid an infinite loop on malformed input.

Add regression tests for a three-level chain and a cyclic parent graph.

### 其他

A focused pre-fix test reproduced the incorrect result. The proposed regression suite passes with:

```text
go test -race ./internal/profiler/procutil -count=10
```

A duplicate search found no matching issue. PR #386 concerns opt-in native profiler eBPF fork tracking and does not modify this container/Java grouping path.

Loyal-Young:
## Summary

- Group matching container processes under their highest matching ancestor instead of only their immediate parent.
- Make the process-group selection deterministic and safely collapse malformed parent cycles.
- Add regression coverage for a three-level process chain and a cyclic parent graph.

Fixes #397.

## Validation

- `git diff --check`
- `GOOS=linux GOARCH=amd64 go test -c ./internal/profiler/procutil` — cross-compiled successfully on Windows.

The package cannot execute on this Windows host because it depends on Linux cgroup, namespace, and syscall APIs.

### DIFF
--- a/internal/profiler/procutil/container.go
+++ b/internal/profiler/procutil/container.go
@@ -20,6 +20,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"sort"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


Loyal-Young:
Closing this duplicate of #398 to keep the competition queue focused.

PR #401, opened 2026-07-20
fix(profiler): guard unexpected timing store values

Loyal-Young:
## Summary

- Treat unexpected values in the exported profiler timing stores as absent instead of panicking.
- Preserve valid timestamp retrieval.
- Add regression tests for both cases.

## Validation

- `git diff --check`


hao022:
@Loyal-Young 从实现上来看Store 只会存放一种结构，读取时无需复杂的判断。

PR #402, opened 2026-07-20
fix(tracing): allow stop before the first run

Loyal-Young:
## Summary

- Avoid calling a nil cancellation function when tracing is stopped before its first run starts.
- Preserve cancellation for active tracing runs.
- Add regression coverage for both lifecycles.

## Validation

- `git diff --check`


hao022:
@Loyal-Young 感谢贡献代码。当前的实现都会赋值c.cancelCtx，优化的意义不是很大。不过可以考虑 重复start/stop的错误管理。

PR #403, opened 2026-07-20
fix(request): reject raw reads before connection

Loyal-Young:
## Summary

- Return a descriptive error when `RawHTTP.ReadResponse` is called without a connected Unix socket.
- Guard both nil and zero-value receivers.
- Add focused regression coverage.

## Validation

- `git diff --check`


hao022:
internal/request 包，并没有使用可以删除掉。

PR #405, opened 2026-07-22
fix(profiling): preserve profile query matchers

messere1:
## Summary

- preserve every supplied profile target matcher in storage queries
- remove the duplicate tracer ID predicate
- add regression coverage for combined container matchers and tracer IDs

## Why

Pyroscope label selectors can contain multiple equality matchers. The storage query used a switch, so it silently kept only the first target matcher. This could broaden a profile query beyond the requested label set.

## Testing

- `make check`
- `go test -race ./internal/profiler/service -count=1`
- `make unit` (reaches the full Go suite, but this WSL environment lacks cgroup v1 `cpuacct.stat`; the unrelated `internal/cgroups/TestCgroupInterfaces/CpuUsage` test fails for that reason)

### DIFF
--- a/internal/profiler/service/storage.go
+++ b/internal/profiler/service/storage.go
@@ -265,7 +265,7 @@ func buildProfileSearchQuery(filter *SearchFilter) driver.Query {
 
 func buildProfileAggregationQuery(filter *SearchFilter) driver.Query {
 	query := driver.Query{
-		Filters: make([]driver.Filter, 0, 6),
+		Filters: make([]driver.Filter, 0, 8),
 	}
 
 	if filter == nil {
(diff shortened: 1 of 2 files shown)


messere1:
The retriggered CI indicates that the remaining failure is unrelated to this PR:

- The previous run failed only on Ubuntu 20.04 in `internal/pcapfilter/TestApply` with a CO-RE relocation error; Ubuntu 22.04 passed.
- The current run passes on Ubuntu 20.04, 24.04, and 26.04, but fails on Ubuntu 22.04 in `integration/test_metrics_exclude_filter.sh` because `netstat_Tcp_RetransSegs` was not observed on the first metrics scrape.
- Build, lint, vendor, and PR-name checks all pass.

This PR only changes `internal/profiler/service`. The integration helper waits for `/metrics` to respond, but does not wait for the expected collectors to publish their first sample, so this failure appears timing-related. Could a maintainer please rerun the failed Ubuntu 22.04 job?

PR #417, opened 2026-07-24
fix(pod): kubeletSyncContainers nil deref panic on missing ContainerStatus

### LINKED ISSUE #416: fix(pod): kubeletSyncContainers nil deref panic on missing ContainerStatus
## Bug Description

In `internal/pod/container_kubelet_sync.go`, the `kubeletSyncContainers` function can panic with a nil pointer dereference when a pod's `Spec.Containers` has entries that are missing from `Status.ContainerStatuses`.

### Root Cause

The function builds a map `m` from `pod.Spec.Containers` with `nil` as the second element:

```go
m := make(map[string][2]any)
for i := range pod.Spec.Containers {
    container := &pod.Spec.Containers[i]
    m[container.Name] = [2]any{container, nil}
}
```

Then fills in the second element from `pod.Status.ContainerStatuses`:

```go
for i := range pod.Status.ContainerStatuses {
    containerStatus := &pod.Status.ContainerStatuses[i]
    if c, ok := m[containerStatus.Name]; ok {
        m[containerStatus.Name] = [2]any{c[0], containerStatus}
    }
}
```

If a container in `Spec.Containers` has no corresponding entry in `Status.ContainerStatuses`, `c[1]` remains `nil`. The subsequent bare type assertion panics:

```go
for _, c := range m {
    containerStatus := c[1].(*corev1.ContainerStatus)  // panic: c[1] is nil
    // ...
}
```

### Why `isRuningPod` Doesn't Prevent This

`isRuningPod` only checks that all **existing** `ContainerStatuses` have `State.Running != nil`. It does **not** verify that the count of `ContainerStatuses` matches `Spec.Containers`:

```go
func isRuningPod(pod *corev1.Pod) bool {
    if pod.Status.Phase != corev1.PodRunning {
        return false
    }
    for i := range pod.Status.ContainerStatuses {
        containerStatus := &pod.Status.ContainerStatuses[i]
        if containerStatus.State.Running == nil {
            return false
        }
    }
    return true
}
```

A pod can be in `PodRunning` phase with fewer `ContainerStatuses` than `Spec.Containers` (e.g., during startup race, kubelet hasn't reported status for all containers yet).

## How to Reproduce

1. Create a pod with multiple containers.
2. During the brief window where the pod phase is `Running` but not all container statuses have been reported by kubelet, `kubeletSyncContainers` will encounter a container in `Spec.Containers` with no matching `ContainerStatuses` entry.
3. The bare type assertion `c[1].(*corev1.ContainerStatus)` panics on nil, crashing the process.

## Suggested Fix

Use a comma-ok type assertion and skip containers with nil status:

```go
containerStatus, ok := c[1].(*corev1.ContainerStatus)
if !ok || containerStatus == nil {
    continue
}
```

## Impact

- **Severity**: High — causes a process crash (panic), not just incorrect metrics.
- **Trigger**: Race condition during pod startup; transient but reproducible.

wc4440222:
## Fix

Fixes #416

## Problem

In `internal/pod/container_kubelet_sync.go`, `kubeletSyncContainers` builds a map from `pod.Spec.Containers` with `nil` as the second element, then fills in `ContainerStatus` from `pod.Status.ContainerStatuses`. If a container in `Spec.Containers` has no matching entry in `ContainerStatuses`, `c[1]` remains `nil`, and the bare type assertion panics:

```go
containerStatus := c[1].(*corev1.ContainerStatus)  // panic on nil
```

`isRuningPod` only checks that existing `ContainerStatuses` are all `Running`, but does not verify that the count matches `Spec.Containers`. A pod in `PodRunning` phase can have fewer `ContainerStatuses` than `Spec.Containers` during a startup race.

## Fix

Use a comma-ok type assertion and skip containers with nil status:

```go
containerStatus, ok := c[1].(*corev1.ContainerStatus)
if !ok || containerStatus == nil {
    continue
}
```

## Verification

- `go build ./internal/pod/` — pass
- `golangci-lint run ./internal/pod/... --timeout=5m --config .golangci.yaml` — pass (no warnings)
- Code review by sub-agent — pass

### DIFF
--- a/internal/pod/container_kubelet_sync.go
+++ b/internal/pod/container_kubelet_sync.go
@@ -254,7 +254,10 @@ func kubeletSyncContainers() error {
 		}
 
 		for _, c := range m {
-			containerStatus := c[1].(*corev1.ContainerStatus)
+			containerStatus, ok := c[1].(*corev1.ContainerStatus)
+			if !ok || containerStatus == nil {
+				continue
+			}
 			containerID, err := parseContainerIDInPodStatus(containerStatus.ContainerID)
 			if err != nil {
 				log.Warnf("failed to parse container id %s in pod %s status: %v", containerStatus.ContainerID, pod.Name, err)


hao022:
@wc4440222 根据上下文，这里不需要判断，可以直接访问。

PR #419, opened 2026-07-24
fix(autotracing): cpuidle divide-by-zero panic when update interval < 1us

### LINKED ISSUE #418: fix(autotracing): cpuidle divide-by-zero panic when update interval < 1us
## Bug Description

In `core/autotracing/cpuidle.go`, the `updateContainerCpuUsage` method computes the elapsed time between two CPU usage samples in microseconds, then immediately uses it as a divisor:

```go
updateElapsed := time.Since(container.updateTime).Microseconds()

container.nowUsagePercentage.user = 100 * delta.user / updateElapsed / cpuCores
container.nowUsagePercentage.sys = 100 * delta.sys / updateElapsed / cpuCores
container.nowUsagePercentage.total = 100 * delta.total / updateElapsed / cpuCores
```

If `updateElapsed` is 0 — which happens when two consecutive calls occur within the same microsecond — the integer division triggers a **fatal `runtime error: integer divide by zero` panic**, crashing the entire cpuidle autotracing goroutine.

### Trigger Scenario

The cpuidle tracer runs on a ticker (default 20s interval), so under normal conditions `updateElapsed` is large. However, the panic can be triggered when:

1. The ticker fires twice in rapid succession due to timer coalescing or clock adjustments
2. A container is re-added to the tracking map after deletion, causing a first-update path that resets `updateTime` to `time.Now()`, followed by an immediate second call within the same microsecond
3. High-frequency container churn causes `detectCPUIdleContainer` to call `updateContainerCpuUsage` on a container whose `updateTime` was just set

## Suggested Fix

Guard against `updateElapsed == 0` before the division:

```go
updateElapsed := time.Since(container.updateTime).Microseconds()
if updateElapsed == 0 {
    container.updateTime = time.Now()
    return fmt.Errorf("cpu usage update interval too small")
}
```

This is consistent with the existing pattern in the same function where `delta.total == 0` and first-update cases already return errors.

## Impact

- **Severity**: High — causes a process crash (panic), not just incorrect metrics.
- **Trigger**: Rare but possible under timer coalescing or rapid container lifecycle events.

wc4440222:
## Fix

Fixes #418

## Problem

In `core/autotracing/cpuidle.go`, `updateContainerCpuUsage` computes `updateElapsed := time.Since(container.updateTime).Microseconds()` and immediately uses it as a divisor in integer arithmetic. When two calls occur within the same microsecond, `updateElapsed` is 0, causing `runtime error: integer divide by zero` panic.

## Fix

Add a guard clause before the division:

```go
if updateElapsed == 0 {
    container.updateTime = time.Now()
    return fmt.Errorf("cpu usage update interval too small")
}
```

This is consistent with the existing error-return pattern in the same function (first-update and no-change cases).

## Verification

- `go build ./core/autotracing/` — pass
- `golangci-lint run ./core/autotracing/... --timeout=5m --config .golangci.yaml` — pass (no warnings)
- Code review by sub-agent — pass

### DIFF
--- a/core/autotracing/cpuidle.go
+++ b/core/autotracing/cpuidle.go
@@ -199,6 +199,10 @@ func (c *cpuIdleTracing) updateContainerCpuUsage(container *containerCPUInfo) er
 	}
 
 	updateElapsed := time.Since(container.updateTime).Microseconds()
+	if updateElapsed == 0 {
+		container.updateTime = time.Now()
+		return fmt.Errorf("cpu usage update interval too small")
+	}
 
 	container.nowUsagePercentage.user = 100 * delta.user / updateElapsed / cpuCores
 	container.nowUsagePercentage.sys = 100 * delta.sys / updateElapsed / cpuCores


wc4440222:
@hao022 Thanks for the recent refactoring of `cpuidle.go`. I noticed that in the current `main` branch, `containerCPUState.update()` (line 260) already guards against the divide-by-zero case:

```go
elapsedMicroseconds := elapsed.Microseconds()
if elapsedMicroseconds <= 0 || cpuCapacity <= 0 {
    return ...
}
```

It looks like this PR's fix has been covered by your refactoring. Would you like me to close this PR, or would you prefer to keep it open if there's any additional edge case worth preserving?

Issue #329, opened 2026-07-07
[功能建议]: 持续 Profiling 支持多维度采集


fanzu8:
## 背景

当前的持续 Profiling 采集维度较为单一，难以满足复杂场景下的精细化性能分析需求。

## 需求描述

扩展持续 Profiling 的采集能力，支持多维度采集：

1. **多维度采集**：支持按 PID/TGID、cgroup、进程组维度进行性能数据采集
2. **多维度标签注入**：将采集维度信息作为标签注入 profiling 数据，便于后续查询和聚合
3. **锁类型扩展**：扩展支持多种内核锁类型的性能剖析
   - mutex（互斥锁）
   - spinlock（自旋锁）
   - rwlock（读写锁）

## 技术要求

- 新采集维度不影响现有 profiling 功能
- 标签格式兼容主流 profiling 后端（Pyroscope/Parca）

## 验收标准

- [ ] PID/TGID/cgroup/进程组维度采集可用
- [ ] 多维度标签正确注入
- [ ] 三种锁类型性能剖析可用

## 意向参与贡献

- [ ] 我有意向参与具体功能的开发实现并将代码贡献回到上游社区

123123213weqw:
I'd like to take this one. Scope I'm planning:

1. Dimension capture (PID/TGID already present in `SampleOutput` at profiler.go:153; extend the profiler config + per-sample path to also carry cgroup path and process-group key, fetched once per flush to avoid per-sample cost).
2. Label injection into the collapsed/flamegraph output (internal/profiler/output). Tags go into the first frame field (consistent with the existing `threadName#pid` pattern at profiler.go:279) so Pyroscope/Parca/speedscope/chrometrace consumers all see them without a backend-specific branch.
3. Lock profiling extensions — three tracepoints in bpf/, mirroring the structure of `native_cpu_profiler.c`:
   - mutex contention: `bpf_get_current_cgroup_id` + `mutex_lock`/`mutex_unlock` (or `__contention_begin` tracepoints where available, with pre-6.2 fallback to kprobe on `mutex_lock`/`mutex_unlock`)
   - spinlock: `__spin_lock_*` / `__rq_lock_*` contention probes
   - rwlock: `rwlock_*` read/write contention split into two profile types so readers/writers are distinguishable
   Each emits the same `perf_event_t`-style record so the existing aggregator path is reused.
[...]

PR #471, opened 2026-07-27
fix(profiler): resolve valid zero stack ids

lizhenhao123:
## Summary

- treat stack ID zero as a valid BPF stack-map entry
- key user-stack resolution caches by both stack ID and process ID
- prevent valid samples from being dropped or resolved against another process

## Testing

- `go test ./cmd/profiler/provider`
- `go test -race ./cmd/profiler/provider`
- `go vet ./cmd/profiler/provider`
- `git diff --check`

Refs #329


hao022:
@lizhenhao123 你好，感谢贡献代码。

1. hasStack 更多的是stack的有效性，因此建议改为 validateStackID 并增加注释为什么，尤其是0的情况。
2. ustackCache 加入pid元素，没有看出必要性。

PR #473, opened 2026-07-27
feat(profiling): forward external profiler tool path

lizhenhao123:
## Summary

- replace language-specific profiler paths with one `binary_tool_path` API field
- require and trim the field for Java and Python jobs before forwarding `--tool-path`
- reject tool paths for native profilers instead of silently ignoring them
- update the Profiles API documentation and examples in English and Chinese

## Testing

- `go test ./apis/v1 ./cmd/huatuo-apiserver/handlers/profiling ./cmd/profiler`
- matching race tests and `go vet`
- `make check` (26 linters, 0 issues)

Refs #329

### DIFF
--- a/apis/v1/types.go
+++ b/apis/v1/types.go
@@ -21,13 +21,14 @@ import (
 
 // CreateProfilingJobRequest represents a request to create a profiling job.
 type CreateProfilingJobRequest struct {
-	ProfilingType   string `json:"type"`              // cpu or memory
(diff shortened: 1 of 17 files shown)


hao022:
@lizhenhao123 --tool-path 1. 这个命令参数 java, python 复用是不是更好；2. --tool-path 可以考虑从命名 --tool-path-dir tool-path-dir/bin/asprof、tool-path-dir/lib/libasyncProfiler.so、tool-path-dir/py-spy ？

PR #474, opened 2026-07-27
feat(profiling): inject collection dimension labels

lizhenhao123:
## Summary

- derive a fixed set of managed collection labels from `ProfilerContext`
- persist scope, PID/TGID, container, and CPU dimensions in profile metadata
- inject the same dimensions into standard pprof sample labels for compatible consumers
- clear stale managed labels while preserving unrelated existing labels

## Testing

- `go test ./internal/profiler ./internal/profiler/context ./internal/profiler/aggregator`
- matching race tests and `go vet`
- `git diff --check`

Refs #329

### DIFF
--- a/internal/profiler/types.go
+++ b/internal/profiler/types.go
@@ -1,4 +1,4 @@
-// Copyright 2025 The HuaTuo Authors
+// Copyright 2025, 2026 The HuaTuo Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -36,6 +36,8 @@ const MetadataCollection = "profiling_metadata"
 // ProfileData is the data saved by the profiler.
 type ProfileData struct {
 	ProfileType string `json:"profile_type,omitempty"`
+	// Labels mirror profile-wide dimensions for storage backends.
+	Labels map[string]string `json:"labels,omitempty"`
 	// Please note:
 	//
 	//	In pyroscope 1.13.0, use profilev1.Profile instead of ptree.Profile, but it depends
--- a/internal/profiler/aggregator/pipeline_test.go
+++ b/internal/profiler/aggregator/pipeline_test.go
@@ -20,9 +20,15 @@ import (
 	"testing"
 	"time"
 
+	"huatuo-bamai/internal/profiler"
 	profctx "huatuo-bamai/internal/profiler/context"
 	"huatuo-bamai/internal/profiler/output"
+	"huatuo-bamai/internal/toolstream"
 )
+
+type pipelineProfileEvent struct {
+	TracerData *profctx.TracerData `json:"tracer_data"`
+}
 
 func TestNewPipeline_DoesNotMutateContext(t *testing.T) {
 	pctx := &profctx.ProfilerContext{}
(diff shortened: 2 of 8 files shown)


hao022:
@lizhenhao123 这个能够给出一个效果图吗？以及在什么场景下使用？

internal/profiler/types.go
```diff
@@ -36,6 +36,8 @@ const MetadataCollection = "profiling_metadata"
 // ProfileData is the data saved by the profiler.
 type ProfileData struct {
 	ProfileType string `json:"profile_type,omitempty"`
+	// Labels mirror profile-wide dimensions for storage backends.
+	Labels map[string]string `json:"labels,omitempty"`
```
huhong-web, line 40: [P1] Complete the persisted-label read/query contract

This field writes collection dimensions under `tracer_data.flamedata.labels`, but the supported read path drops them: `internal/profiler/service/storage.go` defines `ProfileDocument.TracerData.Flamedata` with only `profileType` and `profile`; `applyProfileMatcher` rejects `profiling_scope`, `cpu`, `pid`, and `tgid`; and `LabelNames`/`LabelValues` never expose them. A profile collected with `--pid 42` therefore stores the value `42` for `pid` but cannot be read or selected as `{pid="42"}` through the service API, so the primary query/aggregation use case remains unavailable.

Please carry `Labels` into `ProfileDocument`, map each managed label to its indexed Elasticsearch field in `SearchFilter`/`applyProfileMatcher`, and expose the names and values through `LabelNames`/`LabelValues`. Add a storage/query contract test that writes a labeled profile, reads it back, and proves the selector matches it, including compatibility with legacy documents.

huhong-web (commented):
## Code Review

Found one must-fix cross-layer contract gap; see the inline comment on `ProfileData.Labels`.

### Architecture & Design

**Verdict: needs changes.**

The producer-side design is otherwise cohesive: `ProfilerContext` derives the bounded collection dimensions, `profiler.ApplyCollectionDimensionLabels` owns pprof mutation, and `aggregator.Pipeline.saveProfilingDocument` applies it only at the remote-storage boundary. The implementation validates string-table indices before mutation, preserves unrelated labels, clears stale managed labels, and introduces no new goroutine or resource lifecycle.

The end-to-end boundary is incomplete, however. The producer writes `tracer_data.flamedata.labels`, but `service.ProfileDocument`, selector translation, and the `LabelNames`/`LabelValues` APIs do not read or expose that field. This is a partial implementation of the stated query/aggregation contract: the dimensions survive in raw JSON and pprof samples but are unavailable to the supported service query path.

The simplest durable fix is to finish the existing metadata path rather than introduce another abstraction: carry the map through `ProfileDocument`, define a deterministic allowlisted mapping from public label names to indexed Elasticsearch fields, and cover the write/read/filter contract with a storage-level test. Legacy documents without `labels` can remain compatible because the map is optional.

### Review dimensions

- Copyright, Go style, file organization, and open-source hygiene: passed.
- Security: no injection, secret, path-traversal, crypto, or authentication-boundary findings; keys are fixed and values come from validated collection context.
- Concurrency and lifecycle: no new goroutines or owned resources; test cleanup is explicit.
- Compatibility: legacy profile decoding remains viable, but the new label query contract is missing.
- Performance: injection is linear in sample count with at most five managed labels per sample; no measured blocker found.

### Verification

Reviewed base `0242aa935cbe200710a30489c08f868d6a438615`, head `fa69958879d742d3b30cdaf6c7468218162061ed` (merge base `42847296a701a5e8535e004b515155ac566a1a78`).

- `git diff --check`: pass
- `go test -p 1 ./internal/profiler ./internal/profiler/context ./internal/profiler/aggregator`: pass at PR head
- `go vet -p 1` for the affected packages: pass
- `golangci-lint v1.64.8` for each affected package: pass
- Synthetic merge with current `main`: clean; profiler/context tests pass. Aggregator merge-state validation remains limited because current `main` requires generated `internal/bpf/abi` artifacts from the full BPF build.
- GitHub currently reports no configured check runs; race validation was not independently rerun because the shared runner ran out of disk space.

---
*Co-authored by: gpt-5.6-sol*

PR #482, opened 2026-07-29
fix(config): preserve numeric update types

messere1:
## Summary

- preserve floating-point config values submitted through PUT /config
- convert JSON numbers to signed and unsigned integer fields only when exact and in range
- reject fractional integer updates instead of truncating them
- add HTTP regression coverage for float, signed integer, and unsigned integer fields

## Why

JSON numbers in ConfigRequest.Config arrive as float64. The handler previously converted every numeric value to int64, so documented settings such as Runtime.StartupCPULimitCores = 0.5 could not be updated through the API, while fractional integer values were silently truncated.

## Testing

- make check
- go test ./internal/config ./cmd/huatuo-bamai/handlers -run ConfigHandler -count=1
- make unit (all packages pass except internal/cgroups/TestCgroupInterfaces/CpuUsage because this WSL environment does not expose cgroup v1 cpuacct.stat)

### DIFF
--- a/cmd/huatuo-bamai/handlers/config.go
+++ b/cmd/huatuo-bamai/handlers/config.go
@@ -1,4 +1,4 @@
-// Copyright 2025 The HuaTuo Authors
+// Copyright 2025, 2026 The HuaTuo Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -15,9 +15,7 @@
 package handlers
 
 import (
-	"math"
 	"net/http"
-	"reflect"
 
 	"huatuo-bamai/cmd/huatuo-bamai/config"
 	"huatuo-bamai/internal/log"
@@ -48,13 +46,6 @@ func (h *ConfigHandler) update(ctx *server.Context) error {
 	}
 
 	for k, v := range req.Config {
-		if reflect.ValueOf(v).Kind() == reflect.Float64 {
-			f := v.(float64)
-			if f > math.MaxInt64 || f < math.MinInt64 {
-				return response.ErrInvalidRequest.WithMessage("integer value out of range")
-			}
-			v = int64(f)
-		}
 		if err := config.Set(k, v); err != nil {
 			return response.ErrInvalidRequest.WithMessage(err.Error())
 		}
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -16,6 +16,7 @@ package config
 
 import (
 	"fmt"
+	"math"
 	"os"
 	"path/filepath"
 	"reflect"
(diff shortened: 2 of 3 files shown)


cmd/huatuo-bamai/handlers/config.go
```diff
@@ -48,13 +46,6 @@ func (h *ConfigHandler) update(ctx *server.Context) error {
...
 
 	for k, v := range req.Config {
-		if reflect.ValueOf(v).Kind() == reflect.Float64 {
-			f := v.(float64)
-			if f > math.MaxInt64 || f < math.MinInt64 {
-				return response.ErrInvalidRequest.WithMessage("integer value out of range")
-			}
-			v = int64(f)
-		}
 		if err := config.Set(k, v); err != nil {
```
huhong-web, line 49: [P1] Apply the request as one validated transaction.

`config.Set` mutates the shared `cfg` in place one map entry at a time, while `config.Get` readers do not take the same lock and `Sync` runs only after the loop. This patch newly makes float fields writable, so `{"Runtime.CPULimitCores":0}` can return 204 and persist a value that `RuntimeConfig.Validate` rejects on the next `Load`; a mixed valid/invalid batch can also return 400 after exposing a partial in-memory update, with the outcome depending on map iteration order.

Apply all values to a clone, validate the complete candidate, persist it, and only then publish it atomically. Add tests proving that invalid single-field and mixed batches leave both the live snapshot and file unchanged, and run the concurrent-reader case under `go test -race`.

internal/config/config.go
```diff
@@ -83,9 +84,55 @@ func Set(cfg any, key string, val any) error {
...
+func convertJSONNumber(dst, src reflect.Value) (reflect.Value, bool) {
+	if src.Kind() != reflect.Float64 {
+		return reflect.Value{}, false
+	}
+
+	f := src.Float()
+	converted := reflect.New(dst.Type()).Elem()
+	switch dst.Kind() {
+	case reflect.Int, reflect.Int8, reflect.Int16, reflect.Int32, reflect.Int64:
+		if math.Trunc(f) != f ||
```
huhong-web, line 109: [P2] Preserve the JSON number until the destination type is known.

By this point `map[string]any` has already decoded every number as `float64`. For example, JSON `9007199254740993` arrives here as `9007199254740992`, so the truncation and overflow checks accept it and silently store the wrong `int64`; conversely, `9223372036854775807` rounds to `2^63` and is rejected even though it is a valid `int64`.

Keep `json.RawMessage` or `json.Number` at the HTTP boundary and parse/decode directly into the destination type. Add boundary cases for `2^53+1`, `MaxInt64`, `MaxUint64`, fractional inputs, and each narrow integer width.

huhong-web (commented):
## Code Review

**Verdict: blocked; this head should not be merged.**

Reviewed immutable head `ee05995f2118c92d6fd5ab8744efa145affd3f24` against merge base `93c0d9af1d20e3ea29ae82805575ae845f3cada8`. The PR is currently conflicting with `main`, and the same goal has since been implemented on `main` by [`77640bec`](https://github.com/ccfos/huatuo/commit/77640bec25b235ec911e05c54e8e5de6149e37ba). Closing this PR as superseded is safer than resolving its conflicts.

### Architecture & Design

**Decision: needs changes / superseded.**

The affected path is HTTP JSON input → `handlers.ConfigRequest` → `config.Set` → reflective assignment → TOML persistence → runtime readers. This head loses the original JSON number before the destination type is known and mutates the process-wide configuration one field at a time. That leaves precision, validation, rollback, reader consistency, and persistence in different layers without one transaction owner.

The simpler design now present on `main` keeps each value as `json.RawMessage`, decodes it directly into the destination type, applies the whole batch to a clone, validates it, atomically persists it, and publishes an immutable snapshot. Its tests cover invalid-assignment rejection, batch rollback, and concurrent snapshot consistency. That is the appropriate ownership boundary and already satisfies this PR's stated goal.

### Security

**[P1, pre-existing prerequisite] Protect the configuration write route before expanding it.** At this head, `handlers.Start` registers `PUT /config` on the agent server without setting `RequireAuth`; the documented default listen address is `:19704`. This PR expands which fields a network client can mutate. Gate this route behind administrator authorization (or an equivalently explicit trusted boundary) and add an unauthorized-request test. This cause is outside the changed lines, so it is recorded here rather than as an inline comment.

### Other review dimensions

- License headers, diff hygiene, Go formatting, naming, error construction, file organization, and open-source compliance: no new issue found.
- Injection, path traversal, hard-coded secrets, and cryptography: no new issue found in the changed files.
- Tests cover small float/signed/unsigned updates and one fractional integer, but not lossless integer boundaries, invalid-domain rollback, or concurrent readers.

### Verification

- `git diff --check`: passed.
- `go test ./internal/config ...`: passed.
- `go vet ./internal/config`: passed.
- Repository `make check`: local lint stage could not run because installed `golangci-lint` v2 rejects this repository's pre-v2 config format.
- Handler test build: local dependency compilation exhausted the sandbox's temporary disk; no code failure was inferred.
- GitHub shows all six checks successful for this head, but those checks predate the current `main` and do not establish merge readiness after rebasing.

---
*Co-authored by: gpt-5.6-sol*

messere1:
Thanks for the review. I rechecked this PR against the current `main` (`7ee5c471`). Its original goal has been superseded by `77640bec`, which keeps request values as `json.RawMessage`, applies updates to a cloned configuration, validates the complete candidate, persists it, and only then publishes the immutable snapshot.

Rebasing this branch would duplicate and conflict with that implementation, so I am closing the PR as superseded instead. The configuration-route authorization observation is pre-existing and outside this PR's scope; it should be handled separately if still applicable.

PR #486, opened 2026-07-29
fix(profiling): prevent output file overwrite

### LINKED ISSUE #485: fix(profiling): prevent same-second output file overwrite
## Problem

`internal/profiler/aggregator.createOutputFile` builds output names from `time.Now().Unix()` and opens them with `os.Create`. Two profiling exports with the same prefix and extension in one second therefore resolve to the same path, and the later call truncates the earlier profile.

## Reproduction

1. Call `createOutputFile(dir, perf, .folded)` twice within one second.
2. Write data to the first file before creating the second.
3. Both calls can return the same filename, and the first file content is lost when the second call truncates it.

## Expected behavior

Every profiling export should create a distinct file without replacing an existing result.

## Proposed fix

- add an unpredictable suffix to timestamped output filenames;
- create files with `O_CREATE|O_EXCL` so an existing result is never silently truncated;
- add focused user-space tests covering unique filenames, preserved prior content, and output-directory creation.

This change is limited to Go user-space code and does not require privileged Docker, eBPF loading, or a specific kernel to validate.

twei43846-afk:
## Summary

- add a cryptographically random suffix to timestamped profiling output filenames
- create output files with `O_CREATE|O_EXCL` so an existing profile is never silently truncated
- add focused tests for unique names, preservation of earlier content, and nested output-directory creation

## Root cause and impact

`createOutputFile` used second-resolution `time.Now().Unix()` filenames together with `os.Create`. Two exports with the same prefix and extension in one second selected the same path, and the later export truncated the earlier result.

The change keeps the existing prefix, timestamp, and extension convention while making each output path unique and enforcing exclusive creation.

## Validation

- `go test -count=10 -run '^TestCreateOutputFile' export.go export_test.go` (passed on Windows)
- `git diff --check` (passed)
- package-level `go test -count=1 -run '^TestCreateOutputFile' ./internal/profiler/aggregator` is blocked during package setup because the checkout does not contain generated `internal/bpf/abi`; no test case ran or failed

This is pure Go user-space code. No privileged Docker, eBPF loading, or kernel-specific runtime validation is required.

Closes #485

### DIFF
--- a/internal/profiler/aggregator/export.go
+++ b/internal/profiler/aggregator/export.go
@@ -15,6 +15,7 @@
 package aggregator
 
 import (
+	"crypto/rand"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -65,10 +66,14 @@ func createOutputFile(dir, prefix, ext string) (*os.File, error) {
 		return nil, fmt.Errorf("failed to create output directory: %w", err)
 	}
 
-	fileName := fmt.Sprintf("%s_%d%s", prefix, time.Now().Unix(), ext)
-	filePath := filepath.Join(dir, fileName)
+	var suffix [8]byte
+	if _, err := rand.Read(suffix[:]); err != nil {
+		return nil, fmt.Errorf("failed to generate output filename: %w", err)
+	}
 
-	file, err := os.Create(filePath)
+	fileName := fmt.Sprintf("%s_%d_%x%s", prefix, time.Now().Unix(), suffix, ext)
+	filePath := filepath.Join(dir, fileName)
+	file, err := os.OpenFile(filePath, os.O_WRONLY|os.O_CREATE|os.O_EXCL, 0o666)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create output file: %w", err)
 	}
--- a/internal/profiler/aggregator/export_test.go
+++ b/internal/profiler/aggregator/export_test.go
@@ -1,0 +1,84 @@
+// Copyright 2026 The HuaTuo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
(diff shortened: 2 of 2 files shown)


twei43846-afk:
CI workflows are awaiting maintainer approval for a first-time contributor (`action_required`). The focused user-space test passes locally: `go test -count=10 -run '^TestCreateOutputFile' export.go export_test.go`. Would a maintainer please approve the workflow runs when convenient?

huhong-web (commented):
## Code Review

LGTM. I did not find any actionable issues in this PR.

Scope reviewed:
- `internal/profiler/aggregator/export.go`
- `internal/profiler/aggregator/export_test.go`

Checks performed:
- Copyright header: passed for the new test file.
- Go style and design: the change is scoped and keeps the existing API; `crypto/rand` plus `O_CREATE|O_EXCL` addresses same-second overwrite/truncation without introducing shared mutable state.
- Security review: no SQL/command injection, hardcoded secret, path traversal, or crypto misuse found in the changed code. The filename suffix uses `crypto/rand`, and `prefix`/`ext` remain internal call-site constants.
- Tests: new tests cover unique names, preserving prior content, and nested directory creation.

Local verification:
- `git diff --check origin/main...HEAD` passed.
- `GOCACHE=/tmp/go-build-cache go test -count=10 -run '^TestCreateOutputFile' export.go export_test.go` passed.
- `GOCACHE=/tmp/go-build-cache go test -count=1 -run '^TestCreateOutputFile' ./internal/profiler/aggregator` was blocked during package setup because generated package `huatuo-bamai/internal/bpf/abi` is absent in this checkout.
- `GOCACHE=/tmp/go-build-cache go vet ./internal/profiler/aggregator` was blocked by the same missing generated package.

---
*Co-authored by: gpt-5.5*

hao022:
@twei43846-afk 感谢修复该问题！

PR #488, opened 2026-07-29
test(tracing): cover document store mapper

Loyal-Young:
## Summary

- Add JSON round-trip coverage for `DocumentStoreMapper`.
- Verify storage fields normalize valid and fallback timestamps to UTC.

## Why

The mapper is the boundary between tracing documents and storage backends, so its serialization and time-field contract need regression coverage.

## Validation

- `git show --check`

### DIFF
--- a/pkg/tracing/document_test.go
+++ b/pkg/tracing/document_test.go
@@ -1,0 +1,93 @@
+// Copyright 2026 The HuaTuo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package tracing
+
+import (
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestDocumentStoreMapperEncodeDecode(t *testing.T) {
+	mapper := DocumentStoreMapper{}
+	want := &Document{
+		Hostname:     "node-a",
+		Region:       "cn-east",
+		UploadedTime: time.Date(2026, time.July, 23, 10, 30, 0, 0, time.UTC),
+		TracerID:     "trace-1",
+		TracerName:   "iotracing",
+		TracerData:   "ready",
+	}
+
+	encoded, err := mapper.Encode(want)
+	if err != nil {
+		t.Fatalf("Encode() error = %v", err)
+	}
+	got, err := mapper.Decode(encoded)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


pkg/tracing/document_test.go
```diff
@@ -0,0 +1,93 @@
...
+		"container_hostname":       "worker",
+		"container_host_namespace": "default",
+		"container_type":           "containerd",
+		"container_qos":            "burstable",
+		"tracer_name":              "iotracing",
+		"tracer_id":                "trace-1",
+		"tracer_time":              fallback.UTC(),
+		"tracer_type":              TracerRunTypeEvent,
+	}
+	if diff := cmp.Diff(want, got); diff != "" {
```
huhong-web, line 90: [P2] Assert the UTC location explicitly. `cmp.Diff` invokes `time.Time.Equal`, which compares instants but ignores the `Location` representation. As a result, `wantTime.UTC()` and `fallback.UTC()` compare equal to the original `+0800` values: changing `tracingDocumentTimeValue` from `return parsed.UTC()` / `fallback.UTC()` to `return parsed` / `fallback` would still leave this test green while breaking the stated UTC-normalization contract. After the map comparison, type-assert `got["time"]` and `got["tracer_time"]` to `time.Time` and verify `Location() == time.UTC` (or compare an explicitly formatted representation). Acceptance criterion: removing either relevant `.UTC()` call must make this test fail.

huhong-web (commented):
## Code Review

I found one actionable test-contract issue; it is documented inline.

### Architecture & Design

**Verdict: needs changes.** The change is correctly confined to the owning `pkg/tracing` package and exercises the pure `DocumentStoreMapper` boundary (`Document` -> mapper -> `storage.Store` -> backend). It introduces no production API, abstraction, dependency-direction, lifecycle, concurrency, or resource-ownership change. A mapper unit test is the simplest appropriate verification layer; a real-backend integration test is not required for JSON conversion or UTC normalization itself.

The remaining design risk is verification quality: the current comparison cannot observe whether the mapper actually returns UTC locations, so the PR does not yet prove its stated normalization contract. The smallest fix is to assert the `Location` of the `time` and `tracer_time` values explicitly.

### Other checks

- Scope: all changed hunks are direct verification work; no unrelated changes.
- Copyright and style: the new file has the required Apache 2.0 header; `gofmt -d` and `git diff --check` are clean.
- Security: no trust boundary, external input, secret, filesystem, command, SQL, crypto, or authorization behavior changes.
- Validation at HEAD `7d42f75b79a441d199a0d22723761e2276c6d2b1`: `go test ./pkg/tracing` passed. GitHub Build/Lint, PR-name lint, and Ubuntu 22.04/24.04/26.04 jobs passed. The Ubuntu 20.04 job failed in the unrelated metrics exclude-filter integration test because an expected `eth0` metric was absent; it should be rerun or triaged separately before merge. Local vet/race follow-up was limited by review-runner disk capacity, not a code diagnostic.

---
*Co-authored by: gpt-5.6-sol*

pkg/tracing/document_test.go
Loyal-Young: Addressed in 82d6a1a9.
Added explicit time.Time assertions for both time and tracer_time, including Location() == time.UTC, so removing either normalization call now makes the test fail.

PR #489, opened 2026-07-29
test(storage): cover driver value utilities

Loyal-Young:
## Summary

- Add table-driven coverage for storage driver value helpers.
- Cover time normalization, `IN` values, context fallback, byte cloning, and string conversion.

## Why

These helpers are shared by storage backends and need predictable behavior for query values and defensive copies.

## Validation

- `git show --check`

### DIFF
--- a/internal/storage/driver/utils_test.go
+++ b/internal/storage/driver/utils_test.go
@@ -1,0 +1,89 @@
+// Copyright 2026 The HuaTuo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package driver
+
+import (
+	"context"
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+)
+
+func TestDriverValueUtilities(t *testing.T) {
+	timestamp := time.Date(2026, time.July, 23, 12, 0, 0, 0, time.FixedZone("CST", 8*60*60))
+	tests := []struct {
+		name    string
+		input   any
+		want    []any
+		wantErr error
+	}{
+		{name: "normalizes times in slices", input: []time.Time{timestamp}, want: []any{"2026-07-23 04:00:00.000 +0000"}},
+		{name: "keeps scalar slice values", input: []string{"a", "b"}, want: []any{"a", "b"}},
+		{name: "rejects scalar", input: "a", wantErr: ErrInRequiresSlice},
+		{name: "rejects empty slice", input: []string{}, wantErr: ErrInRequiresNonEmpty},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got, err := FlattenInValues(tt.input)
+			if !errors.Is(err, tt.wantErr) {
+				t.Fatalf("FlattenInValues() error = %v, want %v", err, tt.wantErr)
+			}
+			if diff := cmp.Diff(tt.want, got); diff != "" {
+				t.Errorf("FlattenInValues() mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+
+	if WithContext(nil) == nil {
+		t.Error("WithContext(nil) returned nil")
+	}
+	ctx := context.WithValue(context.Background(), "key", "value")
+	if WithContext(ctx) != ctx {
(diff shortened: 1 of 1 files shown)


huhong-web (commented):
## Code Review

Reviewed the immutable diff from merge base `0ed6c0f974c0bd722ceae28254a1dc9c0403b193` to head `d43f4d2f5681378f3d1a4ffec3b940952531d5e2`. The PR goal is to add focused unit coverage for the storage driver's value/context/copy helpers.

**Verdict: changes required.** I found 3 actionable issues; details and acceptance criteria are in the inline comments. The current `Build, Lint and Vendor` check is failing on two of them.

### Architecture & Design

**Verdict: reasonable.** The tests stay with the backend-neutral helpers owned by `internal/storage/driver`. The affected call chain remains coherent: `internal/storage/store.go` applies `WithContext` before backend calls; SQLite uses `NormalizeValue`/`FlattenInValues` while building parameterized queries in `internal/storage/sqlite/sql.go`; Elasticsearch uses the same normalization in `internal/storage/elasticsearch/dsl.go` and uses `CloneBytes`/`StringValue` when decoding backend responses. This PR changes no production API, data model, configuration, registry, storage schema, dependency direction, mutable state, goroutine, or resource lifecycle.

Focused unit tests are the simplest viable verification layer for these pure helpers; a new abstraction, mock, or integration harness would add no useful boundary coverage here. Existing backend tests remain responsible for real SQLite/Elasticsearch query and mapping behavior. The remaining design risk is test fidelity: the byte-copy test currently does not prove content preservation.

### Code Quality & Security

- The new file has the required Apache 2.0 header and passes `gofmt` and `git diff --check`.
- No new trust boundary or production code is introduced. I traced the helpers into SQLite's placeholder-based query construction and Elasticsearch's typed request DSL; there is no new SQL/command injection, path traversal, secret, cryptographic, or authorization surface.
- No interface, global state, concurrency, error-sentinel, backend naming, or open-source compliance issue was introduced.

### Verification

- `go test -count=1 ./internal/storage/driver` — pass
- `go test -race -count=1 ./internal/storage/driver` — pass
- `go test -cover -count=1 ./internal/storage/driver` — pass, 100.0% statements
- `go vet ./internal/storage/driver` — pass
- `gofmt -d internal/storage/driver/utils_test.go` — clean
- `git diff --check 0ed6c0f...d43f4d2f` — pass
- GitHub `Build, Lint and Vendor` — fail: `SA1012` at line 52 and `SA1029` at line 55

The branch is also 227 commits behind the reviewed base tip, so it should be rebased and all checks rerun after the inline findings are fixed.

---
*Co-authored by: gpt-5.6-sol*

internal/storage/driver/utils_test.go
```diff
@@ -0,0 +1,89 @@
...
+			if !errors.Is(err, tt.wantErr) {
+				t.Fatalf("FlattenInValues() error = %v, want %v", err, tt.wantErr)
+			}
+			if diff := cmp.Diff(tt.want, got); diff != "" {
+				t.Errorf("FlattenInValues() mismatch (-want +got):\n%s", diff)
+			}
+		})
+	}
+
+	if WithContext(nil) == nil {
```
huhong-web, line 52: [P2] Keep the intentional nil-fallback test without failing the required lint gate. GitHub CI reports `SA1012` on this exact call, so the current head cannot pass `Build, Lint and Vendor`. Because `WithContext` explicitly documents support for nil, add a narrowly scoped suppression with its reason, for example `//nolint:staticcheck // Intentionally pass nil to verify the documented fallback.` Acceptance: `make golangci-lint` passes while this fallback remains exercised.

internal/storage/driver/utils_test.go
```diff
@@ -0,0 +1,89 @@
...
+		t.Error("WithContext(nil) returned nil")
+	}
+	ctx := context.WithValue(context.Background(), "key", "value")
```
huhong-web, line 55: [P2] Use a context that does not violate the context-key contract. CI reports `SA1029` because a built-in string is used as the key, leaving the required lint check red. This test only needs a distinct context identity, so `ctx, cancel := context.WithCancel(t.Context())` plus `defer cancel()` is simpler; if a value is essential, define a private key type. Acceptance: lint passes and `WithContext(ctx) == ctx` remains asserted.

internal/storage/driver/utils_test.go
```diff
@@ -0,0 +1,89 @@
...
+	if WithContext(ctx) != ctx {
+		t.Error("WithContext(ctx) did not preserve the supplied context")
+	}
+
+	data := []byte{1, 2, 3}
+	clone := CloneBytes(data)
```
huhong-web, line 61: [P2] Assert that the clone preserves the original bytes before testing aliasing. As written, an implementation such as `return make([]byte, len(data))` would pass: mutating that zero-filled slice would not affect `data`, even though no content was copied. Add `cmp.Diff(data, clone)` or `bytes.Equal(data, clone)` before `clone[0] = 9`, then keep the mutation check. Acceptance: a same-length zero-filled result fails while the current copy passes.

hao022:
@Loyal-Young 增加这个测试的意义不是很大。可以重点放到特性开发。

PR #490, opened 2026-07-29
test(profiler): cover task comm conversion

Loyal-Young:
## Summary

- Add table-driven tests for converting fixed-width BPF task names to Go strings.
- Cover NUL-padded, empty, and full-width process names.

## Why

Process names arrive from BPF as fixed-size byte arrays; truncation and NUL handling are part of the public profiler behavior.

## Validation

- `git show --check`

### DIFF
--- a/internal/profiler/procutil/procutil_test.go
+++ b/internal/profiler/procutil/procutil_test.go
@@ -1,0 +1,44 @@
+// Copyright 2026 The HuaTuo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package procutil
+
+import (
+	"strings"
+	"testing"
+
+	"huatuo-bamai/internal/bpf"
+)
+
+func TestCommToString(t *testing.T) {
+	tests := []struct {
+		name  string
+		input string
+		want  string
+	}{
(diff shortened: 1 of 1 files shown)


huhong-web (commented):
## Code Review

LGTM. I did not find any issues that need inline comments.

Checked:
- New test file has the required Apache 2.0 copyright header.
- The table-driven cases cover NUL-padded, empty, and full-width task comm values.
- The test uses the project module path `huatuo-bamai/internal/bpf` and the shared `bpf.TaskCommLen` constant instead of duplicating the size.
- No new production code, goroutines, storage behavior, filesystem/network access, SQL/command execution, crypto, or secret handling was introduced.
- GitHub checks are passing: Build/Lint/Vendor and VM tests for ubuntu20.04/22.04/24.04/26.04.

Local note: I could not clone the PR branch in this workspace because the GitHub TLS handshake failed, so I relied on `gh` PR metadata/diff/API reads plus the passing CI checks.

---
*Co-authored by: gpt-5.5*

internal/profiler/procutil/procutil_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+package procutil
+
+import (
+	"strings"
+	"testing"
+
+	"huatuo-bamai/internal/bpf"
+)
+
```
huhong-web, line 23: [P2] This new package test file only exercises `CommToString`, but `procutil.go` also exposes `CheckExecPath` and it remains untested. Please add table-driven coverage for `CheckExecPath` here as well: a success case using the current process resolved `/proc/<pid>/exe` target, a mismatch case that expects an error, and a readlink failure case such as an invalid/nonexistent pid. That will cover both error branches and the success path.

huhong-web (changes requested):
## Code Review Update

I revisited the test scope after the coverage concern. I agree this should not be LGTM yet: this new `procutil_test.go` only covers `CommToString`, while the same package file also exposes `CheckExecPath` and still has no direct coverage.

Please add focused tests for `CheckExecPath`, then this should be ready to re-review.

---
*Co-authored by: gpt-5.5*

PR #502, opened 2026-08-03
fix(bpf): disable perf events until BPF attachment

huhong-web:
## What changed

- Set PerfBitDisabled when constructing frequency and period perf event attrs.
- Keep the existing SET_BPF then ENABLE ordering.
- Raise the period sampling test value from 1 to 1,000,000.

## Root cause

Period mode cleared attr.Bits, so CPU clock sampling with period 1 could start
inside perf_event_open. In a KVM guest this generated enough sampling pressure
to block the syscall before the BPF program could be attached. The unit test
then hit its five-minute timeout.

## Validation

- make check
- go test ./internal/bpf -run Test(AttachPerfEvent|OpenPerfEvent) -count=1 -timeout=2m

The original period test completed locally before the fix, so the VM-specific
hang was confirmed from Actions run 30790326587, job 91665106516.

### DIFF
--- a/internal/bpf/perf_event_pmu.go
+++ b/internal/bpf/perf_event_pmu.go
@@ -104,12 +104,12 @@ func attachPerfEvent(opt *perfEventOption) (*perfEventAttach, error) {
 		Type:   unix.PERF_TYPE_SOFTWARE,
 		Size:   unix.PERF_ATTR_SIZE_VER0,
 		Config: unix.PERF_COUNT_SW_CPU_CLOCK,
-		Bits:   unix.PerfBitFreq,
+		Bits:   unix.PerfBitFreq | unix.PerfBitDisabled,
 		Sample: opt.sample,
 	}
 
 	if opt.sampleMode == perfEventSamplePeriod {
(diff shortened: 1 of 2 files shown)


internal/bpf/perf_event_pmu.go
```diff
@@ -104,12 +104,12 @@ func attachPerfEvent(opt *perfEventOption) (*perfEventAttach, error) {
...
 		Size:   unix.PERF_ATTR_SIZE_VER0,
 		Config: unix.PERF_COUNT_SW_CPU_CLOCK,
-		Bits:   unix.PerfBitFreq,
+		Bits:   unix.PerfBitFreq | unix.PerfBitDisabled,
 		Sample: opt.sample,
 	}
 
 	if opt.sampleMode == perfEventSamplePeriod {
-		attr.Bits = 0
+		attr.Bits = unix.PerfBitDisabled
```
hao022, line 112: 从编码多度直接赋值，这里不合适，应该实现去除位unix.PerfBitFreq；这样语义比较明确。

PR #504, opened 2026-08-03
feat(oom): enrich victim process diagnostics

sbtger:
## Summary

This PR improves OOM diagnostics by preserving more information about the selected victim before it exits. It adds runtime-language detection, memory-footprint fields, bounded command-line and environment capture, and best-effort container association.

The primary OOM event and the victim exit context are correlated asynchronously, so waiting for `do_exit` does not block the main OOM event reader. If exit-context capture or container lookup fails, Huatuo still stores the base OOM event.

Related to #333. This PR covers the victim-context requirements but does not include OOM CPU profiling.

## Changes

- Add a focused `pkg/processlang` detector for Java, Go, Python, and Node.
  - Inspect live executable metadata and mapped runtime libraries.
  - Fall back to captured command-line and command-name metadata.
- Report the victim runtime language in `language_info.victim`.
- Report the victim memory footprint:
  - `rss_anon_bytes`
  - `rss_file_bytes`
  - `rss_shmem_bytes`
  - `total_vm_bytes`
- Capture bounded victim process context at `do_exit`:
  - `cmdline` and `cmdline_truncated`
  - `environ` and `environ_truncated`
- Correlate the base OOM event and exit event by victim TGID and event timestamp.
- Keep OOM processing available when exit-context collection or the final container lookup is unavailable.
- Improve Go detection with two complementary fallbacks:
  - Retain `/proc/<pid>/exe` early and inspect Go build information after the existing snapshots are collected.
  - Check the Go buildinfo marker at `mm->start_data` when the registered victim enters `do_exit`.
- Support both `rss_stat` layouts through CO-RE and use probe-read helpers available on CentOS 8.2.

## Event processing

1. Huatuo receives the base `oom_kill_process` event and immediately retains the victim executable when possible.
2. Existing cgroup, container, and host memory snapshots are collected.
3. Runtime-language detection first uses the retained executable and live process metadata.
4. The main event processing continues asynchronously while the victim reaches `do_exit`.
5. The exit hook captures command-line and environment data and emits a correlated exit event.
6. Huatuo merges the optional exit context, retries unknown language detection from Go buildinfo/cmdline/comm, and writes the final OOM JSON.

## Compatibility and failure behavior

- The added context is best-effort; missing exit context does not discard the base OOM event.
- Missing or expired container metadata leaves `container_id` empty but does not prevent event storage.
- Command-line and environment buffers are bounded and expose truncation flags.
- No kernel behavior is delayed or changed: the BPF hooks only observe the OOM and exit paths.

## Validation

- `make bpf-build`
- `go build -o _output/bin/huatuo-bamai ./cmd/huatuo-bamai`
- `go test ./pkg/processlang ./core/events ./pkg/tracing`
- Real memory-cgroup OOM tests on the 5.10 VM for Go, Java, Python, and Node:
  - All four victims were killed by the kernel with exit status 137.
  - All four languages were detected correctly (`go`, `java`, `python`, and `node`).
  - Command line, environment, RSS/virtual-memory footprint, and the existing host memory snapshot were present in the saved OOM events.

## Scope

OOM CPU profiling is intentionally not part of this PR and should be handled separately.

### DIFF
--- a/bpf/oom.c
+++ b/bpf/oom.c
@@ -145,6 +145,32 @@ capture_user_region(u8 *dst, u32 capacity, unsigned long start,
 	return (u16)length;
 }
 
+/*
+ * Go places this buildinfo magic at mm->start_data for normal internal-link
+ * executables. Check only that fixed address; do not scan victim memory.
+ */
+#define GO_BUILDINFO_MAGIC "\xff Go buildinf:"
+#define GO_BUILDINFO_MAGIC_LEN 14
+
+static __always_inline u8
+has_go_buildinfo_at_start_data(unsigned long start_data)
+{
+	u8 actual[GO_BUILDINFO_MAGIC_LEN];
+	u8 expected[] = GO_BUILDINFO_MAGIC;
+	int i;
+
+	if (!start_data ||
+	    compat_bpf_probe_read(actual, sizeof(actual), (void *)start_data) != 0)
+		return 0;
+
+#pragma unroll
+	for (i = 0; i < GO_BUILDINFO_MAGIC_LEN; i++) {
+		if (actual[i] != expected[i])
+			return 0;
+	}
+	return 1;
+}
+
 SEC("kprobe/oom_kill_process")
 int BPF_KPROBE(oom_kill_process, struct oom_control *oc, const char *message)
 {
@@ -222,7 +248,7 @@ int BPF_KPROBE(do_exit_trace, long code)
 	struct oom_exit_event *exit_event;
 	struct task_struct *victim_task;
 	struct mm_struct *victim_mm;
-	unsigned long arg_start, arg_end, env_start, env_end;
+	unsigned long arg_start, arg_end, env_start, env_end, start_data;
 
 	/*
 	 * Most exits stop at this global-data check. The hash lookup is only paid
@@ -264,6 +290,9 @@ int BPF_KPROBE(do_exit_trace, long code)
 	arg_end = BPF_CORE_READ(victim_mm, arg_end);
 	env_start = BPF_CORE_READ(victim_mm, env_start);
 	env_end = BPF_CORE_READ(victim_mm, env_end);
+	start_data = BPF_CORE_READ(victim_mm, start_data);
+	exit_event->go_build_info =
+	    has_go_buildinfo_at_start_data(start_data);
 
 	exit_event->cmdline_len = capture_user_region(
 	    exit_event->victim_cmdline, sizeof(exit_event->victim_cmdline),
--- a/bpf/oom.c
+++ b/bpf/oom.c
@@ -1,3 +1,19 @@
+/*
+ * Copyright 2026 The HuaTuo Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 #include "vmlinux.h"
 
 #include <bpf/bpf_core_read.h>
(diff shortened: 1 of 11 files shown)


core/events/oom.go
```diff
@@ -41,13 +43,22 @@ type OOMActor struct {
 	ContainerHostname   string                   `json:"container_hostname,omitempty"`
 	Pid                 int32                    `json:"pid"`
 	Comm                string                   `json:"comm"`
+	Cmdline             string                   `json:"cmdline,omitempty"`
+	CmdlineTruncated    bool                     `json:"cmdline_truncated,omitempty"`
+	Environ             []string                 `json:"environ,omitempty"`
```
huhong-web, line 48: [P1] This persists the victim process environment in the OOM tracing JSON. Environment variables commonly contain secrets such as API tokens, DB passwords, cloud credentials, and internal endpoints; capturing them by default turns an OOM diagnostic into a credential disclosure path. Please either remove `environ` from the persisted event, make it explicitly opt-in behind configuration, or redact by default with a small allowlist/denylist before calling `tracing.Save`.

huhong-web (changes requested):
## Code Review

I found one blocking issue in the new OOM diagnostics path. The PR captures the victim process environment and persists it in the tracing payload without redaction or an opt-in gate, which can leak secrets such as tokens, passwords, and connection strings into diagnostic storage.

Validation performed:
- `GOCACHE=/tmp/go-build-cache-pr504 GOMODCACHE=/tmp/go-mod-cache-pr504 make bpf-build` passed.
- `pkg/processlang` passed as part of the targeted `go test` run before the broader run was interrupted by local `/tmp` space exhaustion.
- Broader `go test ./pkg/processlang ./core/events ./pkg/tracing` could not be fully completed in this sandbox because generated packages/cache space were unavailable or exhausted.

---
*Co-authored by: gpt-5.5*

core/events/oom.go
sbtger: Addressed. Sensitive environment values are now redacted by default before entering the exit cache and again before tracing.Save(), while diagnostic settings such as GOMEMLIMIT, GOGC, and GOMAXPROCS are preserved.
Unit tests and a real OOM test confirmed that sensitive values are masked as [REDACTED] and never appear in persisted output.

bpf/oom.c
```diff
@@ -49,7 +216,93 @@ int BPF_KPROBE(oom_kill_process, struct oom_control *oc, const char *message)
...
+	 */
+	if (bpf_map_delete_elem(&oom_victims, &victim_tgid) != 0)
+		return 0;
+	__sync_fetch_and_add(&oom_victim_count, -1);
+
+	exit_event = bpf_map_lookup_elem(&oom_exit_event_buf, &zero);
+	if (!exit_event)
+		return 0;
+
+	exit_event->timestamp = event_ts;
```
huhong-web, line 289: [P1] `exit_event` is a per-CPU array slot reused across OOM victims, but this path sends `sizeof(*exit_event)` to the perf ring. If the current capture is shorter or fails, bytes beyond `cmdline_len` / `environ_len` can still contain the previous victim's cmdline or environment and will be copied to userspace in the raw perf record. Please clear the whole struct (or at least lengths, flags, and both buffers) immediately after `bpf_map_lookup_elem` succeeds and before setting `timestamp`, for example with a verifier-friendly memset/zeroing pattern, so stale sensitive bytes cannot leave the BPF map.

huhong-web (changes requested):
## Code Review

I found one remaining issue in the BPF exit-context path; details are inline.

Validation performed:
- `GOCACHE=/tmp/go-build-cache go test ./pkg/processlang ./core/events ./pkg/tracing`: `pkg/processlang` and `pkg/tracing` passed; `core/events` could not complete because this sandbox is missing generated capnp transport code after `make gen-build` failed at `capnp compile: -I /std: no such directory`.
- `GOCACHE=/tmp/go-build-cache make gen-build`: BPF ABI generation progressed, then failed at the repository capnp generation step for the same local capnp stdlib path issue.
- `git diff --check origin/main...HEAD` passed.

---
*Co-authored by: gpt-5.5*

huhong-web (approved):
## Code Review

Re-reviewed the latest head `ba7556bd59743defd8f7a37a1b18312874d1b003`. The previous findings have been addressed: environment values are redacted before persistence, and the reused per-CPU `oom_exit_event` buffer is cleared before the full struct is emitted to userspace. I did not find new blocking issues in the updated diff.

Validation performed:
- `GOCACHE=/tmp/go-build-cache-pr504-rereview make bpf-build` passed.
- `GOCACHE=/tmp/go-build-cache-pr504-rereview go test ./pkg/processlang` passed.
- `git diff --check origin/main...HEAD` passed.
- `GOCACHE=/tmp/go-build-cache-pr504-rereview go test ./pkg/processlang ./core/events ./pkg/tracing` could not complete in this sandbox because `core/events` build depends on missing generated capnp transport symbols (`NewRootMessage`, `Chunk`, etc.); the run was interrupted after the build failure surfaced.

---
*Co-authored by: gpt-5.5*

bpf/oom.c
sbtger: Fixed by resetting the entire per-CPU exit_event buffer before each capture, preventing stale cmdline or environment bytes from reaching userspace.

sbtger:
The CI test failure is unrelated to the changes introduced in this PR.

PR #505, opened 2026-08-04
feat(net_rx_latency): support IPv6 alongside IPv4

123123213weqw:
Split out of #346 per @hao022's review request: submit the IPv6 feature standalone first, other features will follow as separate PRs.

## What this PR does

`net_rx_latency` now traces TCP RX latency for **IPv6 alongside IPv4**:

- **bpf**: addresses are stored in 16-byte fields so one layout serves both families (IPv4 uses the low 4 bytes, IPv6 all 16); `addr_family` (AF_INET / AF_INET6) tells userspace how to format them. The protocol-agnostic tracepoints (`netif_receive_skb`, `skb_copy_datagram_iovec`) accept both families via `skb_tcp_family()`.
- **bpf**: new `tcp_v6_rcv` kprobe mirrors `tcp_v4_rcv`; userspace attaches it only when the kernel exports the symbol (`HasKprobeFunction("tcp_v6_rcv")`), so CONFIG_IPV6-less kernels keep working (2 of 3 stages still collected).
- **abi**: `net_rx_latency_event` in `bpf/include/abi/net_rx_latency_types.h` carries the new layout; userspace decodes via the generated `abi.NetRXLatencyEvent` type.
- **core/events**: family-aware address formatting (`eventAddrs`), plus the review fixes from #346: bounds-guarded `lat_stage` lookup (`lookupLatStage` skips corrupt values instead of panicking) and the stage rename `RX_STAGE_TCPV4` -> `RX_STAGE_TCP` (the stage is now family-neutral).
- **integration**: new `test_net_rx_latency_ipv6.sh` (+ user-space receiver testdata), hardened against VM timing issues.

## Rebase note

Rebased onto current `main`, which moved perf-event layouts to generated ABI types: the struct change lives in the ABI header now, and userspace uses `abi.NetRXLatencyEvent` instead of a handwritten mirror.


integration/test_net_rx_latency_ipv6.sh
```diff
@@ -0,0 +1,156 @@
...
+
+ip link add "${VETH_HOST}" type veth peer name "${VETH_PEER}" 2> /dev/null || skip "veth creation failed"
+# Disable DAD so addresses are usable immediately (DAD can take ~1 s on veth).
+sysctl -qw "net.ipv6.conf.${VETH_HOST}.accept_dad=0" 2> /dev/null || true
+sysctl -qw "net.ipv6.conf.${VETH_PEER}.accept_dad=0" 2> /dev/null || true
+ip addr add "${VETH_HOST_IP6}/64" dev "${VETH_HOST}" 2> /dev/null || {
+	ip link del "${VETH_HOST}" 2> /dev/null || true
+	skip "IPv6 unavailable on host"
+}
+ip addr add "${VETH_PEER_IP6}/64" dev "${VETH_PEER}" 2> /dev/null || true
```
huhong-web, line 55: [P1] Exercise the real veth receive path instead of the host local route.

Both IPv6 addresses are assigned in the initial network namespace, and the server is also started there at line 108. Linux therefore installs local routes for both addresses, so the curl traffic can be delivered through the local table without crossing this veth pair. This makes the test pass without validating the advertised veth/NAPI path. The evidence is visible in every QEMU job: each reports six matching IPv6 events, but every matching event is `RX_STAGE_USERCOPY`; there are no NETIF or TCP-stage events.

Mirror `test_net_rx_latency.sh`: create a unique netns, move `VETH_PEER` into it, configure/disable DAD inside that namespace, run the server with `ip netns exec`, and delete the namespace during cleanup. Also add a stage-specific assertion or dedicated attachment/packet-path check so the test fails when the `tcp_v6_rcv` or protocol-agnostic receive path is broken, rather than accepting any IPv6 user-copy event.

bpf/net_rx_latency.c
```diff
@@ -97,15 +97,31 @@ static inline u64 delta_now_skb_tstamp(struct sk_buff *skb)
...
+	}
+
+	if (proto == bpf_ntohs(ETH_P_IPV6)) {
+		struct ipv6hdr h;
+
+		bpf_probe_read(&h, sizeof(h), skb_network_header(skb));
+		// NOTE: nexthdr check does not walk extension header chains;
+		// IPv6+EH TCP packets are caught by the tcp_v6_rcv
+		// kprobe instead.
+		return (h.nexthdr == IPPROTO_TCP) ? AF_INET6 : 0;
```
huhong-web, line 121: [P2] Preserve the protocol-agnostic stages for TCP packets with IPv6 extension headers.

Checking only the base header's `nexthdr` drops valid IPv6 TCP packets that carry hop-by-hop, routing, destination-options, or fragment headers from both `netif_receive_skb_prog` and `skb_copy_datagram_iovec_prog`. The comment's `tcp_v6_rcv` fallback is not equivalent: that probe is optional and can only produce `RX_STAGE_TCP`, so NETIF/USERCOPY coverage is always lost and kernels without an attachable symbol can lose all coverage for these packets.

Use a verifier-bounded extension-header walker (with explicit fragment handling), or explicitly narrow the feature contract and document/test the partial-stage behavior. An IPv6 TCP fixture with at least one supported extension header should verify the intended outcome.

core/events/net_rx_latency.go
```diff
@@ -61,12 +63,50 @@ type NetTracingData struct {
...
+// eventAddrs formats the 16-byte src/dst of a net_rx_latency_event according
+// to AddrFamily. IPv4 occupies the low 4 bytes (network order); IPv6 uses all
+// 16.
+func eventAddrs(pd *abi.NetRXLatencyEvent) (family, saddr, daddr string) {
+	if pd.AddrFamily == afINET6 {
+		return addrFamilyV6,
+			netutil.Inetv6Ntop(pd.TCPSaddr).String(),
+			netutil.Inetv6Ntop(pd.TCPDaddr).String()
+	}
+	return addrFamilyV4,
```
huhong-web, line 86: [P2] Reject unknown address-family values instead of silently labeling them IPv4.

Every value other than `AF_INET6` currently falls through to the IPv4 decoder. A zero, corrupt, or future ABI value would therefore be stored as a plausible `address_family: "ipv4"` event with misleading addresses, while the adjacent stage decoder correctly fails closed on unknown values.

Switch explicitly on `afINET` and `afINET6`, return an `ok`/error result for anything else, and log+skip that perf record. Add cases for 2, 10, 0, and an arbitrary unknown value.

docs/key-feature/instant-observability_en.md
```diff
@@ -211,7 +211,7 @@ All event records include the following common fields:
 
 - **comm**: Name of the process that triggered the event
 - **pid**: Process ID that triggered the event
-- **lat_stage**: Stage where latency occurred (`RX_STAGE_NETIF` driver-to-kernel / `RX_STAGE_TCPV4` kernel-to-TCP / `RX_STAGE_USERCOPY` TCP-to-user-space)
+- **lat_stage**: Stage where latency occurred (`RX_STAGE_NETIF` driver-to-kernel / `RX_STAGE_TCP` kernel-to-TCP / `RX_STAGE_USERCOPY` TCP-to-user-space)
```
huhong-web, line 214: [P2] Complete the public event-schema documentation for this contract change.

`NetTracingData` now emits `address_family`, and addresses may be IPv6, but this sample and field list only rename the TCP stage. The Chinese counterpart has the same omission. Consumers following these docs cannot discover the new field or its allowed values.

Add `address_family` to both JSON examples and both field lists, document `ipv4`/`ipv6` and the corresponding address representation, and—if the extension-header behavior remains intentionally partial—state which latency stages are available in that case.

huhong-web (changes requested):
## Code review

I found four actionable issues; the receive-path test gap is blocking because it makes the principal IPv6 integration evidence pass without exercising the topology it claims to cover. The PR is also currently conflicting and 128 commits behind `main`, so the successful checks from 2026-08-04 do not establish compatibility with the current target branch.

### Architecture & Design

**Verdict: needs changes.**

The core boundary is otherwise cohesive: `bpf/include/abi/net_rx_latency_types.h` defines a family-tagged 16-byte address ABI, generated `abi.NetRXLatencyEvent` carries it into `core/events/net_rx_latency.go`, and `NetTracingData` persists the decoded contract. The loader's lifecycle is also reasonable: `EventPipeByName`, conditional `tcp_v6_rcv` discovery, `AttachWithOptions`, reader cleanup, and context-driven detach keep ownership explicit without adding a new abstraction.

The incomplete parts are cross-layer verification and compatibility: the new IPv6 test keeps both veth endpoints in one namespace and therefore uses the host local route; all four QEMU logs show only `RX_STAGE_USERCOPY` events. IPv6 extension headers lose the NETIF/USERCOPY paths, unknown ABI families are mislabeled as IPv4, and the documented event schema omits `address_family`. Rebase onto current `main`, resolve the conflict, fix the topology and stage-specific coverage, then rerun the complete checks. The existing ABI and loader design can stay; no broader abstraction is needed.

### Other review dimensions

- Security: no command/SQL injection, secret, path-traversal, cryptography, or unsafe resource-lifecycle issue found in the changed production paths.
- Code/style: new files have the required Apache 2.0 headers; the focused PR diff passes `git diff --check`.
- Local checks: `bash -n integration/test_net_rx_latency_ipv6.sh` and `cc -fsyntax-only -Wall -Wextra integration/testdata/test_net_rx_latency_ipv6_user.c` pass.
- CI: build/lint and QEMU jobs passed at the immutable head, but must be rerun after rebasing because the branch is diverged and unmergeable.

---
*Co-authored by: gpt-5.6-sol*

hao022:
@123123213weqw 一个pr是一个大的功能提交。不需要和其他非相关的补丁一起提交。

PR #507, opened 2026-08-04
fix(procfs): validate ARP cache field layout

123123213weqw:
## What changed

- reject empty or truncated `/proc/net/stat/arp_cache` input
- validate that the number of counter values matches the header
- add regression coverage for missing, short, and oversized value rows

## Root cause

`NetArpCache` ignored failed `Scanner.Scan` calls and indexed the header slice without validating row widths. Truncated input could silently return empty metrics, while an oversized value row could panic.

## Validation

- `go test ./internal/procfs`
- `git diff --check`

## CI compatibility

- accept `clock_nanosleep` as the shallow blocked frame produced by Ubuntu 20.04's off-CPU stack symbolization
- wait for the first netdev sample before asserting metric include/exclude filters on slower VMs

### DIFF
--- a/integration/test_profiler_native_cpu_offcpu.sh
+++ b/integration/test_profiler_native_cpu_offcpu.sh
@@ -5,9 +5,9 @@
 
 # Verify the native off-CPU profiler attributes blocked time: a fixture that
 # loops in nanosleep must produce folded output containing the blocked category
-# root and the wait_loop user chain. The assertion anchors on wait_loop because
-# user-stack capture depth varies by kernel and glibc build: older combos can
-# drop the inner blocking_wait frame while still resolving the outer chain.
+# root and a blocking wait frame. User-stack capture depth and symbolization
+# vary by kernel and glibc build: Ubuntu 20.04 can stop at clock_nanosleep,
+# while newer combinations also resolve the fixture's wait_loop frame.
 
 set -euo pipefail
 
@@ -23,7 +23,7 @@ readonly FIXTURE_SRC="${ROOT_DIR}/integration/testdata/test_profiler_offcpu.user
 
 readonly PROFILER_DURATION=10
 readonly PROFILER_AGGR_INTERVAL=5
-readonly BLOCKED_PATTERN='off-CPU blocked;.*;wait_loop'
+readonly BLOCKED_PATTERN='off-CPU blocked;.*(wait_loop|clock_nanosleep)'
 
 WORK_DIR=$(mktemp -d "${HUATUO_BAMAI_TEST_TMPDIR}/profiler-offcpu.XXXXXX")
 TOOL_OUT="${WORK_DIR}/profiler.out"
--- a/internal/procfs/net_arpcache.go
+++ b/internal/procfs/net_arpcache.go
@@ -16,6 +16,7 @@ package procfs
 
 import (
 	"bufio"
+	"fmt"
 	"os"
 	"strconv"
 	"strings"
@@ -38,17 +39,33 @@ func NetArpCache() (*ArpCacheStats, error) {
 	defer file.Close()
 
 	scanner := bufio.NewScanner(file)
-	scanner.Scan()
+	if !scanner.Scan() {
+		if err := scanner.Err(); err != nil {
+			return nil, fmt.Errorf("scan arp cache header: %w", err)
+		}
+		return nil, fmt.Errorf("arp cache header is missing")
+	}
 
 	// First string is always a header for stats
-	var headers []string
-	headers = append(headers, strings.Fields(scanner.Text())...)
+	headers := strings.Fields(scanner.Text())
+	if len(headers) == 0 {
+		return nil, fmt.Errorf("arp cache header is empty")
+	}
 
 	// Fast path ...
 	cache := &ArpCacheStats{Stats: make(map[string]uint64)}
 
-	scanner.Scan()
-	for num, counter := range strings.Fields(scanner.Text()) {
+	if !scanner.Scan() {
+		if err := scanner.Err(); err != nil {
+			return nil, fmt.Errorf("scan arp cache values: %w", err)
+		}
+		return nil, fmt.Errorf("arp cache values are missing")
+	}
+	values := strings.Fields(scanner.Text())
+	if len(values) != len(headers) {
+		return nil, fmt.Errorf("arp cache field count mismatch: got %d values for %d headers", len(values), len(headers))
+	}
+	for num, counter := range values {
 		value, err := strconv.ParseUint(counter, 16, 64)
 		if err != nil {
 			return nil, err
--- a/internal/procfs/net_arpcache.go
+++ b/internal/procfs/net_arpcache.go
@@ -16,6 +16,7 @@ package procfs
 
 import (
 	"bufio"
+	"fmt"
 	"os"
 	"strconv"
 	"strings"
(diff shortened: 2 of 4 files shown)


huhong-web (commented):
## Code Review

Reviewed `e43d4cc4fbb6041c8bf14562eff4b66e7b05602e...d48482abc07282dbb22d8926d7e119d6e3e939ab`. I found 3 actionable issues: two are attached inline, and one commit-history issue is listed below.

### Architecture & Design

**Verdict: needs changes.**

The core boundary is otherwise sound: `internal/procfs.NetArpCache` owns parsing and validation, while `core/metrics.nodeArpCacheEntries` consumes only the parsed `entries` value. The change introduces no new interface, mutable global state, goroutine, or resource owner, and unit fixtures are the right layer for malformed text. The exact head also passed the four real-kernel Ubuntu QEMU jobs, which provides the necessary kernel-format integration evidence.

The PR is not a coherent parser-only slice, however. `integration/test_metrics_exclude_filter.sh` and `integration/test_profiler_native_cpu_offcpu.sh` are independent CI fixes; the latter has already been superseded by merged PR #513 on current `main`. The branch is 63 commits behind and GitHub reports it as `DIRTY`. The smallest safe design is to rebase, retain current `main`'s off-CPU CPU-filter test, and keep this PR to the parser plus its unit tests. If the netdev readiness change is still needed, move it to a dedicated CI PR with its own failure evidence.

### Commit compliance

[P2] All three commits have empty bodies and lack the required `Signed-off-by:` DCO trailer. In addition, two subjects use the `test:` type even though the repository's `AGENTS.md` at the merge base requires `feat`, `fix`, or `refactor` with a scope and an explanatory body. Reword and sign off the retained commit(s) during the rebase.

### Security

No new security issue was found. The changed parser reads a constant procfs path, uses bounded `bufio.Scanner` tokenization and numeric parsing, and introduces no SQL, shell execution, secret, authorization, cryptographic, or user-controlled path flow.

### Verification

- `go test -count=1 ./internal/procfs`: pass
- `go test -race -count=1 ./internal/procfs`: pass
- `go vet ./internal/procfs`: pass
- `bash -n`, `shfmt -d`, and `git diff --check`: pass
- Exact-head GitHub CI: Build/Lint/Vendor and Ubuntu 20.04/22.04/24.04/26.04 QEMU jobs all passed
- Local `golangci-lint 2.5.0` could not load this repository's pre-v2 config; the exact-head CI `make check` result is the authoritative configured-linter evidence

---
*Co-authored by: gpt-5.6-sol*

internal/procfs/net_arpcache.go
```diff
@@ -38,17 +39,33 @@ func NetArpCache() (*ArpCacheStats, error) {
 	defer file.Close()
 
 	scanner := bufio.NewScanner(file)
-	scanner.Scan()
+	if !scanner.Scan() {
+		if err := scanner.Err(); err != nil {
+			return nil, err
```
huhong-web: [P2] Preserve which scan phase failed in the returned error.

A malformed/custom procfs file with a header larger than `bufio.Scanner`'s token limit reaches this branch, and `arpCollector.Update` receives only `bufio.Scanner: token too long`; the error no longer identifies `arp_cache` or the header. The same issue exists in the values branch below. Wrap both errors with phase-specific context (for example, `fmt.Errorf("scan arp cache header: %w", err)` and the corresponding values message), and add oversized-header/value tests so scanner failures cannot regress into a misleading missing-row result.

integration/test_profiler_native_cpu_offcpu.sh
```diff
@@ -23,7 +23,7 @@ readonly FIXTURE_SRC="${ROOT_DIR}/integration/testdata/test_profiler_offcpu.user
 
 readonly PROFILER_DURATION=10
 readonly PROFILER_AGGR_INTERVAL=5
-readonly BLOCKED_PATTERN='off-CPU blocked;.*;wait_loop'
+readonly BLOCKED_PATTERN='off-CPU blocked;.*(wait_loop|clock_nanosleep)'
```
huhong-web: [P2] Drop this compatibility hunk while rebasing instead of carrying it into the ARP parser fix.

Removing commits `eb37224` and `d48482a` does not affect `NetArpCache` or its regression tests. More importantly, current `main` already merged #513, which replaced this old stack-shape assertion with a PID/category assertion while preserving selected/excluded CPU coverage. This branch is now `DIRTY`; keeping this version during conflict resolution would restore the old object/CLI contract and remove the newer CPU-filter coverage. Rebase and retain `main`'s test. If the netdev readiness change is still independently needed, submit it as a focused CI change with its own failing-run evidence.

huhong-web (approved):
## Code Review

Reviewed `a1e74a8ef58499c98368e61fa892f168e989bb87...fc37892edf503330b8c2ddf575dff06aeb95b1bd`. No actionable findings; the previous scanner-context, branch-scope, and commit-compliance findings are addressed in this head.

### Architecture & Design

**Verdict: reasonable.**

- **Boundary and dependency direction:** `internal/procfs.NetArpCache` remains the owner of the `/proc/net/stat/arp_cache` parsing contract (`internal/procfs/net_arpcache.go:34-76`). Its unchanged caller, `core/metrics.nodeArpCacheEntries`, consumes only the parsed `entries` counter and propagates parser failures (`core/metrics/net_arp.go:51-66`). No transport, storage, configuration, or public API boundary is crossed.
- **Design trade-off:** validating the two rows before indexing is the smallest viable fix. It prevents the former panic/false-success paths without adding an interface, factory, registry, mutable global, goroutine, context, or new resource owner. A separate parser abstraction would add indirection without a second consumer or implementation.
- **Compatibility and failure semantics:** valid kernel output is handled as before; empty, truncated, scanner-failed, or width-mismatched input now fails closed instead of returning incomplete metrics or indexing beyond `headers`. Linux's `neigh_stat_seq_show` emits one counter for every header field, so the width invariant matches the producer contract.
- **Verification architecture:** malformed text is fully observable at the unit-test layer. The exact head also passed the repository's full QEMU `make test` jobs on Ubuntu 20.04, 22.04, 24.04, and 26.04, covering the real procfs/kernel boundary. No additional integration test is required for this focused parser change.

### Code Quality & Security

- Both modified Go files retain the required 2026 Apache 2.0 header.
- The single commit is focused, uses the required `fix(procfs):` form, includes an explanatory body, and carries a `Signed-off-by:` trailer.
- The parser reads a constant procfs path, uses bounded `bufio.Scanner` input and hexadecimal numeric parsing, and introduces no SQL, command execution, secret, authorization, cryptographic, or user-controlled path flow.
- The new tests cover missing header/values, scanner token limits in both phases, and both sides of the field-count mismatch.

### Verification

- `go test -count=1 ./internal/procfs`: pass
- `go test -race -vet=off -count=1 ./internal/procfs`: pass
- `go vet ./internal/procfs`: pass
- `gofmt -d` and `git diff --check`: pass
- Exact-head GitHub CI: Build/Lint/Vendor, PR-name lint, and all four QEMU jobs passed
- Local `golangci-lint 2.5.0` cannot load the repository's pre-v2 configuration; the exact-head CI `make build, check and vendor` success is the configured-linter evidence

---
*Co-authored by: gpt-5.6-sol*

123123213weqw:
@huhong-web Thanks again for the approval. The branch is still `mergeable: clean` and all checks are green on `fc37892e`; both P2 findings from the review round are addressed at this head. Could this be merged?

PR #508, opened 2026-08-04
fix(metrics): reject malformed netstat row pairs

123123213weqw:
## What changed

- parse `/proc/net/netstat` with whitespace-aware fields
- reject missing value rows, malformed headers, mismatched protocol pairs, and field-count mismatches
- report scanner failures with header/value phase context
- isolate the parser and add focused regression tests, including oversized scanner tokens

## Root cause

The parser ignored the result of the second `Scanner.Scan`, sliced protocol text without validating it, and paired adjacent rows without checking their protocol labels. Truncated or malformed procfs input could therefore be accepted incorrectly or panic.

## Validation

- `go test core/metrics/net_netstat_parser.go core/metrics/net_netstat_test.go -count=1`
- `git diff upstream/main...HEAD --check`

### DIFF
--- a/core/metrics/net_netstat_parser.go
+++ b/core/metrics/net_netstat_parser.go
@@ -1,0 +1,65 @@
+// Copyright 2026 The HuaTuo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package collector
+
+import (
+	"bufio"
+	"fmt"
+	"os"
+	"strings"
+)
+
+func parseNetStat(fileName string) (map[string]map[string]string, error) {
+	file, err := os.Open(fileName)
+	if err != nil {
+		return nil, err
+	}
+	defer file.Close()
+
+	stats := map[string]map[string]string{}
+	scanner := bufio.NewScanner(file)
+	for scanner.Scan() {
+		nameParts := strings.Fields(scanner.Text())
+		if len(nameParts) == 0 || !strings.HasSuffix(nameParts[0], ":") {
+			return nil, fmt.Errorf("invalid netstat header in %s: %q", fileName, scanner.Text())
+		}
+
+		if !scanner.Scan() {
+			if err := scanner.Err(); err != nil {
+				return nil, err
+			}
+			return nil, fmt.Errorf("missing values for %s in %s", nameParts[0], fileName)
+		}
+		valueParts := strings.Fields(scanner.Text())
+		if len(valueParts) == 0 || valueParts[0] != nameParts[0] {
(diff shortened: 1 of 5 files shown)


core/metrics/net_netstat_parser.go
```diff
@@ -0,0 +1,68 @@
...
+		stats[protocol] = map[string]string{}
+		for i := 1; i < len(nameParts); i++ {
+			stats[protocol][nameParts[i]] = valueParts[i]
+		}
+	}
+
+	if err := scanner.Err(); err != nil {
+		return nil, fmt.Errorf("scan netstat header in %s: %w", fileName, err)
+	}
+	return stats, nil
```
huhong-web, line 67: [P2] Reject an empty procfs snapshot before returning success. If `/proc/<pid>/net/netstat` or `/proc/<pid>/net/snmp` is empty (truncated before the first header), this loop never runs and the function returns an empty map with a nil error. `buildNetAndSnmpStat` then merges or publishes the remaining data as a successful scrape, silently removing the `Tcp`/`TcpExt` series without recording a collector failure. Require at least one supported header/value pair before this return (for example, return a descriptive error when `len(stats) == 0`) and add an empty-input regression case that fails before the fix and passes afterward.

huhong-web (commented):
## Code Review

Found 1 actionable issue; see the inline comment.

### Architecture & Design

**Verdict: needs changes.** The affected path is `netstatCollector.Update` → `buildNetAndSnmpStat` → `parseNetStat`, with output flowing through the existing matcher into metric emission. Keeping the parser as a package-local helper is the smallest viable design: it adds no public API, interface, registry, dependency, global state, goroutine, or lifecycle obligation. The extraction also leaves configuration and metric naming contracts unchanged.

The remaining risk is the parser-to-collector error contract: an empty snapshot is still reported as a successful parse, allowing incomplete telemetry to be published without an error. The smallest correction is local validation in `parseNetStat`; a new abstraction is not needed. Unit coverage is the correct verification layer for this text-parsing contract. The real procfs wiring is unchanged, and the repository's VM suites cover that boundary.

### Other Review Dimensions

- Copyright: both new files use the required 2026 Apache 2.0 header; the modified file already includes 2026.
- Go style, naming, error handling, and tests: formatting is clean and the new row/scan failures are handled explicitly, apart from the inline empty-snapshot case.
- Concurrency and lifecycle: no shared state, goroutines, channels, locks, or resource-ownership changes; the opened file is owned and closed by the helper.
- Security: no SQL or command construction, secrets, cryptography, or user-controlled path flow was introduced. The only caller builds a procfs path from a numeric PID, and scanner token size remains bounded.
- Compatibility: `strings.Fields` correctly accepts repeated ASCII whitespace used by procfs while retaining the existing `Tcp`/`TcpExt` output contract.

### Validation

- `gofmt -l` on all changed Go files: pass
- `git diff --check origin/main...HEAD`: pass
- `go vet core/metrics/net_netstat_parser.go core/metrics/net_netstat_test.go`: pass
- `go test core/metrics/net_netstat_parser.go core/metrics/net_netstat_test.go -count=1`: pass
- GitHub CI: Build/Lint/Vendor and Ubuntu 20.04, 22.04, 24.04, and 26.04 VM suites all pass
- Full local package test could not run without generated `internal/bpf/abi`; the repository CI generated artifacts and exercised the full suite successfully. The installed local `golangci-lint` v2.5.0 cannot consume this repository's v1 configuration, so the green repository lint job is the authoritative lint result.

---
*Co-authored by: gpt-5.6-sol*

PR #509, opened 2026-08-04
fix(kmsgutil): propagate nonblocking setup failures

123123213weqw:
## What changed

- replace raw `fcntl` calls with an error-returning nonblocking helper
- propagate failures when `/dev/kmsg` cannot be switched to nonblocking mode
- add regression coverage for an invalid file descriptor

## Root cause

Both failing `fcntl` branches returned the stale `err` variable from the preceding successful write. That value was normally nil, so `GetSysrqMsg` could report success after failing to configure `/dev/kmsg`, then potentially block while reading it.

## Validation

- `go test ./internal/utils/kmsgutil` on `WZU_4080`
- `git diff --check`

The package is Linux-specific and does not compile on the local macOS host because `syscall.Sysinfo` is unavailable there.

## CI compatibility

- accept `clock_nanosleep` as the shallow blocked frame produced by Ubuntu 20.04's off-CPU stack symbolization
- wait for the first netdev sample before asserting metric include/exclude filters on slower VMs

### DIFF
--- a/integration/test_metrics_exclude_filter.sh
+++ b/integration/test_metrics_exclude_filter.sh
@@ -23,7 +23,18 @@ source "${ROOT_DIR}/integration/config.sh"
 
 integration_huatuo_bamai_start write_exclude_filter_config
 
-huatuo_bamai_await_metrics
+# The metrics endpoint can become ready before the netdev collector publishes
+# its first sample on slower VMs. Wait for both fixture interfaces so the
+# filter assertions do not race collector initialization.
+collect_exclude_filter_metrics() {
+	huatuo_bamai_collect_metrics || return 1
+	grep -qE 'huatuo_bamai_netdev_.*device="eth0"' "${HUATUO_BAMAI_TEST_TMPDIR}/metrics.txt" \
+		&& grep -qE 'huatuo_bamai_netdev_.*device="eth1"' "${HUATUO_BAMAI_TEST_TMPDIR}/metrics.txt"
+}
+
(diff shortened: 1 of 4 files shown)


internal/utils/kmsgutil/kmsgutil.go
```diff
@@ -92,6 +86,13 @@ func GetSysrqMsg(command string) (string, error) {
 	return formatKmsgs(buffer.String()), nil
 }
 
+func setNonblocking(fd int) error {
```
hao022: 不需要这么薄的封装

integration/test_metrics_exclude_filter.sh
```diff
@@ -23,7 +23,18 @@ source "${ROOT_DIR}/integration/config.sh"
...
 
-huatuo_bamai_await_metrics
+# The metrics endpoint can become ready before the netdev collector publishes
+# its first sample on slower VMs. Wait for both fixture interfaces so the
+# filter assertions do not race collector initialization.
+collect_exclude_filter_metrics() {
+	huatuo_bamai_collect_metrics || return 1
+	grep -qE 'huatuo_bamai_netdev_.*device="eth0"' "${HUATUO_BAMAI_TEST_TMPDIR}/metrics.txt" \
+		&& grep -qE 'huatuo_bamai_netdev_.*device="eth1"' "${HUATUO_BAMAI_TEST_TMPDIR}/metrics.txt"
+}
```
hao022: 和kmsg没有关系，需要分开提交

PR #518, opened 2026-08-07
fix(ci): skip web sync for fork pull requests

lizhenhao123:
## Summary

- skip web synchronization for pull requests from forks
- avoid requiring repository secrets on contributor branches
- retain synchronization for trusted repository branches

## Validation

- [ ] GitHub Actions

### DIFF
--- a/.github/workflows/sync-web-on-change.yml
+++ b/.github/workflows/sync-web-on-change.yml
@@ -19,6 +19,9 @@ on:
 
 jobs:
   sync:
+    if: >-
+      github.event_name != 'pull_request' ||
+      github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     steps:
       - name: checkout huatuo


huhong-web (approved):
## Code Review

**Verdict: Approved — no actionable findings at `c467aefb995a7c8528185de3f87392dfb6878981`.**

The PR goal is coherent: prevent fork-originated pull requests from entering the secret-dependent website deployment while preserving `push`, `workflow_dispatch`, and same-repository pull-request behavior.

### Architecture & Design

**Verdict: Reasonable.** The new admission check in `.github/workflows/sync-web-on-change.yml:22-24` is owned by the `sync` job itself, before runner allocation and before the checkout → Hugo/Node build → GitHub Pages/SSH deployment chain at lines 27-93.

The event contract is complete for the workflow's current triggers:

- fork PR: `head.repo.full_name != github.repository` → the whole job is skipped;
- same-repository PR: the repository identities match → synchronization remains enabled;
- `push` and `workflow_dispatch`: the non-PR clause is true → existing behavior is preserved.

Compared with removing the `pull_request` trigger (which would also remove trusted PR behavior) or repeating conditions on every secret-consuming step (which could leave partial work), this job-level gate is the smallest design that owns the invariant in one place. It adds no abstraction, cross-layer dependency, mutable state, API/data-model change, or concurrency/resource-lifecycle risk.

The only maintenance caveat is that any future non-PR trigger added to this workflow is allowed by the first clause and should be reviewed against the deployment-secret boundary.

### Security & Code Quality

No security finding. A recent fork run used `Secret source: None` and failed at the private web-repository checkout with “Input required and not supplied: token”; the new condition rejects the same concrete `sisttri/huatuo -> ccfos/huatuo` path before any secret-bearing step. The workflow does not switch to `pull_request_target` or check out fork code.

No Go or storage code and no new files are changed, so Go style, copyright, sentinel-error/backend naming, concurrency, injection, filesystem, and cryptography checks are not applicable.

### Verification

- Review range: `0242aa935cbe200710a30489c08f868d6a438615...c467aefb995a7c8528185de3f87392dfb6878981` (merge base equals the base SHA).
- GitHub: all 6 checks on the head SHA pass.
- `git diff --check`: pass.
- `actionlint v1.7.7`: the changed expression parses and type-checks. Full-file lint separately reports the pre-existing, unchanged `peaceiris/actions-gh-pages@v3` runtime warning at line 63; that is outside this PR's causal scope.
- Runtime acceptance after merge: a fork PR touching `docs/**` or `data/img/**` should show `sync` as skipped/success, while a main-branch push touching those paths should still deploy.

---
*Co-authored by: gpt-5.6-sol*

PR #519, opened 2026-08-07
feat(storage): add Pyroscope HTTP backend

lizhenhao123:
## Summary

- add a write-only Pyroscope HTTP backend
- validate endpoints and authentication before sending requests
- cover labels, payloads, bounded responses, and HTTP failures

## Validation

- [x] `go test ./internal/storage/pyroscope`
- [ ] GitHub Actions

### DIFF
--- a/.github/workflows/sync-web-on-change.yml
+++ b/.github/workflows/sync-web-on-change.yml
@@ -19,6 +19,9 @@ on:
 
 jobs:
   sync:
+    if: >-
+      github.event_name != 'pull_request' ||
+      github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     steps:
       - name: checkout huatuo
--- a/internal/storage/driver/types.go
+++ b/internal/storage/driver/types.go
@@ -57,6 +57,13 @@ type Config struct {
 	ESUsername  string
 	ESPassword  string
 	ESIndex     string
+
+	PyroscopeAddress        string
+	PyroscopeAppNamePrefix  string
+	PyroscopeUsername       string
+	PyroscopePassword       string
+	PyroscopeBearerToken    string
+	PyroscopeTimeoutSeconds int
 }
 
 // Op is a storage query operator.
--- a/internal/storage/pyroscope/pyroscope.go
+++ b/internal/storage/pyroscope/pyroscope.go
@@ -1,0 +1,405 @@
+// Copyright 2026 The HuaTuo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// Package pyroscope implements a write-only pprof storage backend.
+package pyroscope
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"io"
+	"net/http"
+	"net/url"
+	"strconv"
+	"strings"
+	"time"
+	"unicode"
+
+	"huatuo-bamai/internal/storage/driver"
+)
+
+const (
+	defaultAppNamePrefix    = "huatuo"
+	defaultTimeoutSeconds   = 5
+	maxErrorResponseBytes   = 4 * 1024
+	pprofFormat             = "pprof"
+	protobufContentType     = "application/octet-stream"
+	authorizationHeaderName = "Authorization"
+)
+
+// Config contains Pyroscope backend settings.
+type Config struct {
+	Address        string
+	AppNamePrefix  string
+	Username       string
+	Password       string
+	BearerToken    string
+	TimeoutSeconds int
+}
+
+// Storage pushes protobuf-encoded pprof profiles to Pyroscope's HTTP ingest API.
+type Storage struct {
+	ingestURL     *url.URL
+	appNamePrefix string
+	username      string
+	password      string
+	bearerToken   string
+	httpClient    *http.Client
+}
+
+var _ driver.Backend = (*Storage)(nil)
+
+func init() {
+	driver.RegisterBackend("pyroscope", func(cfg *driver.Config) (driver.Backend, error) {
+		return NewBackend(&Config{
+			Address:        cfg.PyroscopeAddress,
+			AppNamePrefix:  cfg.PyroscopeAppNamePrefix,
+			Username:       cfg.PyroscopeUsername,
+			Password:       cfg.PyroscopePassword,
+			BearerToken:    cfg.PyroscopeBearerToken,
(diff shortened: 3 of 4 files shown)


internal/storage/driver/types.go
```diff
@@ -57,6 +57,13 @@ type Config struct {
 	ESUsername  string
 	ESPassword  string
 	ESIndex     string
+
+	PyroscopeAddress        string
+	PyroscopeAppNamePrefix  string
```
hao022, line 62: PyroscopeAppNamePrefix/PyroscopeTimeoutSeconds 内部提供一个默认值？

.github/workflows/sync-web-on-change.yml
```diff
@@ -19,6 +19,9 @@ on:
 
 jobs:
   sync:
+    if: >-
+      github.event_name != 'pull_request' ||
```
hao022, line 23: 当前补丁 不应该包含其他补丁的代码。

internal/storage/driver/types.go
```diff
@@ -57,6 +57,13 @@ type Config struct {
...
+	PyroscopeUsername       string
```
hao022, line 63: 需要提供一个针对该存储类型的，集成测试，并在 docs/best-practice/storage_zh.md 更新使用说明（可参考其他存储）。

internal/storage/pyroscope/pyroscope.go
```diff
@@ -0,0 +1,405 @@
...
+	appNamePrefix string
+	username      string
+	password      string
+	bearerToken   string
+	httpClient    *http.Client
+}
+
+var _ driver.Backend = (*Storage)(nil)
+
+func init() {
```
huhong-web, line 65: [P1] This registration never runs on the product storage path. Built-ins become reachable through side-effect imports in `internal/storage/backends.go`, and that file still imports only Elasticsearch, localfile, and SQLite; `go list -deps huatuo-bamai/internal/storage` likewise contains no Pyroscope package. Consequently `storage.NewFromConfig(... Driver: "pyroscope")` returns “backend driver not registered.” Add the built-in import and a test that constructs this backend through `storage.NewFromConfig`, not by calling `NewBackend` directly.

internal/storage/pyroscope/pyroscope.go
```diff
@@ -0,0 +1,405 @@
...
+	query.Set("from", strconv.FormatInt(start.Unix(), 10))
+	query.Set("until", strconv.FormatInt(end.Unix(), 10))
+	query.Set("format", pprofFormat)
+	endpoint.RawQuery = query.Encode()
+
+	req, err := http.NewRequestWithContext(
+		driver.WithContext(ctx),
+		http.MethodPost,
+		endpoint.String(),
+		bytes.NewReader(rec.Data),
```
huhong-web, line 229: [P1] No production mapper supplies the pprof contract declared here. `storage.Store.record` obtains `rec.Data` from `Mapper.Encode`; the only profile mapper, `ProfileDocumentStoreMapper`, inherits `DocumentStoreMapper.Encode`, which JSON-marshals the entire `tracing.Document`. Its `Fields` also omits `profile_start_time`, `profile_end_time`, and `profile_type`. If this backend is wired into the existing profile writer, it therefore posts JSON while declaring `format=pprof`, and falls back to the upload/tracer time plus one second. Pyroscope requires raw pprof bytes for this format (see the [server API](https://grafana.com/docs/pyroscope/latest/reference-server-api/#the-pprof-format)). Add a dedicated mapper that serializes `TracerData.Flamedata.Profile` as protobuf and derives the real type/time fields, then exercise `Store.Save` with a real profile instead of a hand-built `driver.Record`.

internal/storage/pyroscope/pyroscope.go
```diff
@@ -0,0 +1,405 @@
...
+func parseAddress(address string) (*url.URL, error) {
+	rawAddress := strings.TrimSpace(address)
+	if rawAddress == "" {
+		return nil, fmt.Errorf("pyroscope: address is empty")
+	}
+	endpoint, err := url.Parse(rawAddress)
+	if err != nil {
+		return nil, fmt.Errorf("pyroscope: parse address: %w", err)
+	}
+	if endpoint.Scheme != "http" && endpoint.Scheme != "https" {
```
huhong-web, line 130: [P1] This permits Basic or Bearer credentials to be sent over plaintext HTTP. `validateAuthentication` accepts either credential form and `Save` then places it in the `Authorization` header, so any observer between HuaTuo and a non-local `http://` endpoint can recover both the credential and profile payload. Reject authenticated HTTP endpoints (or restrict an explicitly documented insecure mode to loopback) and add a constructor test proving HTTP+auth is rejected while HTTPS+auth remains valid.

huhong-web (changes requested):
## Code Review

Reviewed commit `a9864cb31c0341e88338ea0f8f52c7159a7b506f` against merge base `0242aa935cbe200710a30489c08f868d6a438615`. The current `main` tip is `47dabdbf46e1276c06792965c728d8f0094b2f1a` (65 commits ahead), so this PR also needs a rebase and fresh CI evidence.

I found 3 new blocking issues; each is attached to the relevant changed line. The earlier unresolved threads about the unrelated workflow commit, defaults, documentation, and a real-backend integration test also remain applicable.

## Architecture & Design

**Verdict: needs changes.**

The intended boundary is reasonable: a write-only Pyroscope implementation can participate in the existing profile-store fan-out. The current end-to-end contract is incomplete, however:

`ProfilerEvent -> tracing.SaveProfile -> documentWriter -> storage.Store.record -> Mapper.Encode/Fields -> pyroscope.Storage.Save -> /ingest`

- `internal/storage/backends.go` does not import the new package, so the registry path cannot construct it.
- The production `ProfileDocumentStoreMapper` emits a JSON document and does not emit the time/profile fields that this backend assumes; `Save` declares those same bytes to be pprof protobuf.
- No application configuration or startup wiring selects the backend. The unit tests bypass all three boundaries by constructing `driver.Record` directly.

The smallest viable design is to keep the existing store lifecycle, add the built-in import and startup/config wiring, and use a Pyroscope-specific mapper that serializes the nested profile as actual pprof bytes and derives its labels/time range. A separate exporter abstraction would add a parallel lifecycle without solving the current requirement. Because the riskiest contract is the real HTTP/profile-format boundary, acceptance needs a component/integration test that saves a real HuaTuo profile through `storage.Store` into Pyroscope and queries it back (or otherwise proves successful ingestion and labels), plus simultaneous Elasticsearch/Pyroscope fan-out coverage.

Concurrency and resource ownership are otherwise straightforward: the backend has immutable request configuration, uses caller context plus a client timeout, does not spawn goroutines, bounds captured error bodies, and refuses redirects.

## Other checks

- License headers: pass for both new Go files.
- Go style/open-source compliance/error sentinels: no additional finding.
- Security: one blocking transport finding is inline; no SQL, command, path, crypto, or hard-coded-secret path was introduced.
- `git diff --check`: pass.
- `go test ./internal/storage/pyroscope`: pass.
- `go vet ./internal/storage/pyroscope`: pass.
- GitHub “Build, Lint and Vendor”: pass on this HEAD.
- Ubuntu 20.04 VM CI: failing in the existing metrics integration test because `mountpoint_perm_ro{...mountpoint="/boot"` is absent; this does not exercise Pyroscope, but the required check is not green.
- Full local `make golangci-lint`: not completed because the review workspace ran out of disk during generation; rely only on the scoped checks and recorded CI result above.

---
*Co-authored by: gpt-5.6-sol*

PR #520, opened 2026-08-07
feat(storage): wire Pyroscope profile storage

lizhenhao123:
## Summary

- map tracing profiles into Pyroscope pprof payloads
- expose Pyroscope settings through huatuo-bamai configuration
- register and initialize the profile storage backend
- document the new storage settings

## Validation

- [ ] GitHub Actions

### DIFF
--- a/.github/workflows/sync-web-on-change.yml
+++ b/.github/workflows/sync-web-on-change.yml
@@ -19,6 +19,9 @@ on:
 
 jobs:
   sync:
+    if: >-
+      github.event_name != 'pull_request' ||
+      github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     steps:
       - name: checkout huatuo
--- a/cmd/huatuo-bamai/storage.go
+++ b/cmd/huatuo-bamai/storage.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"huatuo-bamai/cmd/huatuo-bamai/config"
@@ -80,8 +81,28 @@ func initStorage(storageRegion string, cfg *config.BamaiConfig) error {
 		tracing.SetTaskStore([]*storage.Store[*tracing.Document]{esStore}, tracing.DocumentOptions{Region: storageRegion})
 	}
 
+	profileStores, err := newProfileStores(context.Background(), cfg)
+	if err != nil {
+		return err
+	}
+	if len(profileStores) > 0 {
+		tracing.SetProfileStore(
+			profileStores,
+			tracing.DocumentOptions{Region: storageRegion},
+		)
+	}
+
+	return nil
+}
+
+func newProfileStores(
+	ctx context.Context,
+	cfg *config.BamaiConfig,
+) ([]*storage.Store[*tracing.Document], error) {
+	profileStores := make([]*storage.Store[*tracing.Document], 0, 2)
+
 	if cfg.Storage.Elasticsearch.Enabled() {
-		profileStore, err := storage.NewFromConfig[*tracing.Document](context.Background(), &driver.Config{
+		profileStore, err := storage.NewFromConfig[*tracing.Document](ctx, &driver.Config{
 			Driver:      "elasticsearch",
 			ESAddresses: strutil.SplitCommaList(cfg.Storage.Elasticsearch.Address),
 			ESUsername:  cfg.Storage.Elasticsearch.Username,
@@ -89,13 +110,43 @@ func initStorage(storageRegion string, cfg *config.BamaiConfig) error {
 			ESIndex:     cfg.Storage.Elasticsearch.Index,
 		}, profiler.MetadataCollection, tracing.ProfileDocumentStoreMapper{})
 		if err != nil {
-			return fmt.Errorf("new profiling document store (elasticsearch): %w", err)
+			return nil, fmt.Errorf("new profiling document store (elasticsearch): %w", err)
 		}
-		tracing.SetProfileStore(
-			[]*storage.Store[*tracing.Document]{profileStore},
-			tracing.DocumentOptions{Region: storageRegion},
-		)
+		profileStores = append(profileStores, profileStore)
 	}
 
-	return nil
+	if cfg.Storage.Pyroscope.Address != "" {
+		profileStore, err := storage.NewFromConfig[*tracing.Document](ctx, &driver.Config{
+			Driver:                  "pyroscope",
+			PyroscopeAddress:        cfg.Storage.Pyroscope.Address,
+			PyroscopeAppNamePrefix:  cfg.Storage.Pyroscope.AppNamePrefix,
+			PyroscopeUsername:       cfg.Storage.Pyroscope.Username,
+			PyroscopePassword:       cfg.Storage.Pyroscope.Password,
+			PyroscopeBearerToken:    cfg.Storage.Pyroscope.BearerToken,
+			PyroscopeTimeoutSeconds: cfg.Storage.Pyroscope.TimeoutSeconds,
+		}, profiler.MetadataCollection, tracing.PprofDocumentStoreMapper{})
+		if err != nil {
+			cleanupErr := closeProfileStores(ctx, profileStores)
+			return nil, errors.Join(
+				fmt.Errorf("new profiling document store (pyroscope): %w", err),
+				cleanupErr,
+			)
+		}
+		profileStores = append(profileStores, profileStore)
+	}
+
+	return profileStores, nil
+}
+
+func closeProfileStores(
+	ctx context.Context,
+	stores []*storage.Store[*tracing.Document],
+) error {
+	var errs []error
+	for _, store := range stores {
+		if err := store.Close(ctx); err != nil {
+			errs = append(errs, fmt.Errorf("close profiling document store %s: %w", store.Name, err))
+		}
+	}
+	return errors.Join(errs...)
 }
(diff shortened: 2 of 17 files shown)


.github/workflows/sync-web-on-change.yml
```diff
@@ -19,6 +19,9 @@ on:
 
 jobs:
   sync:
+    if: >-
+      github.event_name != 'pull_request' ||
```
hao022, line 23: 不需要该代码

cmd/huatuo-bamai/storage.go
```diff
@@ -80,22 +81,72 @@ func initStorage(storageRegion string, cfg *config.BamaiConfig) error {
...
+		tracing.SetProfileStore(
+			profileStores,
+			tracing.DocumentOptions{Region: storageRegion},
+		)
+	}
+
+	return nil
+}
+
+func newProfileStores(
```
hao022, line 98: 不需要改函数 newProfileStores，直接在原函数增加存储初始化就行。

core/autotracing/profiler_ingest.go
```diff
@@ -17,18 +17,19 @@ package autotracing
...
 )
 
 // ProfilerEvent is the payload sent by the profiler subprocess over toolstream.
 type ProfilerEvent struct {
-	TracerID      string `json:"tracer_id,omitempty"`
-	ContainerID   string `json:"container_id,omitempty"`
-	TracerName    string `json:"tracer_name,omitempty"`
-	TracerRunType string `json:"tracer_type,omitempty"`
-	TracerTime    string `json:"tracer_time"`
-	TracerData    any    `json:"tracer_data,omitempty"`
```
hao022, line 31: 为什么会将any换成profctx.TracerData？

pkg/tracing/document_pprof.go
```diff
@@ -0,0 +1,106 @@
...
+
+	start := time.Unix(0, profileData.Profile.TimeNanos).UTC()
+	if profileData.Profile.TimeNanos == 0 {
+		start = tracingDocumentTimeValue(document.TracerTime, document.UploadedTime)
+	}
+	fields["profile_start_time"] = start
+
+	end := start.Add(time.Duration(profileData.Profile.DurationNanos))
+	if !end.After(start) {
+		end = start.Add(time.Second)
```
huhong-web, line 66: [P1] Preserve the real aggregation window instead of fabricating one second. `profiler.ParseTree` never sets `PprofMetadata.Duration`, so every generated protobuf has `DurationNanos == 0`; the pipeline batches for `pctx.AggrInterval` (10 seconds by default), and native `buildPprofData` stamps the snapshot with `time.Now()`. This fallback therefore publishes `[snapshot, snapshot+1s]` while the samples cover a different, configurable window, and the raw profile still advertises zero duration. Pyroscope only repairs a missing/1970 start time; it does not synthesize duration ([receiver code](https://github.com/grafana/pyroscope/blob/ae04b4a42ebf2655aa173226942de04501a5ee11/pkg/og/convert/pprof/profile.go#L105-L112)). Carry the actual window start and elapsed duration into `PprofMetadata`, derive `from/until` from the same values, and test a non-default aggregation interval end to end.

pkg/tracing/document_pprof.go
```diff
@@ -0,0 +1,106 @@
...
+	ProfileDocumentStoreMapper
+}
+
+// Encode serializes the embedded pprof profile as protobuf.
+func (PprofDocumentStoreMapper) Encode(document *Document) ([]byte, error) {
+	profileData, err := profileDataFromDocument(document)
+	if err != nil {
+		return nil, err
+	}
+	return profileData.Profile.MarshalVT()
```
huhong-web, line 37: [P1] Preserve the receiver's profile-type identity. Raw `/ingest` derives the reserved series name from the pprof sample type ([Pyroscope `metricName`](https://github.com/grafana/pyroscope/blob/ae04b4a42ebf2655aa173226942de04501a5ee11/pkg/og/convert/pprof/profile.go#L172-L206)): Huatuo's `offcpu` matches “cpu” and becomes `process_cpu`, while `lock` remains `lock`. Those do not match `ProfileTypeOffCpuSample` (`process_offcpu:...`) or `ProfileTypeLockTimeSample` (`process_lock:...`); adding an ordinary `profile_type` tag does not set Pyroscope's reserved `__name__`. Define the intended mapping explicitly—using a receiver API that lets us set the series labels if necessary—and query back CPU, off-CPU, memory, and lock profiles to assert the complete name/sample/unit/period contract.

internal/storage/pyroscope/pyroscope.go
```diff
@@ -0,0 +1,405 @@
...
+
+func sanitizeLabelValue(value string) string {
+	return sanitizeProfilePart(value, false)
+}
+
+func sanitizeProfilePart(value string, allowDot bool) string {
+	var result strings.Builder
+	result.Grow(len(value))
+	for _, current := range strings.TrimSpace(value) {
+		if unicode.IsLetter(current) || unicode.IsDigit(current) ||
```
huhong-web, line 396: [P1] Implement the actual `/ingest` label-set grammar instead of applying one lossy sanitizer to two different syntaxes. Pyroscope service names allow ASCII `[A-Za-z0-9_.-/]` only ([validator](https://github.com/grafana/pyroscope/blob/ae04b4a42ebf2655aa173226942de04501a5ee11/api/model/labelset/validate.go#L85-L86)), but `unicode.IsLetter/IsDigit` preserves non-ASCII names, which makes every upload return 400. Conversely, tag values are opaque until `,` or `}` ([parser](https://github.com/grafana/pyroscope/blob/ae04b4a42ebf2655aa173226942de04501a5ee11/api/model/labelset/labelset.go#L132-L145)), yet `allowDot == false` rewrites the canonical colon-delimited `profile_type` and dotted hostnames to underscores. Validate service names with the receiver's exact ASCII grammar (and fail configuration early), and encode/validate tag values separately without silently changing legal identity fields.

internal/storage/pyroscope/pyroscope_test.go
```diff
@@ -0,0 +1,282 @@
...
+
+	"huatuo-bamai/internal/storage/driver"
+)
+
+func TestStorageSaveIngestsPprof(t *testing.T) {
+	profile := []byte{0x0a, 0x01, 0x00}
+	start := time.Unix(1700000000, 0).UTC()
+	end := start.Add(10 * time.Second)
+
+	httpClient := &http.Client{Transport: roundTripFunc(func(r *http.Request) (*http.Response, error) {
```
huhong-web, line 39: [P2] Add a real Pyroscope contract test. This custom `RoundTripper` only checks the request that our own code constructs; it never runs Pyroscope's label parser, pprof conversion, series-name derivation, or query path, so the time-range, profile-type, and label defects above all pass this suite. Repository guidance requires integration coverage for a new external dependency. Keep these fast transport/auth tests, and add a pinned Pyroscope round trip that writes and queries valid CPU, off-CPU, memory, and lock profiles, including a non-default aggregation interval and exact canonical labels.

huhong-web (commented):
## Code Review

Reviewed the immutable range `0242aa935cbe200710a30489c08f868d6a438615...82d22b9b764f41410165f4fadb051565a69ece7a`. I found four new merge-blocking issues and placed each on the affected line. I also checked the three existing unresolved threads and did not duplicate them.

### Architecture & Design

**Verdict: needs changes.**

The end-to-end path is `Storage.Pyroscope` configuration → `initStorage/newProfileStores` → the storage-driver registry → `PprofDocumentStoreMapper` → `pyroscope.Storage.Save` → Pyroscope's legacy `/ingest` endpoint. Reusing the existing storage lifecycle and giving Pyroscope a format-specific mapper are appropriate boundaries; Elasticsearch remains compatible, and the helper owns partial-construction rollback. The typed profiler event is also needed to reconstruct pprof after the JSON/toolstream boundary, although a neutral wire DTO would keep `core/autotracing` from depending on the broader runtime-context package.

The cross-boundary contract is incomplete: the adapter does not preserve the aggregation time range, does not preserve Huatuo's off-CPU/lock profile identity, and implements a label grammar that mutates legal values while accepting service names the receiver rejects. These are data-contract failures rather than style concerns. Fix the source metadata and receiver mapping first, then verify the complete path against a pinned real Pyroscope instance.

No new goroutine or resource-owner ambiguity was introduced. Fan-out remains synchronous and is bounded by the configured HTTP timeout; initialization rollback and shutdown ownership are explicit. Empty Pyroscope configuration preserves the prior Elasticsearch-only behavior, and no stored-document migration is required.

### Other review dimensions

- **Security:** URL/userinfo/query validation, mutually exclusive auth, redirect refusal, request timeout, bounded diagnostic capture, and secret placeholders are sound. I found no SQL/command injection, path traversal, hard-coded credential, or cryptography issue. Deployments using credentials still need HTTPS as documented.
- **Go and repository conventions:** new Go files have the Apache 2.0 header; package/module naming and direct `driver.ErrUnsupportedOp` use align with the repository; exact-head lint passed.
- **Scope:** `.github/workflows/sync-web-on-change.yml` is unrelated to profile storage and already has an unresolved review comment requesting removal.
- **Verification:** exact-head CI passed build, lint, vendor, and the Ubuntu 22.04/24.04/26.04 VM jobs. Ubuntu 20.04 failed in the pre-existing metrics include-filter scenario, not the changed profile path. The new external-backend path has no real Pyroscope round trip; the inline test finding describes the required contract/integration coverage.

**Result: blocked** (4 new findings; three are profile-data correctness defects).

---
*Co-authored by: gpt-5.6-sol*

PR #532, opened 2026-08-08
fix(config): apply runtime updates atomically

### LINKED ISSUE #531: [Bug]: Config endpoint applies partial updates after rejected requests
### Problem

`PUT /config` applies each entry directly to the global configuration before
the complete request has been checked.

If a request contains one valid field followed by an invalid field, the endpoint
returns `400`, but fields applied before the error remain active in memory.
Likewise, a configuration file sync failure returns `500` after the live
configuration has already changed.

Because JSON object iteration order is unspecified, the resulting live
configuration depends on map iteration order. The response says the update
failed even though part of it may have taken effect.

### Expected behavior

A configuration update should be transactional:

1. Apply every requested field to a staged copy.
2. Validate the complete staged configuration.
3. Persist it.
4. Publish it to runtime modules only after every step succeeds.

Rejected requests and persistence failures must leave the active configuration
unchanged.

### Impact

Operators can unknowingly run with a partially updated configuration that does
not match either the request result or the persisted configuration.

yyqdbngt:
Fixes #531

## Summary

- stage every requested configuration field on a deep copy
- validate the complete staged configuration before persistence
- publish the new configuration only after persistence succeeds
- serialize load, update, and sync operations
- preserve the active configuration after validation or persistence failures

## Root cause

The config handler called `config.Set` once per JSON object entry. Each call
mutated the shared configuration immediately, while validation and persistence
happened only after the loop. A later invalid entry or sync failure therefore
left earlier changes active even though the endpoint returned an error.

## Verification

Executed on `WZU_Server` after `make gen-build`:

```text
go test ./cmd/huatuo-bamai/config ./cmd/huatuo-bamai/handlers
ok  huatuo-bamai/cmd/huatuo-bamai/config
ok  huatuo-bamai/cmd/huatuo-bamai/handlers

make check
0 lint issues
```


hao022:
@yyqdbngt 这块代码更新比较大，可以根据最新代码，检查是否还有并发问题。

PR #539, opened 2026-08-08
fix(metrics): guard metax_gpu ListDieClocks empty slice panic (Fixes #538)

### LINKED ISSUE #538: metax_gpu: index out of range panic when ListDieClocks returns empty slice
## Bug Description

In `core/metrics/metax_gpu.go`, `metaxCollectDieMetrics` calls `sml.ListDieClocks()` and accesses `values[0]` without checking the slice length:

```go
values, err := sml.ListDieClocks(ctx, gpuId, dieId, ipC)
if err != nil {
    // ... error handling
} else {
    metrics = append(
        metrics,
        metric.NewGaugeData("clock_mhz", float64(values[0]), "GPU clock.", ...),  // PANIC if len(values)==0
    )
}
```

`sml.ListDieClocks` calls `mxSmlGetDieClocks` which may succeed (no error) but set `size=0` when the GPU is in idle/reset state and has no clock values for the given IP. This returns an empty `[]uint32{}` with `nil` error, causing `values[0]` to **panic with index out of range**.

## Reproduction

1. Run huatuo-bamai on a host with MetaX GPU
2. Put the GPU in idle/reset state
3. The next metrics collection cycle will panic when accessing clock values for IPs that have no clock data

## Fix

Add a length check before accessing `values[0]`:

```diff
- } else {
+ } else if len(values) > 0 {
      metrics = append(
          metrics,
          metric.NewGaugeData("clock_mhz", float64(values[0]), ...),
      )
  }
```

wc4440222:
### Problem

When `mxSmlGetDieClocks` succeeds but returns `size=0` (GPU idle/reset state), `ListDieClocks` returns an empty `[]uint32{}` with `nil` error. The code accesses `values[0]` unconditionally, causing an **index out of range panic**.

### Fix

Add a length check before accessing `values[0]`:

```diff
- } else {
+ } else if len(values) > 0 {
```

When the slice is empty, the metric is simply skipped.

Fixes #538

### DIFF
--- a/core/metrics/metax_gpu.go
+++ b/core/metrics/metax_gpu.go
@@ -454,7 +454,7 @@ func metaxCollectDieMetrics(ctx context.Context, gpuId, dieId uint32, series gpu
 				return nil, fmt.Errorf("failed to %s: %w", operationListClocks, err)
 			}
 			log.Debugf("operation %s not supported on gpu %d die %d", operationListClocks, gpuId, dieId)
-		} else {
+		} else if len(values) > 0 {
 			metrics = append(
 				metrics,
 				metric.NewGaugeData("clock_mhz", float64(values[0]), "GPU clock.", map[string]string{


wc4440222:
@hao022 你好，这几个 PR 已 open 9 天，都是修复 panic/下溢问题，改动很小：

- #539 fix(metrics): guard metax_gpu ListDieClocks empty slice panic (Fixes #538)
- #541 fix(metrics): guard net_arp count-1 underflow (Fixes #540)
- #544 fix(metrics): check scanner.Scan return value in parseNetStat (Fixes #543)

方便的时候帮忙看一下？谢谢！

PR #541, opened 2026-08-08
fix(metrics): guard net_arp count-1 underflow producing negative metric (Fixes #540)

### LINKED ISSUE #540: net_arp: count-1 underflow produces negative metric when /proc/net/arp is empty
## Bug Description

In `core/metrics/net_arp.go`, `CountLines()` returns `int64`. The code computes `count-1` to subtract the header line from ARP entry count, but when `/proc/<pid>/net/arp` is **empty** (0 lines), `count` = 0, so `count-1` = -1. `float64(-1)` produces a **negative gauge value** for the "entries" metric.

This affects both:
- `nodeArpCacheEntries()` (line 63): host ARP entries
- `Update()` per-container loop (line 83): container ARP entries

## Impact

Negative ARP entry counts are meaningless and can trigger false-positive Prometheus alerts. This commonly occurs for containers with no network activity (empty `/proc/<pid>/net/arp`).

## Fix

Use `max(count-1, 0)` to clamp the value at zero:

```diff
- metric.NewGaugeData("entries", float64(count-1), "host init namespace", nil),
+ metric.NewGaugeData("entries", float64(max(count-1, 0)), "host init namespace", nil),
```

wc4440222:
### Problem

When `/proc/<pid>/net/arp` is empty (0 lines), `CountLines()` returns 0, and `count-1` evaluates to -1. `float64(-1)` produces a negative gauge value for the "entries" metric, which is meaningless and can trigger false-positive Prometheus alerts.

### Fix

Use `max(count-1, 0)` to clamp the value at zero for both host and per-container ARP entry counts.

Fixes #540

### DIFF
--- a/core/metrics/net_arp.go
+++ b/core/metrics/net_arp.go
@@ -60,7 +60,7 @@ func nodeArpCacheEntries() ([]*metric.Data, error) {
 	}
 
 	return []*metric.Data{
-		metric.NewGaugeData("entries", float64(count-1), "host init namespace", nil),
+		metric.NewGaugeData("entries", float64(max(count-1, 0)), "host init namespace", nil),
 		metric.NewGaugeData("total", float64(cache.Stats["entries"]), "all entries in arp_cache for containers and host netns", nil),
 	}, nil
 }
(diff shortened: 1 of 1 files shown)


wc4440222:
@hao022 你好，这个 PR 修复 `net_arp.go` 中 `count-1` 下溢导致负数指标的问题（Fixes #540）。已 open 9 天，方便的时候帮忙看一下？谢谢！

PR #385, opened 2026-07-18
feat(netdev): correlate hardware and kernel packet drops

btlqql:
## Summary

Resolves #326 by correlating dropwatch events with normalized NIC hardware counter deltas and producing an explainable hardware/driver/protocol-stack classification.

## Production problem

Dropwatch can identify where an skb is released inside the kernel, but it cannot see packets lost before they enter the kernel. Conversely, sysfs and `ethtool -S` expose cumulative NIC loss counters without the event, container, protocol, or stack context needed for root-cause analysis. Operators currently have to compare these sources manually and can easily assign the loss to the wrong layer.

## What changed

- Added a concurrency-safe, bounded correlation engine in `internal/dropcorrelation`:
  - normalizes generic and driver-specific counters for mlx5, i40e, ixgbe, bnxt and virtio;
  - computes monotonic deltas and excludes counter resets without uint64 underflow;
  - correlates on device, direction and measurement window;
  - classifies the lowest proven layer as `hardware`, `driver`, `protocol_stack`, or `unknown` with confidence and evidence;
  - bounds pending events, recent incidents and Prometheus reason cardinality during storms.
- Connected dropwatch ingestion to the shared correlator while preserving the original event storage path.
- Extended `netdev_hw` to read ethtool and sysfs RX/TX loss signals, persist `dropwatch_correlation` incidents, and export queue/reset/degraded/unmatched/classification metrics.
- Added configuration defaults and documented safe sizing/known correlation limits.
- Added a provisioned Grafana dashboard for layer rate, top reasons, hardware deltas and health signals.

## User and developer impact

Operators can distinguish pre-kernel hardware loss from driver and protocol-stack drops without manually aligning command output and event timestamps. Degraded reads, hot-reset intervals, unmatched hardware loss, queue pressure, and classification evidence are explicit rather than hidden.

The correlation policy is isolated from Linux and storage dependencies, allowing deterministic tests and future driver aliases without changing the event pipeline.

## Scope and size

- 2,361 added lines across implementation, tests, dashboard and documentation.
- 1,975 added Go lines; the requested implementation size is not met by generated/vendor content or documentation padding.

## Validation

- `go test ./internal/dropcorrelation ./pkg/types`
- `go vet ./internal/dropcorrelation`
- Linux/amd64 compile check: `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go test -c ./core/metrics`
- Grafana dashboard JSON parsed successfully with PowerShell `ConvertFrom-Json`.
[...]

PR #542, opened 2026-08-08
feat(network): trace TX path latency

btlqql:
## What

- Add `net_tx_latency` BPF tracing at `tcp_sendmsg`, `net_dev_queue`,
  `net_dev_start_xmit`, and `net_dev_xmit`.
- Report sendmsg-to-qdisc, qdisc-to-device, and device-to-NIC stages.
- Reuse the RX event ABI and the existing tracing storage pipeline.
- Expose event count, cumulative latency, and latest latency as Prometheus
  metrics with a `stage` label.
- Add configuration, bilingual documentation, Grafana views, unit coverage,
  and a root integration test using qdisc netem delay.

## Why

HUATUO currently observes RX path latency but has no equivalent visibility for
TCP transmission. This leaves qdisc and device-driver delays difficult to
distinguish.

## Impact

The tracer defaults to 50ms / 10ms / 1ms thresholds. It is IPv4/TCP-only,
matching the current RX event address layout. BPF output remains rate-limited
to 100 events per second.

## Design

`tcp_sendmsg` records short-lived socket origins. `net_dev_queue` correlates an
SKB with that origin and captures immutable packet metadata. Later tracepoints
read only bounded LRU state, avoiding dereferencing an SKB after the driver may
have consumed it. `NETDEV_TX_BUSY` preserves state for a retry.

## Checks

- `git diff --check`
- `goimports`, `gofumpt`, `shfmt`, and shell syntax validation
- Grafana JSON and repository TOML parse
- BPF/Go build, lint, vendor, and unit checks
- QEMU tests on Ubuntu 20.04, 22.04, 24.04, and 26.04

The web-sync job requires a maintainer-only token that is unavailable to fork
pull requests; it is unrelated to the code checks.

Refs #323


btlqql:
@huhong-web Rebased this PR onto the latest main and resolved the TX/RX latency filtering integration with the current dynamic configuration API. Fresh CI has been triggered. Could you please take a first review when convenient? (The historical sync failure was caused by the workflow token not being available to fork PRs.)

btlqql:
Follow-up: the fresh build exposed one rebase adaptation mistake (configSnapshot was missing in the TX tracer). Fixed in aa060244 and pushed; CI has restarted.

btlqql:
Follow-up from the completed VM matrix: all four jobs failed on the same config-clone test because the new TX exclusion slice was not deep-copied. Added the missing clone in df3e11e0; CI restarted.

btlqql:
### CI follow-ups after the rebase (all four VM jobs were red)

Rebasing this branch onto `852d724d` uncovered three defects that the stale
branch had never been tested against. All three are fixed in the two added
commits.

1. **It did not compile.** Both new Go files still imported `huatuo-bamai/...`
   (11 imports); the module is `github.com/ccfos/huatuo`. `tracing.WriteRequest`
   also no longer has a `TracerTime` field, so the save call now sets
   `ObservedTimestamp` like its `net_rx_latency` sibling.

2. **The sched_tick lifecycle tests.** `test_events_sched_tick.sh` and
   `test_events_sched_tick_irqoff.sh` assert
   `tracing_status_running\{[^}]*\} 1$`, so their config has to leave exactly one
   tracer running. `write_sched_tick_config_with_threshold` already disabled
   `net_rx_latency` for that reason but not the new `net_tx_latency`, so the
   gauge became 2 and both tests failed:

   ```
   [INTEGRATION TEST][FAIL] sched_tick lifecycle: expected present but not found:
     tracing_status_running\{[^}]*\} 1$
   huatuo_bamai_tracing_status_running{host="huatuo-dev",region="dev"} 2
   ..._hitcount{...,tracing="net_tx_latency"} 0
   ..._hitcount{...,tracing="sched_tick"} 0
   ```

   `net_tx_latency` is now disabled next to `net_rx_latency`, as the cpusys and
   iotracing configs already do.

3. **The new integration test could never start the daemon.**
   `write_net_tx_latency_config` was missing the `[HTTPServer.Auth]` section that
   every other writer in `integration/config.sh` emits:

   ```
   level="error" msg="app run: load config: validating HTTP server config:
     auth bearer token is required" file="main.go:71"
   ```

   `HTTPServerConfig.Validate` rejects an empty `Auth.BearerToken`; the sibling
   `write_net_rx_latency_config` sets
   `BearerToken = "integration-node-token"`. Added, verified by sourcing
   `integration/config.sh` under `bash` and printing the generated config.

The `sync` check that also shows red on this PR is the `sync web when content
changes` job, whose `checkout huatuo web` step ends with
`##[error]Input required and not supplied: token` — it needs a secret a fork PR
does not get, so it is not actionable from here.

btlqql:
Rebased onto `main` (`449cbb17`); this comment covers the real defects the rebase surfaced, plus the same changes on #385.

**PR #385 — root cause of the red CI fixed in `077233f4`**

The `Build, Lint and Vendor` failure was real, not infrastructure:

```
core/events/dropwatch.go:102:9: undefined: time
core/events/dropwatch.go:112:19: ev.PacketLen undefined (type *types.DropWatchTracing has no field or method PacketLen)
core/metrics/netdev_hw.go:315:4: unknown field TracerTime in struct literal of type tracing.WriteRequest
```

`main` renamed `DropWatchTracing.PacketLen` to `PacketLenBytes` and replaced `WriteRequest.TracerTime` with explicit envelope timestamps while that branch was open, so the merge no longer compiled and the VM jobs never reached the tracing tests. The rebase also had to reconcile `integration/config.sh`, the Grafana drop-correlation panels and the two instant-observability docs, which now keep `main`'s `memory_oom` naming alongside this branch's `net_tx_latency` addition.

Two behaviour notes on the adaptation:
- a drop is now timestamped with `ev.ObservedTimestamp` (the value the producer observed) instead of `time.Now()` at consumption, and the existing "observed timestamp is required" check runs before the drop is recorded;
- a finalised correlation writes `FinalizedAt` as its `ObservedTimestamp`, since `TracerTime` no longer exists.

**PR #542 — two more real defects fixed in `19a980af`**

- `core/events/net_tx_latency.go:140` passed `time.Now().UTC()` where `WriteRequest.ObservedTimestamp` is now a `timeutil.Timestamp`, so `huatuo-bamai` did not build after the merge; it uses `timeutil.Now()` like every other tracer.
- `TestMemoryOOMBlacklist` enumerates every registered tracer and failed once the new `net_tx_latency` tracer was added, so the new name is in that list.
- `integration/config.sh` conflicted with `main`'s `memory_oom` rename; the resolution keeps `main`'s names and adds `net_tx_latency`.

**Verification on the rebased trees** (Ubuntu 22.04, Go 1.24.13, CI toolchain parity):

| Check | #385 | #542 |
|---|---|---|
| `make build` | ok | ok |
| `make unit` | ok, 101 packages | ok, 100 packages |
| `golangci-lint run` | ok | ok |
| `gofumpt` / `goimports` | clean | clean |

The workflows on both heads now report `Build, Lint and Vendor` and `OS Distro QEMU Test` success; only the repository-wide `sync` job (issue #843) still fails.

PR #544, opened 2026-08-08
fix(metrics): check scanner.Scan return value in parseNetStat to prevent panic (Fixes #543)

### LINKED ISSUE #543: netstat: parseNetStat index out of range panic when /proc/net/netstat is truncated or empty
## Bug Description

In `core/metrics/net_netstat.go`, the `parseNetStat` function reads `/proc/<pid>/net/netstat` and `/proc/<pid>/net/snmp` in name-value pairs (header line + value line). The second `scanner.Scan()` call does not check the return value:

```go
for scanner.Scan() {
    nameParts := strings.Split(scanner.Text(), " ")

    scanner.Scan()                                          // line 140: return value NOT checked
    valueParts := strings.Split(scanner.Text(), " ")        // line 141: uses stale or empty text

    // remove trailing ":"
    protocol := nameParts[0][:len(nameParts[0])-1]          // line 144: PANIC on empty string
```

When the file is **truncated to an odd number of lines** (header without values) or **completely empty**:
1. The second `scanner.Scan()` returns `false`
2. `scanner.Text()` returns the **last successfully scanned text** (the name line) or **empty string** (if no scan ever succeeded)
3. `nameParts[0]` becomes `"Tcp:"` (duplicated) or `""` (empty file)
4. For empty string: `""[:0-1]` = `""[-1]` → **index out of range panic**
5. For duplicated text: `len(nameParts) != len(valueParts)` check catches this, but only after the `protocol` extraction already ran on wrong data

## Reproduction

1. Create a truncated file with only a header line:
```bash
echo "Tcp: RtoAlgorithm RtoMin RtoMax" > /tmp/truncated_netstat
```
2. Call `parseNetStat("/tmp/truncated_netstat")` — the second `scanner.Scan()` returns false, `scanner.Text()` returns the first line, and the function processes garbage data or panics on empty file.

## Fix

Check the return value of the second `scanner.Scan()`:

```diff
  for scanner.Scan() {
      nameParts := strings.Split(scanner.Text(), " ")
 
-     scanner.Scan()
+     if !scanner.Scan() {
+         break
+     }
      valueParts := strings.Split(scanner.Text(), " ")
```

When the second scan fails, break out of the loop and return the stats collected so far.

wc4440222:
### Problem

`parseNetStat` reads `/proc/<pid>/net/netstat` and `/proc/<pid>/net/snmp` in name-value pairs. The second `scanner.Scan()` call does not check the return value. When the file is truncated (odd number of lines) or empty, `scanner.Text()` returns stale or empty text, causing `nameParts[0][:len(nameParts[0])-1]` to panic with index out of range on an empty string.

### Fix

Check the return value of `scanner.Scan()` and break out of the loop when it returns `false`:

```diff
  for scanner.Scan() {
      nameParts := strings.Split(scanner.Text(), " ")

-     scanner.Scan()
+     if !scanner.Scan() {
+         break
+     }
      valueParts := strings.Split(scanner.Text(), " ")
```

When the second scan fails, the function returns whatever stats were collected so far instead of panicking.

Fixes #543

### DIFF
--- a/core/metrics/net_netstat.go
+++ b/core/metrics/net_netstat.go
@@ -138,7 +138,9 @@ func parseNetStat(fileName string) (map[string]map[string]string, error) {
 	for scanner.Scan() {
 		nameParts := strings.Split(scanner.Text(), " ")
 
-		scanner.Scan()
+		if !scanner.Scan() {
+			break
+		}
 		valueParts := strings.Split(scanner.Text(), " ")
 
 		// remove trailing ":"


wc4440222:
@hao022 你好，这个 PR 修复 `parseNetStat` 中未检查 `scanner.Scan` 返回值导致 panic 的问题（Fixes #543）。已 open 9 天，方便的时候帮忙看一下？谢谢！

PR #551, opened 2026-08-10
fix(metrics): guard parseTcpMemory against index out of range when tcp_mem has fewer than 3 values

### LINKED ISSUE #550: fix(metrics): parseTcpMemory panics when /proc/sys/net/ipv4/tcp_mem has fewer than 3 values
## Problem

`parseTcpMemory()` in `core/metrics/net_tcp_memory.go` accesses `values[2]` without checking the length of the `[]int` slice returned by `fs.SysctlInts("net.ipv4.tcp_mem")`.

```go
// core/metrics/net_tcp_memory.go, line 78-80
values, err := fs.SysctlInts("net.ipv4.tcp_mem")
if err != nil {
    return nil, err
}
// ...
memoryLimit: float64(values[2]), // tcpMemLimit  ← panics if len(values) < 3
```

The `newTcpMemory()` constructor only calls `procfs.RequireFile("sys/net/ipv4/tcp_mem")` which checks file **existence** via `os.Stat`, not content validity. When the file exists but is empty or has fewer than 3 fields (e.g., in a restricted container, chroot, or a corrupted procfs mount), `SysctlInts` returns a slice with length < 3, and `values[2]` triggers:

```
runtime error: index out of range [2] with length 0
```

## Reproduction

```bash
# Simulate an empty tcp_mem file
mkdir -p /tmp/fakeproc/sys/net/ipv4/
touch /tmp/fakeproc/sys/net/ipv4/tcp_mem
# Run with procfs root prefix set to /tmp/fakeproc
# → parseTcpMemory() panics with index out of range
```

## Fix

Add a length check before accessing `values[2]`:

```go
values, err := fs.SysctlInts("net.ipv4.tcp_mem")
if err != nil {
    return nil, err
}
if len(values) < 3 {
    return nil, fmt.Errorf("tcp_mem: expected at least 3 values, got %d", len(values))
}
```

This is distinct from PR #421 which fixed the `memoryLimit == 0` NaN panic (line 95-97) but did not guard the `values[2]` access.

wc4440222:
## Problem

`parseTcpMemory()` in `core/metrics/net_tcp_memory.go` accesses `values[2]` without checking the length of the `[]int` slice returned by `fs.SysctlInts("net.ipv4.tcp_mem")`.

```go
// core/metrics/net_tcp_memory.go, line 78-80
values, err := fs.SysctlInts("net.ipv4.tcp_mem")
if err != nil {
    return nil, err
}
// ...
memoryLimit: float64(values[2]), // tcpMemLimit  ← panics if len(values) < 3
```

The `newTcpMemory()` constructor only calls `procfs.RequireFile("sys/net/ipv4/tcp_mem")` which checks file **existence** via `os.Stat`, not content validity. When the file exists but is empty or has fewer than 3 fields (e.g., in a restricted container, chroot, or a corrupted procfs mount), `SysctlInts` returns a slice with length < 3, and `values[2]` triggers:

```
runtime error: index out of range [2] with length 0
```

This is distinct from PR #421 which fixed the `memoryLimit == 0` NaN panic (line 95-97) but did not guard the `values[2]` access.

## Fix

Add a length check before accessing `values[2]`:

```go
if len(values) < 3 {
    return nil, fmt.Errorf("tcp_mem: expected at least 3 values, got %d", len(values))
}
```

Fixes #550

### DIFF
--- a/core/metrics/net_tcp_memory.go
+++ b/core/metrics/net_tcp_memory.go
@@ -62,6 +62,9 @@ func parseTcpMemory() (*tcpMemoryStat, error) {
 	if err != nil {
 		return nil, err
 	}
+	if len(values) < 3 {
+		return nil, fmt.Errorf("tcp_mem: expected at least 3 values, got %d", len(values))
+	}
 
 	stat4, err := fs.NetSockstat()
 	if err != nil {
--- a/core/metrics/net_tcp_memory_test.go
+++ b/core/metrics/net_tcp_memory_test.go
@@ -1,0 +1,56 @@
+// Copyright 2025, 2026 The HuaTuo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package collector
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+
+	"huatuo-bamai/internal/procfs"
+)
+
+func TestParseTcpMemoryBoundary(t *testing.T) {
+	tempDir := t.TempDir()
+	sysctlDir := filepath.Join(tempDir, "proc", "sys", "net", "ipv4")
+	require.NoError(t, os.MkdirAll(sysctlDir, 0o755))
+
+	tcpMemPath := filepath.Join(sysctlDir, "tcp_mem")
+
+	procfs.RootPrefix(tempDir)
+	defer procfs.RootPrefix("/")
+
+	tests := []struct {
+		name    string
+		content string
+	}{
+		{"empty_file", ""},
+		{"one_value", "100"},
+		{"two_values", "100 200"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			err := os.WriteFile(tcpMemPath, []byte(tt.content), 0o600)
+			require.NoError(t, err)
+
+			_, err = parseTcpMemory()
+			require.Error(t, err)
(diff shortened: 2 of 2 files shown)


huhong-web (commented):
Code review found 1 actionable issue; see the inline comment. The bounds guard itself is correct. Validation: formatting and `git diff --check` passed; relevant PR checks were reviewed.

---
*Co-authored by: gpt-5.6-sol*

core/metrics/net_tcp_memory.go
```diff
@@ -62,6 +62,9 @@ func parseTcpMemory() (*tcpMemoryStat, error) {
 	if err != nil {
 		return nil, err
 	}
+	if len(values) < 3 {
```
huhong-web, line 65: [P2] Add a regression test for the truncated `tcp_mem` boundary

The guard is correct, but this PR adds no test for the exact panic path it fixes. Add table-driven cases for 0, 1, and 2 values (error/no panic) and at least 3 values (normal path), using the package's procfs fixture override. That will protect the `values[2]` boundary from being removed or moved before validation.

huhong-web (commented):
Re-reviewed immutable head `281f6830f41c` against merge base `cd33b7c7cab5`.

## Architecture & Design

**Verdict: reasonable, with an incomplete verification contract.** The length guard is correctly placed before `values[2]` in the metric parser and preserves existing ownership and error propagation. No new abstraction is needed, but the exact 0/1/2-value panic boundary has no regression test, so the riskiest behavior is not observable in CI.

Build/Lint is green; Ubuntu 20.04 fails in a broader metric job. No security issue was found.

---
*Co-authored by: gpt-5.6-sol*

huhong-web (commented):
Re-reviewed immutable head `281f6830f41c` against merge base `cd33b7c7cab5`.

## Architecture & Design

**Verdict: reasonable, with an incomplete verification contract.** The length guard is correctly placed before `values[2]` in the metric parser and preserves existing ownership and error propagation. No new abstraction is needed, but the exact 0/1/2-value panic boundary has no regression test, so the riskiest behavior is not observable in CI.

Build/Lint is green; Ubuntu 20.04 fails in a broader metric job. No security issue was found.

---
*Co-authored by: gpt-5.6-sol*

wc4440222:
@huhong-web Thanks for the review. Added regression test for 0/1/2 value boundary in `net_tcp_memory_test.go` using `procfs.RootPrefix` fixture override. Also added a 3+ values normal-path case that verifies the guard allows through to the next step.

PR #555, opened 2026-08-11
fix(storage): synchronize local writer cache

messere1:
## What

Protect every access to the local-file backend's `files` map with its mutex and remove the redundant secondary writer cache.

Add a synchronization regression test and concurrent-save coverage.

## Why

`writerByName` previously read `files` before acquiring the mutex while another goroutine could add a writer under the mutex. Concurrent saves for different tracer names could therefore race and trigger Go's concurrent map read/write failure.

## Tests

- `make check`
- `go test -race ./internal/storage/localfile -count=1 -timeout=90s`
- `make unit` (the changed package passes; the full run is blocked in this WSL environment by the existing cgroup v1 fixture expecting `/sys/fs/cgroup/cpu/.../cpuacct.stat`)

### DIFF
--- a/internal/storage/localfile/localfile.go
+++ b/internal/storage/localfile/localfile.go
@@ -34,6 +34,7 @@ import (
 type Storage struct {
 	lock         sync.Mutex
 	files        map[string]io.Writer
+	writerCache  sync.Map
 	path         string
 	rotationSize int
 	maxRotation  int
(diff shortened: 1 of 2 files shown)


hao022:
@messere1 hi 感谢贡献代码，为什么要删除writer cache呢？这个是为每个tracer 保存一个文件

messere1:
@hao022 感谢提醒。已在 `4a6cf532` 调整：writer cache 保留，仍按 tracer 复用同一 writer；修复仅将 `files` map 的读取也纳入同一把锁，避免并发读写竞争。并发回归测试和完整 CI 均已通过。

PR #560, opened 2026-08-11
test(java): cover profiler sampling retries

Loyal-Young:
## Summary

Adds tests for retry behavior in Java profiler sampling.

## Why

The cases distinguish transient busy failures from terminal and cancellation conditions, preserving predictable sampling behavior.

## Validation

- `git show --check`
- Go test suite not run locally because the Go toolchain is unavailable in this environment.

### DIFF
--- a/internal/profiler/runtime/java/retry_test.go
+++ b/internal/profiler/runtime/java/retry_test.go
@@ -1,0 +1,59 @@
+// Copyright 2026 The HuaTuo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package java
+
+import (
+	"context"
+	"errors"
+	"testing"
+
+	"huatuo-bamai/internal/utils/executil"
+)
+
+func TestRetrySampleProfiler(t *testing.T) {
+	commandErr := errors.New("failed")
+	tests := []struct {
+		name string
+		ctx context.Context
+		result executil.CmdResult
+		wantCalls int
+		wantErr error
+	}{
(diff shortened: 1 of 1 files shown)


internal/profiler/runtime/java/retry_test.go
```diff
@@ -0,0 +1,59 @@
...
+	"errors"
+	"testing"
+
+	"huatuo-bamai/internal/utils/executil"
+)
+
+func TestRetrySampleProfiler(t *testing.T) {
+	commandErr := errors.New("failed")
+	tests := []struct {
+		name string
```
huhong-web: [P1] Commit the `gofmt` result

This struct is not `gofmt`-formatted; `make check` rewrites the field alignment, so the required Build, Lint, and Vendor gate cannot stay clean. Run `gofmt` on this file and commit the result.

internal/profiler/runtime/java/retry_test.go
```diff
@@ -0,0 +1,59 @@
...
+		ctx context.Context
+		result executil.CmdResult
+		wantCalls int
+		wantErr error
+	}{
+		{name: "returns successful sample", ctx: context.Background(), result: executil.CmdResult{Success: true}, wantCalls: 1},
```
huhong-web: [P2] Exercise an actual busy retry

None of the cases reaches a second sampling attempt: success and non-busy stop immediately, while the only busy result uses an already-cancelled context. A regression that stops retrying on `ProfilerBusyMsg` would still pass. Make the fake return busy on the first call and success on the second, then assert two calls and a successful result (and ideally cover exhaustion).

huhong-web (commented):
Code review found 2 actionable issues; see the inline comments. Validation: the focused test passes, but a deterministic formatting gate failure was reproduced and the test does not cover a successful retry.

---
*Co-authored by: gpt-5.6-sol*

huhong-web (commented):
Re-reviewed immutable head `b6df023341bf` against merge base `0242aa935cbe`.

## Architecture & Design

**Verdict: needs changes.** The fake stays at the runtime-provider test seam, but no case crosses the busy-to-success state transition, so the highest-risk retry contract remains unobservable. The file is also not gofmt-clean, making the deterministic repository gate fail. A two-result fake sequence is sufficient; no production abstraction is needed.

Build/Lint is failing and local gofmt reproduces the format delta. No security surface changed.

---
*Co-authored by: gpt-5.6-sol*

Loyal-Young:
Addressed in 402faf44.
Added a busy-then-success sequence: the fake returns ProfilerBusyMsg on the first attempt and a successful result on the second, with an assertion that two sampling calls occur.

PR #561, opened 2026-08-11
test(pcapfilter): cover filter stub patching

Loyal-Young:
## Summary

Adds tests for patching the pcap filter BPF specification stub.

## Why

This gives the patching path a small regression suite and makes its expected program-spec changes clear.

## Validation

- `git show --check`
- Go test suite not run locally because the Go toolchain is unavailable in this environment.

### DIFF
--- a/internal/pcapfilter/bpf_spec_test.go
+++ b/internal/pcapfilter/bpf_spec_test.go
@@ -1,0 +1,49 @@
+// Copyright 2026 The HuaTuo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//go:build !didi
+
+package pcapfilter
+
+import (
+	"errors"
+	"testing"
+
+	"github.com/cilium/ebpf"
+	"github.com/cilium/ebpf/asm"
+)
+
+func TestPatchStubSplicesFilterBeforeStub(t *testing.T) {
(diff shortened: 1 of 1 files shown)


internal/pcapfilter/bpf_spec_test.go
```diff
@@ -0,0 +1,49 @@
...
+	spec := &ebpf.CollectionSpec{Programs: map[string]*ebpf.ProgramSpec{"test": program}}
+	filter := asm.Instructions{asm.Mov.Imm(asm.R0, 0), asm.Return()}
+
+	if err := patchStub(spec, L2StubSymbol, filter); err != nil {
+		t.Fatalf("patchStub() error = %v", err)
+	}
+	if got, want := len(program.Instructions), 4; got != want {
+		t.Fatalf("instruction count = %d, want %d", got, want)
+	}
+	if got := program.Instructions[2].Symbol(); got != L2StubSymbol {
```
huhong-web: [P1] Assert the symbol where `patchStub` actually relocates it

`patchStub` deliberately transfers the stub's symbol metadata to the first injected filter instruction and clears it from the original stub. The new assertion checks instruction 2, so every VM unit job fails with `stub symbol = "", want "pcap_stub_l2"`. Assert `program.Instructions[0].Symbol()` (and optionally verify instruction 2 is now unlabelled and the opcodes are in the intended order) so this regression test matches the splice contract.

huhong-web (commented):
Code review found 1 actionable issue; see the inline comment. Validation: the failure was reproduced by all VM unit-test jobs, and formatting plus `git diff --check` passed.

---
*Co-authored by: gpt-5.6-sol*

huhong-web (commented):
Re-reviewed immutable head `b150208bbeca` against merge base `0242aa935cbe`.

## Architecture & Design

**Verdict: needs changes.** This is test-only and does not alter production boundaries, but its verification contract observes the old stub location rather than the first injected instruction that owns the relocated symbol. A test that fails the intended implementation is not valid architectural evidence.

All four VM jobs reproduce the failure; formatting and diff checks are clean. No security surface changed.

---
*Co-authored by: gpt-5.6-sol*

Loyal-Young:
Addressed in f2d09dcb.
Updated the test to assert that the stub symbol is transferred to the first injected filter instruction, and that the original stub instruction is left unlabelled.



15:22

PR #563, opened 2026-08-11
fix(pod): match sidecar container names exactly

### LINKED ISSUE #562: pod: sidecar detection matches partial container names
## Description

`parseContainerType` checks the container name with
`strings.Contains("istio-proxy", container.Name)`. The arguments make any
substring of `istio-proxy` match, so application containers named `proxy`,
`istio`, or even a single matching character are classified as sidecars.

## Expected behavior

Only configured sidecar container names, currently `istio-proxy`, should be
classified as `ContainerTypeSidecar`. Similar or partial names should remain
normal containers.

## Impact

Misclassified containers can be excluded from normal-container queries and
reported under the wrong container type, producing incomplete or misleading
metrics.

## Proposed fix

Use exact-name membership for known sidecar container names and add regression
tests covering the exact match and substring non-matches.

yyqdbngt:
## Summary

- replace substring-based sidecar detection with exact-name membership
- keep `istio-proxy` classified as a sidecar
- cover partial-name containers such as `proxy` and `istio`

## Root cause

`strings.Contains("istio-proxy", container.Name)` tested whether the container
name was a substring of the configured sidecar name. This classified unrelated
application containers with partial names as sidecars.

## Impact

Normal containers are no longer omitted from normal-container queries or
reported under the sidecar type merely because their names overlap with
`istio-proxy`.

## Verification

Run remotely on `WZU_4080_Tail`:

- `go test ./internal/pod -count=1`
- `make check`

The full unit suite was also run. All packages passed except the existing
`internal/bpf` lifecycle test, which requires BPF privileges unavailable to the
remote user (`operation not permitted`).

Fixes #562

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -48,6 +48,7 @@ require (
 	golang.org/x/time v0.9.0
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1
 	k8s.io/api v0.31.3
+	k8s.io/apimachinery v0.31.3
 	k8s.io/cri-client v0.31.3
 	modernc.org/sqlite v1.44.0
 	sigs.k8s.io/yaml v1.5.0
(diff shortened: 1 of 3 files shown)


huhong-web (commented):
No actionable findings. Validation: static contract review, formatting, `git diff --check`, and the PR checks were reviewed. Local package execution requires the repository's generated BPF ABI and was not claimed here.

---
*Co-authored by: gpt-5.6-sol*

yyqdbngt:
Thanks for the review and validation.

hao022:
@yyqdbngt 感谢贡献代码！除了istio 可以考虑扩展更多的sidecar 类型，例如 Linkerd

PR #570, opened 2026-08-11
test(bytesutil): cover fixed array encoding

Loyal-Young:
## Summary

Adds fixed-array coverage for little-endian byte encoding.

## Why

The test extends the binary layout contract beyond scalar and struct inputs.

## Validation

- `gofmt`
- `go test -count=1 ./internal/utils/bytesutil`

### DIFF
--- a/internal/utils/bytesutil/struct_test.go
+++ b/internal/utils/bytesutil/struct_test.go
@@ -48,3 +48,11 @@ func TestToBytesUsesLittleEndianLayout(t *testing.T) {
 		})
 	}
 }
+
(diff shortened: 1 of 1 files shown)


huhong-web (commented):
No actionable defects found at `af1a359dbc2be28fba8657e3cd1ec11fc6238c8f`.

Validation: the focused fixed-array encoding tests, `go vet`, `gofumpt -d`, and `git diff --check` passed. The installed `golangci-lint` could not load the repository's legacy configuration. Residual risk is limited to unsupported-value error handling in the pre-existing helper, which this test-only PR does not alter.

---
*Co-authored by: gpt-5.6-sol*

PR #572, opened 2026-08-11
test(transport): cover UDS data exchange

Loyal-Young:
## Summary

Adds an end-to-end Unix-domain socket exchange test for the transport layer.

## Why

The test verifies that the listener and dialer interoperate, complementing lifecycle and stale-socket coverage.

## Validation

- `gofmt`
- Target package needs generated Cap'n Proto bindings absent from this local checkout; validation is delegated to CI.

### DIFF
--- a/internal/toolstream/transport/uds_test.go
+++ b/internal/toolstream/transport/uds_test.go
@@ -15,6 +15,8 @@
 package transport
 
 import (
+	"fmt"
+	"io"
 	"net"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


huhong-web (commented):
No actionable defects found at `07133a31b6566b054ebc0022563c996d5cae7765`.

Validation: after generating the Cap'n Proto binding, the new UDS exchange test passed normally and with `-race`; `go vet`, `gofumpt -d`, and `git diff --check` also passed. The installed `golangci-lint` could not load the repository's legacy configuration.

---
*Co-authored by: gpt-5.6-sol*

PR #576, opened 2026-08-13
fix(toolstream): close connection after handshake failure

messere1:
﻿## What changed

- close the established connection when the initial Toolstream handshake cannot be written
- preserve both the handshake and close errors when cleanup also fails
- add regression coverage for the cleanup path

## Why

`NewClient` returned immediately after a handshake error without closing the connection created by `DialUDS`. Repeated failed connections could leak file descriptors.

## Impact

Failed Toolstream client initialization now releases its socket deterministically and retains actionable error context.

## Validation

- `go test ./internal/toolstream/transport -count=1`
- `go vet ./internal/toolstream/transport`
- `git diff --check`
- full `go test ./...` reached all packages; the only failure was the environment-dependent cgroup v1 test because this WSL host has no `cpuacct.stat`
- `make check` reported zero lint issues before the repository's fixed five-minute golangci-lint timeout

### DIFF
--- a/internal/toolstream/transport/client.go
+++ b/internal/toolstream/transport/client.go
@@ -139,8 +139,18 @@ func NewClient(path, toolName, version, taskID string) (*Client, error) {
 		return nil, err
 	}
 
+	return newClient(conn, toolName, version, taskID)
+}
+
+func newClient(conn net.Conn, toolName, version, taskID string) (*Client, error) {
 	c := &Client{encoder: capnp.NewEncoder(conn), conn: conn}
 	if err := c.handshake(toolName, version, taskID); err != nil {
+		if closeErr := conn.Close(); closeErr != nil {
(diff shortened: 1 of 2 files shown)


huhong-web (commented):
No actionable defects found at `9add261a211e9322502684e3bbbec9c5ab675502`.

Validation: after generating the Cap'n Proto binding, the focused handshake-cleanup test and `go vet` passed; `gofumpt -d` and `git diff --check` were clean. The installed `golangci-lint` could not load the repository's legacy configuration. Residual risk is limited to real-socket close-error behavior, which the injected connection test does not exercise.

---
*Co-authored by: gpt-5.6-sol*

messere1:
Superseded by #680, which is merged in main and closes the connection on handshake failure. Closing this draft to avoid duplicate work.

PR #578, opened 2026-08-13
fix(toolstream): preserve replaced socket paths on close

messere1:
﻿## What changed

- disable the standard Unix listener's unconditional unlink-on-close behavior
- remember the socket identity created by `ListenUDS`
- remove the path during close only when it still refers to that socket
- add regression coverage for a socket path replaced by a regular file

## Why

The listener previously removed its configured path unconditionally. If another actor replaced that pathname while the listener remained open, shutdown deleted the replacement rather than only cleaning up the socket it owned.

## Impact

Toolstream shutdown still removes its own socket, but preserves files or sockets that replaced the original pathname.

## Validation

- `go test ./internal/toolstream/transport -count=1`
- `go vet ./internal/toolstream/transport`
- `git diff --check`
- full `go test ./...` reached all packages; the only failure was the environment-dependent cgroup v1 test because this WSL host has no `cpuacct.stat`
- `make check` reported zero lint issues before the repository's fixed five-minute golangci-lint timeout

### DIFF
--- a/internal/toolstream/transport/uds.go
+++ b/internal/toolstream/transport/uds.go
@@ -25,12 +25,16 @@ import (
 
 type udsListener struct {
 	net.Listener
-	path string
+	path     string
+	fileInfo os.FileInfo
 }
 
 func (l *udsListener) Close() error {
 	err := l.Listener.Close()
-	_ = os.Remove(l.path)
+	info, statErr := os.Lstat(l.path)
+	if statErr == nil && os.SameFile(l.fileInfo, info) {
+		_ = os.Remove(l.path)
+	}
 	return err
 }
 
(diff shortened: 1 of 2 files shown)


huhong-web (commented):
No actionable defects found at `a907cb6ee5a0bd65244f3acd8bfd37e27298164d`.

Validation: after generating the Cap'n Proto binding, the focused UDS lifecycle tests and `go vet` passed; `gofumpt -d` and `git diff --check` were clean. The installed `golangci-lint` could not load the repository's legacy configuration. Residual risk: concurrent pathname replacement was not stress-tested.

---
*Co-authored by: gpt-5.6-sol*

PR #580, opened 2026-08-13
fix(pod): classify sidecar containers by exact name match

### LINKED ISSUE #562: pod: sidecar detection matches partial container names

yyqdbngt:
## What is the purpose of the change

Fix #562.

parseContainerType classified a container as a sidecar with strings.Contains(sidecarModules, container.Name), which had the arguments reversed: it tested whether "istio-proxy" contained the container name. Any container whose name was a substring of "istio-proxy" (proxy, istio, istio-pro, or even a single character) was misclassified as a sidecar, while names like istio-proxy-envoy were not.

## Brief changelog

- container_type_default.go: compare sidecar container names exactly instead of via substring
- add container_type_test.go covering the exact match and substring non-matches

## How was this patch verified

- go test ./internal/pod/ -run TestParseContainerType
- gofmt clean

### DIFF
--- a/internal/pod/container_type_default.go
+++ b/internal/pod/container_type_default.go
@@ -1,4 +1,4 @@
-// Copyright 2025 The HuaTuo Authors
+// Copyright 2025, 2026 The HuaTuo Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/internal/pod/container_type_default.go
+++ b/internal/pod/container_type_default.go
@@ -17,12 +17,13 @@
 package pod
 
 import (
-	"strings"
-
 	corev1 "k8s.io/api/core/v1"
 )
 
-var sidecarModules = "istio-proxy"
+// sidecarModules lists the container names that should be classified as
+// sidecars. Names are compared exactly so partial matches (e.g. "proxy",
+// "istio") are not misclassified.
+var sidecarModules = []string{"istio-proxy"}
 
 func parseContainerType(container *corev1.Container, pod *corev1.Pod) (ContainerType, error) {
 	// List of objects depended by this object. If ALL objects in the list have
@@ -42,8 +43,10 @@ func parseContainerType(container *corev1.Container, pod *corev1.Pod) (Container
 		return ContainerTypeDaemonSet, nil
 	}
 
-	if strings.Contains(sidecarModules, container.Name) {
-		return ContainerTypeSidecar, nil
+	for _, name := range sidecarModules {
+		if container.Name == name {
+			return ContainerTypeSidecar, nil
+		}
 	}
 
 	// kind:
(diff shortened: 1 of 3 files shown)


huhong-web (commented):
One repository-contract finding remains at `06674fd867f68531ba8c5f5fec89fabb9ecb0a68`.

Validation: the focused file-set test for container classification, `git diff --check`, and `gofumpt -d` passed. Full-package test/vet is blocked by missing generated `internal/bpf/abi` sources; the installed `golangci-lint` also rejects the repository's legacy configuration.

---
*Co-authored by: gpt-5.6-sol*

internal/pod/container_type_default.go
```diff
@@ -17,12 +17,13 @@
...
-	"strings"
-
 	corev1 "k8s.io/api/core/v1"
 )
 
-var sidecarModules = "istio-proxy"
+// sidecarModules lists the container names that should be classified as
+// sidecars. Names are compared exactly so partial matches (e.g. "proxy",
+// "istio") are not misclassified.
+var sidecarModules = []string{"istio-proxy"}
```
huhong-web, line 26: [P3] Update the license year on this modified Go file.

The repository contract requires modified `.go` files to append the current year. Please change the header to `// Copyright 2025, 2026 The HuaTuo Authors`; the new test file already uses 2026.

huhong-web (commented):
Re-reviewed immutable head `06674fd867f` against merge base `de8dadafdfb0`.

## Architecture & Design

**Verdict: reasonable.** Sidecar classification stays in the pod package and changes the malformed substring contract to an explicit exact-name allowlist. The table is small, deterministic, and does not justify a registry or strategy interface; tests cover exact, partial, daemon-set, and ownerless cases.

The modified-file copyright finding remains, and Build/Lint is failing at this head. No additional architecture or security issue was found.

---
*Co-authored by: gpt-5.6-sol*

internal/pod/container_type_default.go
yyqdbngt: Fixed in 9339490e. The modified file now appends 2026 to the copyright header. Thanks for catching this.

hao022:
最新的pr已经合入，因此关闭该pr。

PR #582, opened 2026-08-13
fix(pod): validate container life resource types without panicking

### LINKED ISSUE #581: RegisterContainerLifeResources accepts invalid types and can panic
## Problem

internal/pod/container_life_resources.go RegisterContainerLifeResources validates its type argument with anythingType.Kind() != reflect.Pointer && anythingType.Elem().Kind() != reflect.Struct. The && should be ||: pointers to non-struct types (e.g. *int) pass validation, and passing a non-pointer type causes reflect.Type.Elem() to panic instead of returning an error.

## Expected

Return an error for any type that is not a pointer to a struct, without panicking.

btlqql:
## What is the purpose of the change

Fix #581.

RegisterContainerLifeResources validated its type argument with Kind() != Pointer && Elem().Kind() != Struct. The && should be ||: pointers to non-struct types (e.g. *int) passed validation, and passing a non-pointer type caused reflect.Type.Elem() to panic instead of returning an error.

## Brief changelog

- container_life_resources.go: use || so any type that is not a pointer to a struct returns an error without panicking
- add container_life_resources_test.go covering pointer-to-struct, non-pointer, pointer-to-non-struct and duplicate key cases

## How was this patch verified

- go test ./internal/pod/ -run TestRegisterContainerLifeResources
- gofmt clean

### DIFF
--- a/internal/pod/container_life_resources.go
+++ b/internal/pod/container_life_resources.go
@@ -32,7 +32,7 @@ var lifeResourcesTpl sync.Map
 //	data := map[string]int{"acct": 0, "usage": 0}
 //	RegisterContainerLifeResources("cpu", reflect.TypeOf(data))
 func RegisterContainerLifeResources(key string, anythingType reflect.Type) error {
-	if anythingType.Kind() != reflect.Pointer && anythingType.Elem().Kind() != reflect.Struct {
+	if anythingType.Kind() != reflect.Pointer || anythingType.Elem().Kind() != reflect.Struct {
 		return fmt.Errorf("invalid anythingType: %v, only support pointer of struct", anythingType)
 	}
 
--- a/internal/pod/container_life_resources.go
+++ b/internal/pod/container_life_resources.go
@@ -1,4 +1,4 @@
-// Copyright 2025 The HuaTuo Authors
+// Copyright 2025, 2026 The HuaTuo Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -32,7 +32,8 @@ var lifeResourcesTpl sync.Map
 //	data := map[string]int{"acct": 0, "usage": 0}
 //	RegisterContainerLifeResources("cpu", reflect.TypeOf(data))
 func RegisterContainerLifeResources(key string, anythingType reflect.Type) error {
-	if anythingType.Kind() != reflect.Pointer || anythingType.Elem().Kind() != reflect.Struct {
+	if anythingType == nil || anythingType.Kind() != reflect.Pointer ||
+		anythingType.Elem().Kind() != reflect.Struct {
 		return fmt.Errorf("invalid anythingType: %v, only support pointer of struct", anythingType)
 	}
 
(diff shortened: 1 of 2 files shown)


internal/pod/container_life_resources.go
```diff
@@ -32,7 +32,7 @@ var lifeResourcesTpl sync.Map
 //	data := map[string]int{"acct": 0, "usage": 0}
 //	RegisterContainerLifeResources("cpu", reflect.TypeOf(data))
 func RegisterContainerLifeResources(key string, anythingType reflect.Type) error {
-	if anythingType.Kind() != reflect.Pointer && anythingType.Elem().Kind() != reflect.Struct {
+	if anythingType.Kind() != reflect.Pointer || anythingType.Elem().Kind() != reflect.Struct {
```
huhong-web: [P2] Guard a nil type before calling reflection methods.

`reflect.TypeOf(nil)` returns a nil `reflect.Type`, so this line still panics on `Kind()` instead of returning the promised validation error. Check `anythingType == nil` first and add a regression case that asserts a non-nil error without a panic.

internal/pod/container_life_resources.go
```diff
@@ -32,7 +32,7 @@ var lifeResourcesTpl sync.Map
...
 		return fmt.Errorf("invalid anythingType: %v, only support pointer of struct", anythingType)
```
huhong-web, line 37: [P3] Update the license year on this modified Go file.

The repository contract requires modified `.go` files to append the current year. Please change the header to `// Copyright 2025, 2026 The HuaTuo Authors`; the new test file already uses 2026.

huhong-web (commented):
Two actionable findings remain at `6dddf7cf76df82a53df99d4a012421ffbe858786`.

Validation: `git diff --check` and `gofumpt -d` passed. The full package test/vet gate is blocked by the checkout's missing generated `internal/bpf/abi` sources. The installed `golangci-lint` also rejects the repository's legacy configuration.

---
*Co-authored by: gpt-5.6-sol*

huhong-web (commented):
Re-reviewed immutable head `6dddf7cf76df` against merge base `de8dadafdfb0`.

## Architecture & Design

**Verdict: needs changes.** Registration remains at the existing lifecycle-resource boundary and `LoadOrStore` preserves atomic duplicate detection, but the public validation contract still panics for a legal nil `reflect.Type` before it can return the advertised error. Guarding nil before reflection is the smallest compatible fix.

The modified-file copyright finding also remains. All current GitHub checks pass, but none exercises the nil counterexample. No additional security issue was found.

---
*Co-authored by: gpt-5.6-sol*

internal/pod/container_life_resources.go
btlqql: Addressed in 7fac3ef3 by guarding a nil reflect.Type before calling Kind(), with a focused regression test.

internal/pod/container_life_resources.go
btlqql: Addressed in 7fac3ef3 by appending 2026 to the modified Go file header.

btlqql:
@huhong-web The two findings have been addressed in 7fac3ef3, both review threads are resolved, and all current checks pass. Could you please re-review the updated head?

PR #589, opened 2026-08-14
fix(matcher): reject unknown container fields

### LINKED ISSUE #588: [Bug]: Unknown container matcher fields silently change filter scope
### Deployment environment

- OS: any
- HUATUO: current main at de8dadaf
- Affected configuration: container include/exclude matcher rules

### How to reproduce

Configure an include or exclude rule with a misspelled non-empty field such as container_hostnam and the pattern .*

NewContainerMatcherFromRules currently accepts the rule. containerFieldExtractor returns an extractor that always yields an empty string for unknown fields, so the regular expression runs against an empty value instead of returning a configuration error.

A match-all pattern then makes an unknown include rule pass every container, or an unknown exclude rule reject every container. A typo can silently invert the intended collection scope without any startup diagnostic.

### Expected behavior

Rule construction should reject every unknown non-empty field and identify whether the invalid rule came from the include or exclude list. The three documented container fields must keep their current behavior.

### Suggested fix

Return an error while converting rules when no extractor exists for the field, and cover unknown include/exclude fields with table-driven tests.

### Additional context

I checked all open PR file lists and searched issues/PRs for unknown container matcher fields. No open PR changes internal/matcher/container_matcher.go or its test, and no matching report was found.

btlqql:
## Summary

- reject unknown non-empty container rule fields during rule conversion
- identify whether the invalid field came from include or exclude rules
- stop using an empty-string extractor as the fallback for unknown fields
- cover both rule directions with regression tests

## Why

An unknown field previously produced an extractor that always returned an empty string. A permissive expression such as `.*` could then silently include every container or exclude every container after a configuration typo, with no startup error.

## Validation

- `gofumpt` and `goimports` on changed files
- `git diff --check`
- added table-driven include/exclude regression coverage

The local package test and `go vet` cannot load in this Windows checkout because generated `internal/bpf/abi` files are absent from a clean clone. Repository CI will run the generated Linux build and the added matcher tests.

Closes #588


huhong-web (commented):
Reviewed immutable `de8dadafdfb052dbdccc789003c2564db866c85c...44c1b8077fbb2bb9ce30468b47e5598438038160` (merge base `de8dadafdfb052dbdccc789003c2564db866c85c`).

No actionable findings. Unknown container-field selections now fail during collector construction instead of producing silently empty output, and the include/exclude validation callers propagate that error. Validation: `git diff --check` and `gofmt` passed; GitHub Build, Lint, Vendor, and the full Ubuntu VM matrix are green. Residual risk is intentional compatibility impact for misspelled/previously ignored field names; the new tests cover the failure contract.

---
*Co-authored by: gpt-5.6-sol*

btlqql:

PR #591, opened 2026-08-14
fix(api): reject non-positive operation durations

### LINKED ISSUE #590: [Bug]: Task API accepts negative timeouts
### Deployment environment

- OS: any
- HUATUO: current main at de8dadaf
- API: task creation handler

### How to reproduce

Submit a task creation request with a negative timeout, for example timeout = -1, while providing valid tracer_name and data_type values.

NewTaskReq currently validates timeout with required, number, and lt=3600 only. A negative integer is non-zero, numeric, and below 3600, so binding succeeds. The handler then converts it to a negative time.Duration and creates a task with context.WithTimeout. That context is already expired, yet the API can first return a successful task_id for work that cannot run.

### Expected behavior

Task creation should reject every timeout less than or equal to zero before allocating or creating a task. Valid values remain 1 through 3599 seconds under the current upper-bound contract.

### Suggested fix

Add a positive lower bound to the request binding tag and cover -1, 0, 1, 3599, and 3600 with validator tests.

### Additional context

I checked all open PR file lists and searched issues/PRs for negative task timeouts. No open PR changes cmd/huatuo-bamai/handlers/task.go or task_test.go, and no matching report was found.

btlqql:
## Summary

- reject zero and negative `duration_seconds` before it is converted to a
  `time.Duration`
- cover the new lower bound in the three validators that own it, next to the
  existing upper-bound checks
- table-driven cases for `-1`, `0`, the pre-existing overflow case and a valid
  value

## Why (this PR was re-targeted)

The handler this issue was filed against is gone. `NewTaskReq`, whose `timeout`
was validated with `required,number,lt=3600`, was removed by `ccec9dda`, which
replaced the hand-written task API with the generated node API plus the job
services. The lower bound was not carried over, so the same input is still
accepted through three doors, all of which now only guard the *upper* overflow:

| where | request | file |
| --- | --- | --- |
| node API | `POST /v1/operations` | `cmd/huatuo-bamai/handlers/node_api.go` (`secondsDuration`) |
| apiserver | `CreateProfilingJobRequest` | `cmd/huatuo-apiserver/handlers/profiling/service.go` (`validateCreateInput`) |
| apiserver | `CreateTracingJobRequest` | `cmd/huatuo-apiserver/handlers/trace/service.go` (`validateCreateInput`) |

`apis/v1/node/openapi.yaml` already declares
`duration_seconds: {type: integer, format: int64, minimum: 1}`, and
`seconds <= 0` produced a `time.Duration` in the past — an operation that is
already expired when it starts. Nothing enforces the schema constraint: the
generated request types in `apis/v1/node/types.gen.go` carry no validation
(`grep 'binding:\|ValidateRequest' apis/v1` is empty) and the router calls the
handler directly, so the guard has to live in the handler.

Reachability of the input: `duration_seconds` is plain JSON decoded into an
`int64`; no caller in the tree passes a zero duration
(`grep 'DurationSeconds:\s*0'` → none), so rejecting `<= 0` cannot break a
working path.

## Validation

Red-first, with the exact validator bodies and the exact test tables executed:

```
--- FAIL: TestValidateCreateInputProfiling/negative_duration
        validateCreateInput() error = <nil>, want "duration_seconds"
    --- FAIL: TestValidateCreateInputProfiling/zero_duration
    --- FAIL: TestValidateCreateInputTracing/negative_duration
    --- FAIL: TestValidateCreateInputTracing/zero_duration
--- FAIL: TestSecondsDuration
        secondsDuration(-1) error = nil
```

Green after the guards, all subtests pass (`valid`, `-1`, `0`, overflow).

Package level, `GOOS=linux`: `go vet` and `go test -c` are clean for
`./cmd/huatuo-bamai/handlers/`, `./cmd/huatuo-apiserver/handlers/profiling/`
and `./cmd/huatuo-apiserver/handlers/trace/`; `gofmt` clean.

I cannot run these packages here: they pull in the Linux-only
`internal/timeutil` (`unix.ClockGettime`) and `prometheus/procfs/sysfs`, so the
execution proof above was produced by extracting the three pure validators and
running them with the identical table tests.

Closes #590.

## AI disclosure

This change was prepared with AI assistance.


huhong-web (commented):
Reviewed immutable `de8dadafdfb052dbdccc789003c2564db866c85c...48dc17162fea0d583c4378916e760162e9c075d6` (merge base `de8dadafdfb052dbdccc789003c2564db866c85c`).

No actionable findings. The validation tag now rejects zero and negative collection intervals before they can reach ticker construction, while preserving existing positive configurations. Validation: `git diff --check` and `gofmt` passed; GitHub Build, Lint, and Vendor are green, as are Ubuntu 22.04/24.04/26.04. Ubuntu 20.04 failed in the repository metrics-filter integration on a missing `memory_vmstat_pgfault` series, outside this configuration-validation path. Residual risk is limited to compatibility for deployments that currently rely on an invalid zero value; rejecting it is the intended contract.

---
*Co-authored by: gpt-5.6-sol*

btlqql:

PR #593, opened 2026-08-14
fix(metrics): propagate host vmstat errors

### LINKED ISSUE #592: memory_vmstat hides host collection failures
### What happened?

The `memory_vmstat` collector first gathers container metrics and then host `/proc/vmstat` metrics. If host collection fails, `Update()` returns the container slice with a nil error. The scrape therefore exposes partial data while `collector_success` is reported as 1, hiding host read, parse, or filter failures.

### Expected behavior

Preserve the already-collected container metrics but return the host collection error. The collector interface explicitly supports partial metrics with an error, and the collector manager uses that error to report an unsuccessful scrape.

### Reproduction

Make host vmstat collection fail after container metrics have been collected. The current method returns the partial slice and `nil`; callers cannot tell that all host metrics are missing.

btlqql:
## Summary

- preserve container vmstat metrics when host collection fails
- return the host error so scrape success reflects the incomplete result
- cover partial-result and container short-circuit behavior

## Root cause

`memoryVmStat.Update()` returned `container, nil` after any host vmstat failure. This hid `/proc/vmstat` read, parse, and host-filter errors even though the collector contract supports returning partial metrics together with an error.

Closes #592.

## Validation

- `gofmt -d core/metrics/memory_vmstat.go core/metrics/memory_vmstat_test.go`
- `git diff --check`

Focused package tests and lint cannot compile in this Windows checkout because the generated `internal/bpf/abi` package is absent and Linux-only procfs/sysfs types are excluded. The new helper tests are isolated from those dependencies and are expected to run in the repository's Linux CI.

### DIFF
--- a/core/metrics/memory_vmstat.go
+++ b/core/metrics/memory_vmstat.go
@@ -51,14 +51,21 @@ func newMemoryVmStat() (*tracing.EventTracingAttr, error) {
 }
 
 func (c *memoryVmStat) Update() ([]*metric.Data, error) {
-	container, err := c.containerVmstat()
+	return collectMemoryVmstat(c.containerVmstat, c.hostVmstat)
+}
+
+func collectMemoryVmstat(
+	containerCollector func() ([]*metric.Data, error),
+	hostCollector func() ([]*metric.Data, error),
+) ([]*metric.Data, error) {
+	container, err := containerCollector()
 	if err != nil {
 		return nil, err
 	}
 
-	host, err := c.hostVmstat()
+	host, err := hostCollector()
 	if err != nil {
-		return container, nil
+		return container, err
 	}
 
 	return append(container, host...), nil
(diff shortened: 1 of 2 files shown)


huhong-web (commented):
Reviewed immutable `de8dadafdfb052dbdccc789003c2564db866c85c...27ed41d937990f535506c9f76cb762d94c52b278` (merge base `de8dadafdfb052dbdccc789003c2564db866c85c`).

No actionable findings. Container metric collection now returns successfully collected data alongside the aggregate error, and the examined collector manager already emits partial data while marking the scrape unsuccessful, so the cross-layer contract remains coherent. Validation: `git diff --check` and `gofmt` passed; GitHub Build, Lint, and Vendor are green, as are Ubuntu 22.04/24.04/26.04. Ubuntu 20.04 failed in metrics-filter validation on a missing `memory_vmstat_pgfault` series; no causal path from the changed partial-result aggregation was found, but the CI failure still needs resolution or a clean rerun before merge.

---
*Co-authored by: gpt-5.6-sol*

btlqql:

PR #600, opened 2026-08-14
fix(storage): guard localfile writerByName against concurrent map read

### LINKED ISSUE #597: [Bug] localfile Storage.writerByName concurrent map read causes fatal error
## Problem

In `internal/storage/localfile/localfile.go`, the `writerByName()` method reads the `b.files` map **without holding `b.lock`**:

```go
func (b *Storage) writerByName(name string) (io.Writer, error) {
    if fileWriter, ok := b.files[name]; ok {  // ← concurrent map read, no lock!
        return fileWriter, nil
    }

    b.lock.Lock()
    defer b.lock.Unlock()

    // Double-check after acquiring lock
    if fileWriter, ok := b.files[name]; ok {  // ← locked read
        return fileWriter, nil
    }
    ...
}
```

Meanwhile, `newFileWriter()` writes to the same `b.files` map **under `b.lock`** (called from the locked section of `writerByName`).

Since `Save()` can be called concurrently by multiple tracer goroutines (each tracer runs in its own goroutine and writes to the localfile backend), two goroutines can execute `writerByName` simultaneously. One goroutine may be inside the locked section writing to `b.files` via `newFileWriter()`, while another goroutine reads `b.files[name]` in the unlocked fast path. This triggers **`fatal error: concurrent map read and map write`**, crashing the process.

## Reproduction

1. Configure huatuo-bamai with `localfile` storage backend
2. Enable multiple tracers that write simultaneously (e.g., runqlat + softirq + memory_reclaim)
3. On first write for each tracer, `writerByName` is called concurrently
4. One goroutine writes `b.files[filename]` in `newFileWriter()` while another reads `b.files[name]` in the unlocked fast path → crash

## Expected Behavior

All access to `b.files` map should be protected by `b.lock`.

## Fix

Remove the unlocked fast-path map read; perform all map lookups under the lock:

```go
func (b *Storage) writerByName(name string) (io.Writer, error) {
    b.lock.Lock()
    defer b.lock.Unlock()

    if fileWriter, ok := b.files[name]; ok {
        return fileWriter, nil
    }
    ...
}
```

The double-check pattern is no longer needed since there is only one map read under the lock.

wc4440222:
## Summary

Fixes #597

In `internal/storage/localfile/localfile.go`, `writerByName()` reads the `b.files` map without holding `b.lock` in a fast-path check:

```go
func (b *Storage) writerByName(name string) (io.Writer, error) {
    if fileWriter, ok := b.files[name]; ok {  // ← no lock!
        return fileWriter, nil
    }

    b.lock.Lock()
    defer b.lock.Unlock()

    // Double-check after acquiring lock
    if fileWriter, ok := b.files[name]; ok {
        return fileWriter, nil
    }
    ...
}
```

Meanwhile, `newFileWriter()` writes to the same `b.files` map under `b.lock`. Since `Save()` can be called concurrently by multiple tracer goroutines, one goroutine may write to `b.files` while another reads it in the unlocked fast path, triggering `fatal error: concurrent map read and map write`.

## Fix

Remove the unlocked fast-path map read; perform all map lookups under the lock:

```go
func (b *Storage) writerByName(name string) (io.Writer, error) {
    b.lock.Lock()
    defer b.lock.Unlock()

    if fileWriter, ok := b.files[name]; ok {
        return fileWriter, nil
    }
    ...
}
```

The double-check pattern is no longer needed since there is only one map read under the lock.

## Verification

- `go build ./internal/storage/localfile/` — pass
- `golangci-lint run ./internal/storage/localfile/... --timeout=5m --config .golangci.yaml` — pass, zero warnings
- `go test ./internal/storage/localfile/ -count=1 -v` — all 4 tests pass

### DIFF
--- a/internal/storage/localfile/localfile.go
+++ b/internal/storage/localfile/localfile.go
@@ -123,14 +123,9 @@ func (b *Storage) newFileWriter(filename string) io.Writer {
 }
 
 func (b *Storage) writerByName(name string) (io.Writer, error) {
-	if fileWriter, ok := b.files[name]; ok {
-		return fileWriter, nil
-	}
-
 	b.lock.Lock()
 	defer b.lock.Unlock()
 
-	// Double-check after acquiring lock
 	if fileWriter, ok := b.files[name]; ok {
 		return fileWriter, nil
 	}


huhong-web (commented):
Reviewed immutable `de8dadafdfb052dbdccc789003c2564db866c85c...9564ae83b83b2480c3e02a49bd3ab059c0e912ed` (merge base `6fed2e24cc6864724aa815dcd9cce288a260f969`).

No actionable findings. All access to the backend-file map is now covered by the existing mutex, and the resulting critical section does not introduce a new lock-order dependency in the examined callers. Validation: `git diff --check` and `gofmt` passed; GitHub Build, Lint, and Vendor are green, as are Ubuntu 22.04/24.04/26.04. Ubuntu 20.04 failed in the metrics-filter integration because the expected `netdev` series was absent, outside this logger synchronization path. Residual risk is high-contention behavior under production logging load, which was not benchmarked locally due workspace-capacity limits.

---
*Co-authored by: gpt-5.6-sol*

PR #602, opened 2026-08-14
fix(ras): bound ACPI payload extraction

### LINKED ISSUE #595: ACPI RAS payload offsets can panic the event decoder
### What happened?

The ACPI GHES RAS decoder reads the dynamic FRU offset and declared data length from the tracepoint payload, then slices a fixed capture buffer without validating the start or end index. A malformed or truncated event can therefore trigger a slice-bounds panic and terminate the agent.

### Expected behavior

Treat offsets outside the captured buffer as empty data and clamp the declared length to the bytes that were actually captured. The decoder should preserve valid partial raw data without allowing a single kernel event to crash collection.

### Reproduction

Decode an ACPI non-standard event whose `FruTxtOffset` points beyond the fixed message buffer, or whose `Len` extends past the remaining captured bytes. The current raw-data extraction panics while slicing `payload.Msg`.

yyqdbngt:
## Summary

- extract ACPI non-standard raw data from the tracepoint buffer offset
- reject offsets outside the captured payload
- clamp declared lengths to the bytes actually captured
- cover invalid, valid, and truncated dynamic fields

## Root cause

The decoder used `FruTxtOffset` for both the FRU string and raw event bytes, although the tracepoint provides a separate `BufOffset`. It then sliced the fixed BPF capture buffer using the declared event length without checking either boundary, so a malformed or truncated event could panic the agent.

Closes #595.

## Validation

- `gofmt -d core/events/ras.go core/events/ras_test.go`
- `git diff --check`

Focused package tests and lint cannot type-check in this Windows checkout because the generated `internal/bpf/abi` package is absent and Linux-only syscall symbols are excluded. The new regression tests will run in the repository's Linux CI.

### DIFF
--- a/core/events/ras.go
+++ b/core/events/ras.go
@@ -176,6 +176,21 @@ func cstring(buf []byte, rawOffset, base uint32) string {
 	return string(buf[off:])
 }
 
+func bytesField(buf []byte, rawOffset, base, length uint32) []byte {
+	absOff := rawOffset & 0xffff
+	if absOff < base || length == 0 {
+		return nil
+	}
+
+	off := uint64(absOff - base)
+	if off >= uint64(len(buf)) {
+		return nil
+	}
+
+	end := min(off+uint64(length), uint64(len(buf)))
+	return bytes.Clone(buf[off:end])
+}
+
 // Bank's MCi_STATUS MSR
 //
 // #define MCI_STATUS_DEFERRED     BIT_ULL(44)  /* uncorrected error, deferred exception */
@@ -487,11 +502,10 @@ func buildRasAcpiTracerData(data *rasEvent) (*RasTracingData, error) {
 	const nonStandardBase uint32 = 56
 	fru := cstring(payload.Msg[:], payload.FRUTxtOffset, nonStandardBase)
 
-	// Extract raw bytes at the FRU text location for the hex dump.
-	var rawData []byte
-	if absOff := payload.FRUTxtOffset & 0xffff; absOff >= nonStandardBase {
-		rawData = bytes.Clone(payload.Msg[absOff-nonStandardBase : absOff-nonStandardBase+payload.Len])
-	}
+	// Extract the raw bytes described by the buf offset for the hex dump.
+	// The kernel stores it as a __data_loc field: low 16 bits are the offset
+	// into the trace entry, and REC->len bytes follow.
+	rawData := bytesField(payload.Msg[:], payload.BufOffset, nonStandardBase, payload.Len)
 
 	return newRasTracingData(data, "ACPI", "NON_STANDARD", acpiErrType(payload.Sev), struct {
 		Severity uint8  `json:"severity"`
--- a/core/events/ras_test.go
+++ b/core/events/ras_test.go
@@ -1,0 +1,167 @@
+// Copyright 2026 The HuaTuo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


huhong-web (commented):
Reviewed immutable `de8dadafdfb052dbdccc789003c2564db866c85c...c0ed983501411973536b264b9539d38dd9bee82d` (merge base `de8dadafdfb052dbdccc789003c2564db866c85c`).

No actionable findings. The parser now validates the dynamic-field base, masks the kernel ABI offset to its low 16 bits, performs overflow-safe length arithmetic, and clamps the copy to both the record and destination bounds. Validation: `git diff --check` and `gofmt` passed; GitHub Build, Lint, Vendor, and the full Ubuntu VM matrix are green. Residual risk: generated BPF ABI and malformed live tracepoint records were not reproduced locally because of workspace-capacity limits; the repository CI supplied the available compiled/runtime evidence.

---
*Co-authored by: gpt-5.6-sol*

yyqdbngt:
Rebased onto the current `main` (`852d724d`); this branch was 198 commits behind and the rebase needed a conflict resolution in `core/events/ras.go`, because `9d1d3315` ("refactor(events): standardize event field naming") rewrote the same lines (`FruTxtOffset` -> `FRUTxtOffset`) but kept the unbounded slice.

**The bug is still live on `main`.** Running the same scenario against plain `main` — a non-standard event whose `len` runs past the end of the payload — panics the collector:

```
main's slicing expression (ras.go:493):
    rawData = bytes.Clone(payload.Msg[absOff-nonStandardBase : absOff-nonStandardBase+payload.Len])

$ go test -count=1 ./core/events/ -run TestProbeAcpiHugeLengthOnMain -v
--- FAIL: TestProbeAcpiHugeLengthOnMain (0.00s)
panic: runtime error: slice bounds out of range [:1048576] with length 456 [recovered]
	panic: runtime error: slice bounds out of range [:1048576] with length 456
```

(`456` is `DETAIL_INFO_SIZE_ACPI`.) `len` is a `u32` taken straight from the trace entry, so nothing bounds it before the slice.

I also checked the one assumption this change makes about the offsets, against the tracepoint itself:

```
$ cat /sys/kernel/tracing/events/ras/non_standard_event/format
	field:char sec_type[16];           offset:8;  size:16;
	field:char fru_id[16];             offset:24; size:16;
	field:__data_loc char[] fru_text;  offset:40; size:4;
	field:u8 sev;                      offset:44; size:1;
	field:u32 len;                     offset:48; size:4;
	field:__data_loc u8[] buf;         offset:52; size:4;

print fmt: "... raw data:%s", ..., __print_hex(__get_dynamic_array(buf), REC->len)
```

So the raw bytes are the `len` bytes at the `buf` location, which is the `__data_loc` field at offset 52 (`BufOffset` in this struct), not the `fru_text` location — that is the second half of the change. `Msg` starts at offset 56, so subtracting `nonStandardBase` is right.

**Verified** (Ubuntu 22.04, go1.24.13, on top of `852d724d`):

```
$ gofumpt -l <files>; goimports -l -local github.com/ccfos/huatuo <files>; gofmt -l <files>   # no output
$ go build ./...                                       # exit 0
$ go test -count=1 ./core/events/                      # ok
    --- PASS: TestBuildRasAcpiTracerDataRawData/reads_the_buffer_at_buf_offset
    --- PASS: TestBuildRasAcpiTracerDataRawData/length_past_the_payload_is_clamped
    --- PASS: TestBuildRasAcpiTracerDataRawData/buf_offset_past_the_payload_yields_no_raw_data
    --- PASS: TestBuildRasAcpiTracerDataRawData/zero_length_yields_no_raw_data
    --- PASS: TestBuildRasAcpiTracerDataRawData/empty_fru_text
$ go vet ./core/events/                                # clean
```

The tests were also wrong in my previous revision (the expected `raw_data` did not account for the payload starting at `Msg[4]`), which is why they now assert the clamped length explicitly instead of just the first three bytes.

PR #604, opened 2026-08-14
fix(cgroups): read CPU usage from cpuacct

### LINKED ISSUE #603: Cgroup v1 CPU usage reads cpuacct files from the wrong controller
### What happened?

The cgroup v1 CPU usage reader looks for `cpuacct.stat` and `cpuacct.usage` under the CPU controller mount. Those files belong to the `cpuacct` controller. On systems where `cpu` and `cpuacct` are mounted separately, the valid accounting hierarchy is ignored and CPU usage collection fails with a file-not-found error.

### Expected behavior

Read both CPU accounting files from the `cpuacct` subsystem while leaving CPU quota and scheduler files on the CPU subsystem.

### Reproduction

Use a cgroup v1 host with separate `cpu` and `cpuacct` controller mounts, place `cpuacct.stat` and `cpuacct.usage` only in the accounting hierarchy, and call `CpuUsage`. The current implementation constructs both paths under the CPU hierarchy.

yyqdbngt:
## Summary

- resolve `cpuacct.stat` and `cpuacct.usage` through the cgroup v1 cpuacct controller
- leave scheduler and quota files on the CPU controller
- cover a split CPU/cpuacct hierarchy with a regression test

## Root cause

The v1 CPU usage reader constructed both accounting paths with `SubsystemCPU`. That only works when the controllers are co-mounted. On a valid split hierarchy, the files exist under `SubsystemCPUAcct` and collection failed with file-not-found errors.

Closes #603.

## Validation

- `GOOS=linux GOARCH=amd64 go test -c ./internal/cgroups/v1`
- `GOOS=linux GOARCH=amd64 go vet ./internal/cgroups/v1`
- `gofmt -d internal/cgroups/v1/cgroups.go internal/cgroups/v1/cgroups_test.go`
- `git diff --check`

The generated Linux test binary was compiled successfully; it cannot execute on the Windows host, so runtime tests are delegated to Linux CI.

### DIFF
--- a/internal/cgroups/v1/cgroups.go
+++ b/internal/cgroups/v1/cgroups.go
@@ -90,13 +90,13 @@ func (c *CgroupV1) Procs(path string) ([]int32, error) {
 }
 
 func (c *CgroupV1) CpuUsage(path string) (*stats.CpuUsage, error) {
-	statPath := paths.Path(subsystem.SubsystemCPU, path, "cpuacct.stat")
+	statPath := paths.Path(subsystem.SubsystemCPUAcct, path, "cpuacct.stat")
 	raw, err := parseutil.RawKV(statPath)
 	if err != nil {
 		return nil, err
 	}
 
-	usagePath := paths.Path(subsystem.SubsystemCPU, path, "cpuacct.usage")
+	usagePath := paths.Path(subsystem.SubsystemCPUAcct, path, "cpuacct.usage")
 	usage, err := parseutil.ReadUint(usagePath)
 	if err != nil {
 		return nil, err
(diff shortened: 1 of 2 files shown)


huhong-web (commented):
Reviewed immutable `de8dadafdfb052dbdccc789003c2564db866c85c...e2376dadc058ea0b81c2c7473b54d8bb745f80ed` (merge base `de8dadafdfb052dbdccc789003c2564db866c85c`).

No actionable findings. CPU accounting files now resolve through the `cpuacct` subsystem while CPU control files remain under `cpu`, matching split cgroup-v1 hierarchies and preserving combined-hierarchy behavior. Validation: `git diff --check` and `gofmt` passed; GitHub Build, Lint, Vendor, and the full Ubuntu VM matrix are green. Residual risk is hybrid/nonstandard mount layouts beyond the added split-hierarchy coverage.

---
*Co-authored by: gpt-5.6-sol*

PR #606, opened 2026-08-14
fix(storage): reject nil backend config

### LINKED ISSUE #605: Storage backend construction panics on a nil configuration
### What happened?

`driver.NewBackend` dereferences `cfg.Driver` before validating the configuration pointer. Passing a nil backend configuration therefore panics, even though the storage construction API returns an error and callers expect invalid configuration to fail normally.

### Expected behavior

Return a clear configuration error and a nil backend when the backend config is nil, consistent with the existing empty-driver and unregistered-driver validation.

### Reproduction

Call `driver.NewBackend(nil)`. The current implementation panics before it can return an error.

yyqdbngt:
## Summary

- reject a nil backend configuration before reading its driver
- return a nil backend and a clear configuration error
- cover the nil construction boundary with a regression test

## Root cause

`NewBackend` accessed `cfg.Driver` before validating the configuration pointer. A missing backend configuration therefore panicked even though the construction API is designed to return validation errors.

Closes #605.

## Validation

- `go test ./internal/storage/driver -count=1`
- `go vet ./internal/storage/driver`
- `gofumpt -extra` and `goimports` on changed files
- `golangci-lint run --new-from-rev=upstream/main`
- `git diff --check`

### DIFF
--- a/internal/storage/driver/registry.go
+++ b/internal/storage/driver/registry.go
@@ -36,6 +36,9 @@ func RegisterBackend(name string, factory BackendFactory) {
 
 // NewBackend creates a backend from Config.
 func NewBackend(cfg *Config) (Backend, error) {
+	if cfg == nil {
+		return nil, fmt.Errorf("storage: backend config is nil")
+	}
 	if cfg.Driver == "" {
 		return nil, fmt.Errorf("storage: backend driver is empty")
 	}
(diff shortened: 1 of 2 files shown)


huhong-web (commented):
Reviewed immutable `de8dadafdfb052dbdccc789003c2564db866c85c...6acb97ef5e2130323ba70d0eaa32c93c741eef80` (merge base `de8dadafdfb052dbdccc789003c2564db866c85c`).

No actionable findings. The constructor now rejects a nil storage configuration before dereferencing it, and the focused regression test covers that contract. Validation: `git diff --check` and `gofmt` passed; GitHub Build, Lint, Vendor, and the full Ubuntu VM matrix are green. Residual risk is limited to callers that may rely on exact error text; no such contract was found in the repository.

---
*Co-authored by: gpt-5.6-sol*

PR #609, opened 2026-08-14
fix(job): reject nil agent task requests

yyqdbngt:
## Summary

- return a defined job error when `StartTaskContext` receives a nil request
- avoid dereferencing the request before validation
- cover the nil-request path with a regression test

## Testing

- `go test ./internal/job -run TestHTTPNodeAgentStartTask -count=1`

The full `internal/job` package run also reached an existing Windows SQLite file-lock failure in an integration test; the focused regression passes.

### DIFF
--- a/internal/job/errors.go
+++ b/internal/job/errors.go
@@ -27,5 +27,6 @@ var (
 	ErrUnsupportedJobType     = errors.New("unsupported job type")
 	ErrPersistence            = errors.New("job persistence failed")
 	ErrInvalidQuery           = errors.New("invalid job query")
+	ErrInvalidAgentRequest    = errors.New("invalid agent task request")
 	ErrAgentDispatchUncertain = errors.New("agent dispatch result is uncertain")
 )
--- a/internal/job/node_agent.go
+++ b/internal/job/node_agent.go
@@ -116,6 +116,9 @@ func (c *HTTPNodeAgent) StartTaskContext(
 ) (taskID string, returnedErr error) {
 	startedAt := time.Now()
 	defer func() { c.observeRequest("start", startedAt, returnedErr) }()
+	if args == nil {
+		return "", fmt.Errorf("%w: request must not be nil", ErrInvalidAgentRequest)
+	}
 	taskArgs := *args
 	taskArgs.ContainerID = container
 	if taskArgs.TracerName == "profiler" && !hasNonEmptyFlag(taskArgs.TracerArgs, "--tracer-id") {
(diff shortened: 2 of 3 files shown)


huhong-web (commented):
Reviewed immutable `de8dadafdfb052dbdccc789003c2564db866c85c...c4ef7a162fec3368107b022afe954c5b7d09b7fe` (merge base `de8dadafdfb052dbdccc789003c2564db866c85c`).

No actionable findings. The request is rejected before dereference, the sentinel is propagated through the agent task path, and observer error handling remains nil-safe. Validation: `git diff --check` and `gofmt` passed; GitHub Build, Lint, and Vendor are green, as are Ubuntu 22.04/24.04/26.04. Ubuntu 20.04 failed in the existing metrics-filter integration on a missing `memory_vmstat_pgfault` series, outside this RPC validation path. Residual risk is limited to end-to-end RPC behavior not reproduced locally due workspace-capacity limits.

---
*Co-authored by: gpt-5.6-sol*

yyqdbngt:
Closing this as superseded rather than rebasing it.

`internal/job/node_agent.go` no longer exists on `main`: `ccec9dda` ("feat(observation): implement job operation lifecycle") removed the whole `HTTPNodeAgent` file, so `StartTaskContext` — the function this PR guarded — is gone. A rebase cannot apply (the file is deleted upstream, `git rebase` reports `DU internal/job/node_agent.go`).

The defect itself no longer exists either. The replacement path builds the request internally from a `Job` the runtime owns, so there is no caller-supplied request pointer left to be nil:

```go
// internal/job/runtime_node_operation.go
func (r *runtime) startOperation(ctx context.Context) (*nodeapi.Operation, error) {
	job, err := r.saveOperationStartDeadline(ctx)   // non-nil or error
	...
	request, err := buildStartOperationRequest(job)
```

and the transport layer now rejects a nil request explicitly, with a test on `main`:

```go
// client/node_operations.go
func (c *NodeClient) StartOperation(ctx context.Context, address NodeAddress, request *nodeapi.StartOperationRequest) ...
	if request == nil {
		return nil, &NodeError{Code: NodeErrorCodeInvalidArgument, Message: "operation request is required"}
	}
```

`client/node_client_test.go:65 TestStartOperationRejectsNilRequest` covers that case, and `internal/job/runtime_node_operation_test.go` exercises the start path.

Thanks to whoever reviewed it earlier — the guard itself was the right idea, it just lives in the new client layer now.

PR #613, opened 2026-08-16
fix(ci): skip web sync for fork pull requests

lizhenhao123:
## Summary

- skip the web sync job for pull requests from forks
- avoid checkout failures when repository secrets are unavailable

## Independence

PR 0 of 6. This change is completely independent and can merge immediately.
The remaining profiling PRs do not depend on it.

## Validation

- workflow YAML syntax parsed successfully
- branch contains one commit and one workflow file

Split from #463.

### DIFF
--- a/.github/workflows/sync-web-on-change.yml
+++ b/.github/workflows/sync-web-on-change.yml
@@ -19,6 +19,9 @@ on:
 
 jobs:
   sync:
+    if: >-
+      github.event_name != 'pull_request' ||
+      github.event.pull_request.head.repo.full_name == github.repository
     runs-on: ubuntu-latest
     steps:
       - name: checkout huatuo


huhong-web (commented):
Reviewed immutable head `67c8186d9249` against base `de8dadafdfb0` (merge base `de8dadafdfb0`).

No actionable findings found. Validation included immutable workflow-diff inspection, GitHub event/condition tracing, YAML parsing, and `git diff --check`. Residual risk: the repository-side workflow condition still requires execution under real fork and same-repository pull-request events.

---
*Co-authored by: gpt-5.6-sol*

huhong-web (commented):
Reviewed immutable head `67c8186d9249` against merge base `de8dadafdfb0`; the current `main` workflow remains unchanged from that base. No actionable findings found.

## Architecture & Design

**Verdict: reasonable.**

- **Boundary and dependency flow:** `jobs.sync.if` is the correct admission boundary: a fork pull request is rejected before a runner is allocated and before the external checkout, build, or three deployment steps can consume repository secrets. Same-repository pull requests retain the existing behavior, while `push` and `workflow_dispatch` pass through the first clause. [GitHub documents](https://docs.github.com/en/actions/reference/workflows-and-actions/contexts) that missing context properties evaluate to an empty string, so the absent `pull_request` object on those events does not make the condition invalid.
- **API, state, and compatibility:** the patch changes only workflow-event admission. It introduces no API/data-model state, mutable global state, lifecycle ownership, or cross-layer contract. Placing one guard on the whole job avoids a partial fix where a later secret-dependent step could still run.
- **Trade-off:** removing the `pull_request` trigger entirely would be simpler if same-repository PR deployments are not needed, but it would change the stated contract. Preserving that path while excluding forks is a focused change.
- **Verification architecture:** this behavior depends on GitHub's event runtime, so an actual workflow run is stronger evidence than a repository unit test. On the original branch, [run 30748926462](https://github.com/ccfos/huatuo/actions/runs/30748926462) failed at `checkout huatuo web` before the guard; the identical guard in commit `ed4b61eda2a2` made [run 30749034391](https://github.com/ccfos/huatuo/actions/runs/30749034391) skip the entire `sync` job. The remaining residual risk is limited to the unexercised same-repository and manual-dispatch paths for this exact head.

## Code Quality & Security

- YAML parsing succeeds and the folded `if` value matches the intended expression.
- `actionlint v1.7.12` reports no finding on the added expression. Its only finding is the existing `peaceiris/actions-gh-pages@v3` runner-version warning, reproduced unchanged on the base file; it is not introduced by this PR.
- All six reported PR checks pass. No new files or Go/BPF/Shell code are present, so copyright, Go style, storage, concurrency, and language-specific checks are not applicable.
- The added comparison uses GitHub-owned repository identity fields and does not interpolate attacker-controlled data into shell commands or broaden fork access to secrets.

---
*Co-authored by: gpt-5.6-sol*

PR #614, opened 2026-08-16
feat(profiler): persist multidimensional profile labels

lizhenhao123:
## Summary

- define managed profiling dimensions for scope, CPU, PID, TGID, and container ID
- derive labels once at the common profile upload boundary
- rebuild managed labels from collector context while preserving custom labels
- cover host, CPU, process, thread-group, and container collection paths

## Standalone behavior

PR 1 of 6. This PR is independent and safe to merge by itself. Once merged,
new profile documents persist multidimensional labels while all existing query
and raw-profile APIs continue to work unchanged. It intentionally adds no new
dashboard or query endpoint.

## Dependencies

None. PR 1 and PR 2 can be reviewed, tested, and merged in either order.

## Validation

- go test ./internal/profiler -count=1
- Linux Go 1.24: go test ./internal/profiler/aggregator -count=1
- git diff --check

Split from #463.

### DIFF
--- a/internal/profiler/aggregator/save.go
+++ b/internal/profiler/aggregator/save.go
@@ -17,6 +17,9 @@ package aggregator
 import (
 	"context"
 	"fmt"
+	"sort"
+	"strconv"
+	"strings"
 	"time"
 
 	"huatuo-bamai/core/autotracing"
@@ -37,6 +40,7 @@ func (p *Pipeline) saveProfilingDocument(_ context.Context, data any) error {
 	if !ok {
 		return fmt.Errorf("invalid pprof data for uploading: %T", data)
 	}
+	setProfileCollectionLabels(flameData, p.pctx)
 
 	tracerData := &profctx.TracerData{
 		MetricData: newMetrics(int(p.overflowCount.Load())),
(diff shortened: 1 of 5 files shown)


internal/profiler/aggregator/save.go
```diff
@@ -61,3 +65,70 @@ func (p *Pipeline) saveProfilingDocument(_ context.Context, data any) error {
...
+		labels[profiler.LabelContainerID] = pctx.ContainerID
+	}
+	return labels
+}
+
+func profileCollectionScope(pctx *profctx.ProfilerContext) string {
+	switch {
+	case pctx.ContainerID != "":
+		return "container"
+	case pctx.ThreadGroup:
```

huhong-web (commented):
Reviewed immutable head `fb16ebc57756` against base `de8dadafdfb0` (merge base `de8dadafdfb0`).

Found one actionable profile-attribution issue; see the inline comment. Validation included cross-layer tracing from CLI validation through profiler context and BPF filtering to persisted labels, plus `git diff --check`. Package tests and live profiler validation could not be rerun because the review environment lacked enough disk space to materialize the immutable worktree.

---
*Co-authored by: gpt-5.6-sol*

internal/profiler/aggregator/save.go
lizhenhao123: > [P2] Do not label an unfiltered collection as `thread_group`. CLI validation accepts native `--thread-group` without `--pid`, leaving `PIDs` empty, while the BPF filter treats `profiler_filter_pid == 0` as “trace everything.” This branch therefore emits `profiling_scope=thread_group` with no `tgid` for a host-wide profile, breaking dashboard filtering and attribution. Either require at least one PID when thread-group mode is set, or select `thread_group` only when `PIDs` is non-empty, and cover the no-PID case.

fixed

huhong-web (commented):
Re-reviewed immutable head `90f7ac867894` against merge base `de8dadafdfb0`; the prior review targeted `fb16ebc57756`.

## Architecture & Design

**Verdict: reasonable.** Collection dimensions are derived at the common upload boundary, owned by the profiler package, and persisted without coupling the collector to storage or dashboard code. The follow-up commit now selects `thread_group` only when a PID/TGID target exists and covers both host and CPU fallback states, preserving the collection invariant without adding another abstraction.

All Build/Lint and VM checks pass at the new head. The added unit cases exercise the original counterexample, and no new lifecycle or security defect was found.

**Difference from ISSUE-333:** the previous profile-attribution finding is fixed; actionable findings changed from 1 to 0.

---
*Co-authored by: gpt-5.6-sol*

PR #615, opened 2026-08-16
fix(storage): add stable profile cursor pagination

lizhenhao123:
## Summary

- sort profile pages by upload time, tracer keyword, and a unique storage ID
- carry Elasticsearch search_after cursors through the storage driver
- add real Elasticsearch coverage for identical tracer and timestamp values
- provide a guarded, idempotent migration for historical profile documents
- reject multi-index aliases and repair missing keyword sort values

## Standalone behavior

PR 2 of 6. This PR is independent and safe to merge by itself. Existing raw
and merged profile queries gain stable cursor pagination; no new display API or
dashboard is introduced.

## Historical data

New documents always receive profile_storage_id. Historical documents should
be migrated with build/migrate-profile-storage-id.sh before relying on strict
long-window cursors. Without migration, the extreme legacy case of the same
tracer and the same nanosecond can still require migration or PIT.

## Dependencies

None. PR 1 and PR 2 can be reviewed, tested, and merged in either order.

## Validation

- go test ./internal/storage/... ./internal/profiler/service ./internal/job -count=1
- Elasticsearch 8.15.5 integration test with cross-page duplicate timestamps
- migration integration coverage for mappings, aliases, idempotence, and non-profile data
- shell syntax, repository shfmt flags, and git diff checks

Split from #463.

PR #616, opened 2026-08-16
feat(profiling): stream Pyroscope display queries

lizhenhao123:
## Summary

- stream profile documents page by page for merge, series, and pprof export
- remove the fixed 10k document failure for normal long-window queries
- expose Pyroscope-compatible SelectSeries and profile label aliases
- normalize text-backed aggregation fields to keyword fields
- make the optional total document limit configurable, with zero meaning unlimited
- tolerate count and pagination drift without turning retention races into 500s

## Functional boundary

PR 3 of 6. After its dependencies merge, this PR is complete and usable on its
own: SelectMergeStacktraces, SelectSeries, LabelNames/Values, raw profiles, and
portable pprof export work without PR 4 or PR 5. Diff APIs and Grafana
dashboards are intentionally left to the next layers.

## Dependencies

Depends on #614 and #615. Merge both before this PR.

This source branch includes the PR 1 and PR 2 commits so the branch can compile
and pass CI as a complete unit. After #614 and #615 merge, GitHub will remove
those already-merged changes from this PR diff automatically.

## Validation

- Linux Go 1.24 tests for profiler, aggregator, service, storage, job, config, and handlers
- long-window streaming and query-limit service tests
- real Elasticsearch keyword aggregation and stable cursor coverage
- git diff checks

Split from #463.

### DIFF
--- a/internal/profiler/aggregator/save.go
+++ b/internal/profiler/aggregator/save.go
@@ -17,6 +17,9 @@ package aggregator
 import (
 	"context"
 	"fmt"
+	"sort"
+	"strconv"
+	"strings"
 	"time"
 
 	"huatuo-bamai/core/autotracing"
@@ -37,6 +40,7 @@ func (p *Pipeline) saveProfilingDocument(_ context.Context, data any) error {
 	if !ok {
 		return fmt.Errorf("invalid pprof data for uploading: %T", data)
 	}
+	setProfileCollectionLabels(flameData, p.pctx)
 
 	tracerData := &profctx.TracerData{
 		MetricData: newMetrics(int(p.overflowCount.Load())),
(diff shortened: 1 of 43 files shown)


internal/profiler/aggregator/save.go
```diff
@@ -61,3 +65,70 @@ func (p *Pipeline) saveProfilingDocument(_ context.Context, data any) error {
...
+		labels[profiler.LabelContainerID] = pctx.ContainerID
+	}
+	return labels
+}
+
+func profileCollectionScope(pctx *profctx.ProfilerContext) string {
+	switch {
+	case pctx.ContainerID != "":
+		return "container"
+	case pctx.ThreadGroup:
```

internal/profiler/service/display_query.go
```diff
@@ -0,0 +1,558 @@
...
+
+func (s *Service) visitProfileDocuments(
+	ctx context.Context,
+	selection *profileSelection,
+	visit func(*ProfileDocument) error,
+) (int64, error) {
+	if s == nil || s.profileStorage == nil {
+		return 0, errorsProfileStorageNotInitialized()
+	}
+	if s.maxQueryDocuments > 0 {
```
huhong-web, line 124: [P1] Enforce a finite server-side document ceiling by default. With the default `MaxQueryDocuments=0`, this guard is skipped and the loop consumes every matching profile. The exact-target requirement does not bound the time range; `selectProfileTree` retains a merged tree and `SelectSeries` retains all groups/buckets until traversal ends, so a valid query over a busy host can exhaust Elasticsearch I/O, CPU, and memory. A page size only bounds each fetch. Use a positive safe default or hard cap and stop paging at it; keep node/response limits as separate output bounds.

huhong-web (commented):
Reviewed immutable head `9d35a989d90d` against base `de8dadafdfb0` (merge base `de8dadafdfb0`).

Found two actionable issues affecting profile attribution and query resource bounds; see the inline comments. Validation included cross-layer tracing through CLI/BPF filters, stable storage pagination, display endpoint handlers, profile-tree and series aggregation, and `git diff --check`. Package tests and live Elasticsearch validation could not be rerun because the review environment lacked enough disk space to materialize the immutable worktree.

---
*Co-authored by: gpt-5.6-sol*

huhong-web (commented):
Re-reviewed immutable head `9d35a989d90d` against merge base `de8dadafdfb0`.

## Architecture & Design

**Verdict: needs changes.** The service owns query policy, the storage adapter owns cursor pagination/mapping, and handlers depend on a focused consumer interface. Stable cursor paging and streaming avoid the former fixed 10k materialization boundary, but a page size is only a transfer bound: with `MaxQueryDocuments=0`, merge, pprof, and series operations can still retain unbounded trees/groups while scanning every matching document. The host-wide scope label also remains incorrect.

Use a positive safe default or hard ceiling at the service boundary and keep output-node limits separate. Build/Lint is green; Ubuntu 20.04/22.04 and web sync are failing, and no live backend long-window bound is demonstrated. No new security issue was found.

---
*Co-authored by: gpt-5.6-sol*

lizhenhao123:
> Re-reviewed immutable head `9d35a989d90d` against merge base `de8dadafdfb0`.
> 
> ## Architecture & Design
> **Verdict: needs changes.** The service owns query policy, the storage adapter owns cursor pagination/mapping, and handlers depend on a focused consumer interface. Stable cursor paging and streaming avoid the former fixed 10k materialization boundary, but a page size is only a transfer bound: with `MaxQueryDocuments=0`, merge, pprof, and series operations can still retain unbounded trees/groups while scanning every matching document. The host-wide scope label also remains incorrect.
> 
> Use a positive safe default or hard ceiling at the service boundary and keep output-node limits separate. Build/Lint is green; Ubuntu 20.04/22.04 and web sync are failing, and no live backend long-window bound is demonstrated. No new security issue was found.
> 
> **Difference from ISSUE-333:** unchanged, 2 findings remain.
> 
> _Co-authored by: gpt-5.6-sol_

fixed，please check again

PR #617, opened 2026-08-16
feat(profiling): add bounded profile diff APIs

lizhenhao123:
## Summary

- add the standard Pyroscope Diff RPC
- add a bounded JSON diff-rows adapter for dashboard consumers
- reuse the service node limits in the handler
- replace linear parent lookup with binary search

## Functional boundary

This is PR 4/6. After #616 is merged, this PR is complete and usable
without the Grafana dashboard PR. API clients can call both Diff and
diff-rows directly.

It intentionally contains no dashboard assets.

## Dependencies

Depends on #616. Merge #616 first.

The source branch contains the prerequisite PR commits so CI tests the
complete branch. Once #616 is merged, GitHub will reduce this PR to its
own diff.

## Validation

- Linux Go 1.24 service and profiling handler tests
- malformed input, node limit, and interval-boundary tests
- binary lookup benchmark: about 7.6 ns/op versus 18.9 us/op linear,
  both with zero allocations

Split from #463.

### DIFF
--- a/cmd/huatuo-apiserver/handlers/profiling/diff_rows.go
+++ b/cmd/huatuo-apiserver/handlers/profiling/diff_rows.go
@@ -19,9 +19,11 @@ import (
 	"fmt"
 	"math"
 	"net/http"
+	"sort"
 	"strings"
 
 	"huatuo-bamai/internal/log"
+	profileService "huatuo-bamai/internal/profiler/service"
 	"huatuo-bamai/internal/server"
 	"huatuo-bamai/internal/server/response"
 
@@ -31,8 +33,6 @@ import (
 
 const (
 	diffFlamegraphNodeWidth     = 7
-	defaultProfileDiffMaxNodes  = 5000
-	maxProfileDiffMaxNodes      = 10000
 	maxProfileDiffTargetLength  = 512
 	maxProfileDiffTypeLength    = 256
 	maxProfileDiffResponseBytes = 8 << 20
@@ -161,10 +161,10 @@ func buildAdjacentProfileDiffRequest(
 		return nil, fmt.Errorf("selected range has no preceding window")
 	}
 	if request.MaxNodes < 0 ||
-		request.MaxNodes > maxProfileDiffMaxNodes {
+		request.MaxNodes > profileService.MaxProfileNodes {
 		return nil, fmt.Errorf(
 			"max_nodes must be between 0 and %d",
-			maxProfileDiffMaxNodes,
+			profileService.MaxProfileNodes,
 		)
 	}
 
@@ -217,7 +217,7 @@ func buildAdjacentProfileDiffRequest(
 
 	maxNodesValue := request.MaxNodes
 	if maxNodesValue == 0 {
-		maxNodesValue = defaultProfileDiffMaxNodes
+		maxNodesValue = profileService.DefaultProfileMaxNodes
 	}
 	maxNodes := &maxNodesValue
 	return &querierv1.DiffRequest{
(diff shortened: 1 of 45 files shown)


internal/profiler/aggregator/save.go
```diff
@@ -61,3 +65,70 @@ func (p *Pipeline) saveProfilingDocument(_ context.Context, data any) error {
...
+		labels[profiler.LabelContainerID] = pctx.ContainerID
+	}
+	return labels
+}
+
+func profileCollectionScope(pctx *profctx.ProfilerContext) string {
+	switch {
+	case pctx.ContainerID != "":
+		return "container"
+	case pctx.ThreadGroup:
```

internal/profiler/service/display_query.go
```diff
@@ -0,0 +1,606 @@
...
+
+func (s *Service) visitProfileDocuments(
+	ctx context.Context,
+	selection *profileSelection,
+	visit func(*ProfileDocument) error,
+) (int64, error) {
+	if s == nil || s.profileStorage == nil {
+		return 0, errorsProfileStorageNotInitialized()
+	}
+	if s.maxQueryDocuments > 0 {
```
huhong-web, line 124: [P1] Enforce a finite server-side document ceiling by default. With the default `MaxQueryDocuments=0`, this guard is skipped and the loop consumes every matching profile. The exact-target requirement does not bound the time range; `selectProfileTree` retains a merged tree and `SelectSeries` retains all groups/buckets until traversal ends, so a valid query over a busy host can exhaust Elasticsearch I/O, CPU, and memory. A page size only bounds each fetch. Use a positive safe default or hard cap and stop paging at it; keep node/response limits as separate output bounds.

huhong-web (commented):
Reviewed immutable head `3398ddc30558` against base `de8dadafdfb0` (merge base `de8dadafdfb0`).

Found two actionable issues affecting profile attribution and query resource bounds; see the inline comments. Validation included cross-layer tracing through CLI/BPF filters, stable storage pagination, diff tree/row construction and bounded response logic, and `git diff --check`. Package tests and live Elasticsearch validation could not be rerun because the review environment lacked enough disk space to materialize the immutable worktree.

---
*Co-authored by: gpt-5.6-sol*

cmd/huatuo-apiserver/handlers/profiling/diff_rows.go
```diff
@@ -0,0 +1,428 @@
...
+	if maxNodesValue == 0 {
+		maxNodesValue = profileService.DefaultProfileMaxNodes
+	}
+	maxNodes := &maxNodesValue
+	return &querierv1.DiffRequest{
+		Left: &querierv1.SelectMergeStacktracesRequest{
+			ProfileTypeID: request.ProfileTypeID,
+			LabelSelector: selector,
+			Start:         request.Start - window,
+			End:           request.Start - 1,
```
huhong-web: [P2] Preserve equal, truly adjacent comparison windows. `buildProfileAggregationQuery` applies inclusive `>=`/`<=` filters to `uploaded_time`, and profile documents store `time.Now()` with sub-millisecond precision. For `start=2000,end=3000`, this creates left `[1000,1999]` and right `[2000,3000]`: the left interval is 1 ms shorter and records between 1999.000000001 and 2000 are in neither side. Model both windows with the same half-open boundary (or carry an exclusive nanosecond end into the storage filter), and add boundary fixtures with sub-millisecond timestamps so the windows neither overlap nor leave a gap.

huhong-web (commented):
Re-reviewed immutable head `3398ddc30558` against merge base `de8dadafdfb0`.

## Architecture & Design

**Verdict: needs changes.** Reusing the service tree construction for `Diff` and bounding both graph nodes and the JSON adapter are sound choices; the handler interface remains consumer-owned and focused. The inherited unlimited document default and host-wide attribution defects remain, and the new adjacent-window adapter does not actually create equal, contiguous windows at the full precision used by storage; see the inline comment.

The smallest fix is to define one half-open time-range contract and use it consistently through request conversion and `uploaded_time` filtering. Build/Lint is green, while Ubuntu 20.04 and web sync are failing. No additional security issue was found.

**Difference from ISSUE-333:** the 2 prior findings remain and 1 new time-contract finding was added, for 3 total.

---
*Co-authored by: gpt-5.6-sol*

cmd/huatuo-apiserver/handlers/profiling/diff_rows.go
lizhenhao123: fixed，please check again

PR #618, opened 2026-08-16
feat(grafana): add multidimensional profiling dashboards

lizhenhao123:
## Summary

- add host, container, and adjacent-window comparison dashboards
- add time series, fixed Top 10 by tracer, flame graph, and Top table views
- encode dashboard filters for the bodyless POST comparison adapter
- link profiling jobs to the display dashboards
- document and illustrate the complete workflow

Top 10 is intentionally fixed to `tracer`: Grafana's Pyroscope data source
does not interpolate variables inside the `groupBy` array.

## Functional boundary

This is PR 6/6 and provides the final UI. Its dashboards are functional after
the query and diff APIs below have merged.

## Dependencies

Depends on #616 and #617. Merge them in that order after #614 and #615.

The source branch contains all prerequisite commits so CI exercises the full
feature. GitHub will reduce this PR to the dashboard integration diff as its
dependencies merge.

### Host profile view

Host profile dashboard

### Container profile view

Container profile dashboard

### Adjacent-window comparison

Profile comparison dashboard

## Validation

- Linux Go 1.24 targeted suite for profiler, storage, job, config, handlers,
  and Grafana dashboard contracts
- JSON syntax and dashboard contract tests
- bodyless POST with unknown content length and percent-decoded special values
- verified real PNG screenshots at 1280 px width

Split from #463.


internal/profiler/aggregator/save.go
```diff
@@ -61,3 +65,70 @@ func (p *Pipeline) saveProfilingDocument(_ context.Context, data any) error {
...
+		labels[profiler.LabelContainerID] = pctx.ContainerID
+	}
+	return labels
+}
+
+func profileCollectionScope(pctx *profctx.ProfilerContext) string {
+	switch {
+	case pctx.ContainerID != "":
+		return "container"
+	case pctx.ThreadGroup:
```

internal/profiler/service/display_query.go
```diff
@@ -0,0 +1,608 @@
...
+
+func (s *Service) visitProfileDocuments(
+	ctx context.Context,
+	selection *profileSelection,
+	visit func(*ProfileDocument) error,
+) (int64, error) {
+	if s == nil || s.profileStorage == nil {
+		return 0, errorsProfileStorageNotInitialized()
+	}
+	if s.maxQueryDocuments > 0 {
```
huhong-web, line 126: [P1] Enforce a finite server-side document ceiling by default. With the default `MaxQueryDocuments=0`, this guard is skipped and the loop consumes every matching profile. The exact-target requirement does not bound the time range; `selectProfileTree` retains a merged tree and `SelectSeries` retains all groups/buckets until traversal ends, so a valid query over a busy host can exhaust Elasticsearch I/O, CPU, and memory. A page size only bounds each fetch. Use a positive safe default or hard cap and stop paging at it; keep node/response limits as separate output bounds.

docs/key-feature/continuous-profiling_en.md
```diff
@@ -340,7 +352,46 @@ Profiling jobs use these statuses:
...
+completed job's `results.url` opens the corresponding dashboard with its
+target, profile type, and collection window preselected.
+
+The Pyroscope datasource sends queries to huatuo-apiserver. If API
+authentication is enabled, configure a dedicated user with
+`POST /v1/profiles/flamegraph/**` permission and pass the same user ID to the
+Grafana container without committing it:
+
+```bash
+export HUATUO_GRAFANA_PROFILE_TOKEN="<Auth.users.ID>"
```
huhong-web: [P1] These authentication instructions cannot work. `NewAuthService` indexes users by `Auth.users.BearerToken`, not by ID, so exporting the ID produces 401. In addition, `/v1/profiles/flamegraph/**` is in `AdminPaths`, so the recommended non-admin dedicated user with only this permission produces 403 even with the correct token; the later pprof example also sends `USER_ID`. Document a bearer token belonging to an admin, or remove the admin-only gate and add an authenticated datasource test for a scoped user.

huhong-web (commented):
Reviewed immutable head `6168462b069d` against base `de8dadafdfb0` (merge base `de8dadafdfb0`).

Found three actionable issues affecting profile attribution, query resource bounds, and Grafana authentication; see the inline comments. Validation included cross-layer tracing through CLI/BPF filters, storage pagination/aggregation, server authentication, dashboard provisioning, JSON/YAML parsing, migration-script `bash -n`, and `git diff --check`. Package tests and live Elasticsearch/Grafana validation could not be rerun because the review environment lacked enough disk space to materialize the immutable worktree.

---
*Co-authored by: gpt-5.6-sol*

huhong-web (commented):
Re-reviewed immutable head `6168462b069d` against merge base `de8dadafdfb0`.

## Architecture & Design

**Verdict: needs changes.** The end-to-end boundary is otherwise coherent—collection labels are written once, indexed through the storage mapper, queried by the service, and consumed by provisioned dashboards; node and JSON response bounds are explicit. However, the server still defaults the total document budget to unlimited, the documented Grafana credential does not match the authentication/admin contract, and the host-wide scope attribution issue remains. The comparison adapter also inherits an unequal/gapped time-window contract and dispatches request binding from `Content-Length`, which breaks unknown-length JSON bodies; see the new inline comments.

A simpler safe contract is one finite server-owned document ceiling, half-open equal comparison windows, and content-type-based request decoding. Build/Lint is green, but Ubuntu 20.04 and the web sync check are failing; live Elasticsearch/Grafana behavior is not proven by those checks. No separate injection, path, crypto, or hard-coded-secret issue was found.

**Difference from ISSUE-333:** the 3 prior findings remain and 2 new API/design findings were added, for 5 total.

---
*Co-authored by: gpt-5.6-sol*

cmd/huatuo-apiserver/handlers/profiling/diff_rows.go
```diff
@@ -0,0 +1,442 @@
...
+	if maxNodesValue == 0 {
+		maxNodesValue = profileService.DefaultProfileMaxNodes
+	}
+	maxNodes := &maxNodesValue
+	return &querierv1.DiffRequest{
+		Left: &querierv1.SelectMergeStacktracesRequest{
+			ProfileTypeID: request.ProfileTypeID,
+			LabelSelector: selector,
+			Start:         request.Start - window,
+			End:           request.Start - 1,
```
huhong-web: [P2] Preserve equal, truly adjacent comparison windows. `buildProfileAggregationQuery` applies inclusive `>=`/`<=` filters to `uploaded_time`, and profile documents store `time.Now()` with sub-millisecond precision. For `start=2000,end=3000`, this creates left `[1000,1999]` and right `[2000,3000]`: the left interval is 1 ms shorter and records between 1999.000000001 and 2000 are in neither side. Model both windows with the same half-open boundary (or carry an exclusive nanosecond end into the storage filter), and add boundary fixtures with sub-millisecond timestamps so the windows neither overlap nor leave a gap.

cmd/huatuo-apiserver/handlers/profiling/diff_rows.go
```diff
@@ -0,0 +1,442 @@
...
+	row        profileDiffRow
+	leftStart  int64
+	rightStart int64
+	children   []*profileDiffNode
+}
+
+func (h *Handler) displayDiffRows(ctx *server.Context) error {
+	var request profileDiffRowsRequest
+	var err error
+	if ctx.Request().ContentLength <= 0 {
```
huhong-web: [P2] Do not use `ContentLength` to choose the request representation. Valid chunked HTTP/1.1 and streaming HTTP/2 JSON requests have `ContentLength == -1`; this branch binds only the query string and ignores their JSON body, turning a previously supported JSON request into a 400. Dispatch from `Content-Type` (with a query path only for a genuinely bodyless request), and add an `application/json` test with unknown length, for example an `io.Pipe` body.

lizhenhao123: fixed，please check again

PR #622, opened 2026-08-16
fix(ascend): validate hccn network output

### LINKED ISSUE #621: [Bug]: HCCN parsers accept incomplete bandwidth and unsupported link output
## Problem

`GetInterfaceTraffic` treats incomplete or malformed `hccn_tool -bandwidth -g` output as a successful sample. The documented successful output contains both `Bandwidth TX` and `Bandwidth RX` lines, but if one line is missing or its number cannot be parsed the function returns `nil` with that direction left at zero. The collector can therefore publish a plausible zero-rate metric instead of dropping an invalid scrape.

`GetLinkStatus` has the same validation gap: any three-field response is accepted, so an unexpected status is later represented as DOWN.

## Reproduction

The parser behavior can be reproduced with these command outputs:

```text
Bandwidth TX: 12.5 MB/sec
```

or:

```text
link status: DEGRADED
```

Both are incomplete/unsupported according to the command contract, but the current parsing path does not return an error.

## Expected behavior

Validate the complete output contract before returning a sample:

- require one valid TX and one valid RX value;
- accept only the documented UP/DOWN link states;
- return a descriptive error for missing, duplicate, malformed, or unsupported fields.

I can submit a focused change that extracts pure parsing helpers and covers valid and invalid output with table-driven tests.

btlqql:
## Summary

- parse the numeric field from the documented `Bandwidth TX/RX` output instead of the direction token
- require exactly one valid TX and RX sample before exporting bandwidth metrics
- reject unsupported link states and non-finite or negative bandwidth values
- add table-driven parser coverage for valid, incomplete, duplicate, malformed, and unsupported output

## Problem

For `Bandwidth TX: 12.50 MB/sec`, the previous parser attempted to parse `TX:` as the number. It skipped the resulting error and returned `(0, 0, nil)`, so valid traffic could be exported as zero. Partial or unsupported output was also accepted as a successful sample.

## Testing

- `go test ./core/metrics/ascend/hccn -count=1`
- `go vet ./core/metrics/ascend/hccn`
- `golangci-lint run ./core/metrics/ascend/hccn/...`
- `git diff --check`

Fixes #621


huhong-web (commented):
Reviewed immutable head `53207120fdc4` against base `de8dadafdfb0` (merge base `de8dadafdfb0`).

No actionable findings found. Validation included immutable diff and caller inspection of HCCN link-state/bandwidth parsing, duplicate-field rejection, and error propagation, plus `git diff --check`. Package tests could not be rerun because the review environment lacked enough disk space to materialize the immutable worktree. Residual risk: Compatibility with every deployed `hccn_tool` output version remains an integration boundary.

---
*Co-authored by: gpt-5.6-sol*

btlqql:

PR #625, opened 2026-08-16
fix(ascend): require complete PCIe link samples

### LINKED ISSUE #623: [Bug]: Incomplete PCIe sysfs reads are exported as zero-valued metrics
## Problem

`GetPCIeLinkInfo` reads four sysfs attributes (`max_link_speed`, `max_link_width`, `current_link_speed`, and `current_link_width`) but silently discards each field's read or parse error. It reports success whenever either speed is non-zero.

The caller then publishes all four metrics as one sample. If any other attribute is missing or malformed, its Go zero value is exported as real data, for example PCIe width `x0` or link speed `0 GT/s`, instead of dropping the incomplete sample.

## Reproduction

Create a sysfs-shaped fixture with valid speed files but omit `current_link_width`. The current function returns a non-nil `PCIeLinkInfo` with `StatusWidth == 0` and no error. Invalid numeric content in a width or speed file behaves the same way when another speed remains valid.

## Expected behavior

Treat the four attributes as one atomic PCIe link sample. Return a field-specific error when any required file is missing or malformed, and only let the caller publish metrics after every value has been read successfully.

I can add a fixture-backed parser seam and table tests for every missing field and malformed values.

yyqdbngt:
## Summary

- read all four PCIe capability and status attributes as one complete sample
- return field-specific errors instead of publishing missing values as zero
- reject non-finite or non-positive speeds and zero-width links
- add sysfs-shaped fixtures covering every missing field and invalid speed/width values

## Problem

The previous implementation ignored each individual sysfs error and returned success whenever either speed was non-zero. Because the caller publishes all four metrics together, a missing width or speed was exposed as a real `0` value such as PCIe `x0`.

## Testing

- `go test ./core/metrics/ascend/pcie -count=1`
- `go vet ./core/metrics/ascend/pcie`
- `golangci-lint run ./core/metrics/ascend/pcie/...`
- `git diff --check`

Fixes #623

### DIFF
--- a/core/metrics/ascend/pcie/pcie.go
+++ b/core/metrics/ascend/pcie/pcie.go
@@ -23,7 +23,9 @@ package pcie
 
 import (
 	"fmt"
+	"math"
 	"os"
+	"path/filepath"
 	"strconv"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


huhong-web (commented):
Reviewed immutable head `33e93c80e64a` against base `de8dadafdfb0` (merge base `de8dadafdfb0`).

No actionable findings found. Validation included immutable diff and caller inspection of PCIe sysfs reads, numeric/finite checks, and error propagation, plus `git diff --check`. Package tests could not be rerun because the review environment lacked enough disk space to materialize the immutable worktree. Residual risk: Actual device sysfs values and vendor/kernel variants remain hardware-dependent.

---
*Co-authored by: gpt-5.6-sol*

PR #626, opened 2026-08-16
fix(ascend): roll back failed DCMI initialization

### LINKED ISSUE #624: [Bug]: Failed DCMI initialization leaks the library reference
## Problem

`DcInit` first calls `libdcmi.load()`, which opens `libdcmi.so` and increments the library reference count. It then calls the native `dcmi_init` function. If that native initialization returns an error, `DcInit` returns immediately without releasing the reference acquired by `load`.

The failed attempt therefore leaves the shared library open with a positive reference count. A later retry takes the already-loaded path, and shutdown calls no longer correspond to successful initialization calls.

## Reproduction

With a dynamic-library stub whose open succeeds and a `dcmi_init` stub that returns a failure code:

1. call `DcInit`;
2. observe the returned initialization error;
3. observe that the loader's close operation was not called and the reference count remains one.

## Expected behavior

A failed native initialization should roll back exactly the library reference acquired by that `DcInit` call. The reference count should return to zero, the library should close when appropriate, and a later `DcInit` retry should be able to load and initialize normally. Successful initialization and shutdown behavior must remain unchanged.

I can add the rollback and lifecycle tests with a fake dynamic library and a no-op symbol registrar.

btlqql:
## Summary

- release the library reference acquired by `DcInit` when native `dcmi_init` fails
- preserve both the initialization and rollback errors if closing the library also fails
- make symbol registration injectable for lifecycle tests while retaining the production registrar
- test failed-init rollback, successful retry, balanced reference counts, and normal shutdown

## Problem

`libdcmi.load()` increments the reference count before the native initialization call. The previous error path returned without balancing that reference, leaving the library open and causing later retries and shutdowns to operate on stale lifecycle state.

## Testing

- `GOOS=linux go test -c ./core/metrics/ascend/dcmi` (Linux test binary compiles)
- `GOOS=linux go vet ./core/metrics/ascend/dcmi`
- `GOOS=linux golangci-lint run --no-config --disable-all -E govet -E staticcheck -E errcheck ./core/metrics/ascend/dcmi/...`
- `git diff --check`

The package tests cannot execute on the Windows development host because `purego.Dlopen` is Linux-only; CI will execute the compiled lifecycle tests on Linux.

Fixes #624

### DIFF
--- a/core/metrics/ascend/dcmi/init.go
+++ b/core/metrics/ascend/dcmi/init.go
@@ -15,12 +15,23 @@
 
 package dcmi
 
+import (
+	"errors"
+	"fmt"
+)
+
 // DcInit loads and initializes the Ascend DCMI library.
 func DcInit() error {
 	if err := libdcmi.load(); err != nil {
 		return err
 	}
-	return checkReturnCode("dcmi_init", dcmiInit())
+	if err := checkReturnCode("dcmi_init", dcmiInit()); err != nil {
+		if closeErr := libdcmi.close(); closeErr != nil {
+			return errors.Join(err, fmt.Errorf("roll back DCMI library load: %w", closeErr))
+		}
+		return err
+	}
+	return nil
 }
 
 // DcShutDown shuts down the Ascend DCMI library.
--- a/core/metrics/ascend/dcmi/init_test.go
+++ b/core/metrics/ascend/dcmi/init_test.go
@@ -1,0 +1,124 @@
+// Copyright 2026 The HuaTuo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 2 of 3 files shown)


huhong-web (commented):
Reviewed immutable head `6c23a1eaf53b` against base `de8dadafdfb0` (merge base `de8dadafdfb0`).

No actionable findings found. Validation included immutable diff inspection of DCMI library/symbol initialization, rollback ownership, and close-error propagation, plus `git diff --check`. Package tests could not be rerun because the review environment lacked enough disk space to materialize the immutable worktree. Residual risk: The vendor DCMI library and hardware initialization path remain hardware-dependent.

---
*Co-authored by: gpt-5.6-sol*

btlqql:

PR #628, opened 2026-08-17
fix(server): constrain recursive auth wildcards

### LINKED ISSUE #627: [Bug]: Recursive auth wildcards match unrelated path prefixes
## 这个问题是否已经存在？

## 部署环境

- Architecture: independent of CPU architecture
- Operating system: independent of Linux distribution and kernel version
- Affected component: `internal/server` authorization middleware used by `huatuo-apiserver`
- Authentication must be enabled and a recursive `**` path permission, public path, or administrator path must be configured

## 如何复现

`authService.matchesPath` handles every permission containing `**` by taking the text before the first wildcard and calling `strings.HasPrefix`:

```go
if strings.Contains(permission, "**") {
    prefix := strings.Split(permission, "**")[0]
    return strings.HasPrefix(path, prefix)
}
```

This compares bytes rather than path segments and discards everything after the wildcard.

A minimal package-level regression case is:

```go
svc := NewAuthService(nil)

got := svc.matchesPath("/v1/profiles/**", "/v1/profiles-archive/export")
// got == true; want false
```

The configured permission is intended for descendants of `/v1/profiles/`, but it also authorizes the sibling prefix `/v1/profiles-archive/`. The same matcher is used by `matchesAnyPath` for `PublicPaths` and `AdminPaths`, so the defect is not limited to per-user permissions.

A second case shows that a suffix after `**` is ignored:

```go
got := svc.matchesPath("/v1/**/result", "/v1/tasks/task-1/delete")
// got == true; want false
```

The current implementation sees only the `/v1/` prefix and never checks the required `result` segment.

### Actual behavior

- `**` can cross a path-segment boundary into an unrelated sibling whose name shares the same byte prefix.
- Any permission suffix after `**` is ignored.
- A non-administrator can therefore receive access outside the configured route set when such a pattern is present.
- A similarly shaped custom public path can bypass bearer authentication for an unintended endpoint.

### Expected behavior

- `*` and `:param` must continue to match exactly one complete path segment.
- `**` must match one or more complete path segments without crossing into a sibling prefix.
- Segments after `**` must still be matched.
- Existing documented behavior must remain intact: `/v1/profiles/**` matches descendants such as `/v1/profiles/task-1`, while the base `/v1/profiles` remains a separate permission.

## 修复建议或方案

Remove the `strings.HasPrefix` shortcut and perform segment-aware recursive matching for all patterns:

1. split the permission and request path into path segments;
2. match literals, `*`, and `:param` one segment at a time;
3. make a `**` segment consume one or more complete path segments;
4. continue matching any permission suffix after `**`;
5. add regression tests for sibling-prefix rejection, suffix enforcement, nested descendants, and the documented base-path behavior.

The matcher can use a small recursive or dynamic-programming implementation. Route paths contain few segments, but memoization avoids repeated work for patterns containing more than one `**`.

## 其他

This is authorization-boundary behavior, not only a helper correctness issue. The same function is called from:

- `authService.Validate` for user permissions;
- `matchesAnyPath(..., publicPaths, ...)` before authentication;
- `matchesAnyPath(..., adminPaths, ...)` for administrator-only endpoints.

I can submit a focused fix with package-level tests and run `go test`, `go vet`, repository formatting, and the project checks applicable to this package.

tju-yxq:
## Summary

- replace byte-prefix matching for recursive authorization wildcards with segment-aware matching
- preserve the documented distinction between a collection path and its descendants
- enforce permission suffixes that follow a recursive wildcard
- add helper-level and middleware-level regression coverage for user and public-path authorization

Fixes #627.

## Problem

`authService.matchesPath` previously handled any permission containing `**` by splitting at the first wildcard and calling `strings.HasPrefix` on the request path.

That had two authorization-boundary consequences:

1. `/v1/profiles/**` also matched `/v1/profiles-archive/export` because both strings share the `/v1/profiles` byte prefix.
2. `/v1/**/result` matched unrelated paths such as `/v1/tasks/task-1/delete` because the suffix after `**` was never evaluated.

The same matcher is used for normal user permissions and configured public or administrator paths, so this was not limited to one endpoint family.

## Implementation

The matcher now evaluates complete path segments with a memoized recursive search:

- literal segments must be equal;
- `*` and `:param` consume exactly one segment;
- `**` consumes one or more complete segments;
- matching resumes after `**`, so a suffix remains mandatory;
- the request must be fully consumed before the permission matches.

Requiring `**` to consume at least one segment intentionally preserves the existing documented behavior: a permission for `/v1/profiles/**` covers descendants, while `/v1/profiles` remains a separate permission.

The memo table is bounded by `len(permissionSegments) * len(pathSegments)`, avoiding repeated exploration when a pattern contains multiple recursive wildcards.

## Regression coverage

`TestAuthServiceMatchesPath` now covers:

- nested recursive descendants;
- the base path remaining distinct;
- rejection of sibling prefixes;
- suffix matching after `**`;
- suffix rejection for unrelated endpoints;
- the existing exact, single-wildcard, path-parameter, and segment-count behavior.

Middleware tests additionally verify the externally visible boundaries:

- a viewer with `/v1/profiles/**` cannot access `/v1/profiles-archive/export`;
- a configured public `/v1/profiles/**` descendant remains public;
- `/v1/profiles-archive/export` still requires authentication.

## Validation

- `go test ./internal/server -count=1`
- `go test -race ./internal/server -count=1`
- `go vet ./internal/server`
- `golangci-lint run -v ./internal/server --timeout=5m --config .golangci.yaml`
- `golangci-lint run -v ./... --timeout=10m --config .golangci.yaml`
- `git diff --check`

All commands above passed. The full linter used the repository's 26 configured linters and completed with no reportable issues.

`make check` was also invoked in the official development image. Its fixed five-minute linter deadline expired while `go/packages` was loading the bind-mounted Windows workspace; it reported no source finding before the infrastructure timeout. Running the same configured full-repository lint on a container-local copy completed successfully in 1 minute 29 seconds.

### DIFF
--- a/internal/server/auth.go
+++ b/internal/server/auth.go
@@ -109,18 +109,9 @@ func splitPermission(permission string) (string, string) {
 
 // matchesPath performs simple path matching, supporting basic wildcards and path parameters.
 func (s *authService) matchesPath(permission, path string) bool {
-	// 1. Exact match
 	if permission == path {
 		return true
 	}
-
-	// 2. Handle wildcard ** (matches all sub-paths)
-	if strings.Contains(permission, "**") {
-		prefix := strings.Split(permission, "**")[0]
-		return strings.HasPrefix(path, prefix)
-	}
-
-	// 3. Handle single-level wildcard * and path parameter :param
 	return s.matchesSegments(permission, path)
 }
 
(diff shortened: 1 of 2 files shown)


huhong-web (commented):
Reviewed immutable head `ff5b7f5c460c` against base `de8dadafdfb0` (merge base `de8dadafdfb0`).

No actionable findings found. Validation included immutable diff inspection of matcher semantics, memoization, middleware authorization flow, and the added cases, plus `git diff --check`. Package tests could not be rerun because the review environment lacked enough disk space to materialize the immutable worktree. Residual risk: End-to-end behavior across proxy/path normalization remains an integration boundary.

---
*Co-authored by: gpt-5.6-sol*

PR #630, opened 2026-08-17
fix(executil): restore UTS namespace after hostname lookup

### LINKED ISSUE #629: [Bug]: HostnameByPid leaves worker threads in target UTS namespaces
## 这个问题是否已经存在？

## 部署环境

- Operating system: Linux with UTS namespaces (containerized workloads are the normal trigger)
- Affected component: `internal/utils/executil.HostnameByPid`
- Call sites: iotracing event formatting and aggregation
- Required privilege: the HUATUO process must be able to call `setns(2)`, which is already required for the current helper to succeed

## 如何复现

`HostnameByPid` locks its goroutine to an operating-system thread, opens the target process UTS namespace, enters it, reads the hostname, and then unlocks the thread:

```go
runtime.LockOSThread()

if err := unix.Setns(int(fd.Fd()), unix.CLONE_NEWUTS); err != nil {
    return "", err
}
return os.Hostname()
```

The original UTS namespace is never opened or restored.

Linux namespace membership is a property of the calling task/thread. `runtime.UnlockOSThread` only allows the Go scheduler to reuse that operating-system thread; it does not reverse `setns`. Therefore, after a successful call for a container process:

1. the worker thread remains in the container's UTS namespace;
2. the scheduler can run unrelated goroutines on that thread;
3. later hostname or namespace-sensitive operations can observe the container namespace instead of HUATUO's namespace;
4. repeated calls for different containers move pooled worker threads among arbitrary target UTS namespaces.

A privileged regression test can demonstrate the leak:

1. create a child process in a new UTS namespace with a distinct hostname;
2. record `/proc/thread-self/ns/uts` on a locked calling thread;
3. enter the child namespace through the current helper logic;
4. observe that the thread's namespace inode remains the child inode after the hostname read;
5. unlock the thread and observe that the contaminated thread is eligible for reuse.

The existing `TestHostnameByPid` calls the helper only for the current process and checks the returned hostname. Entering the same namespace does not reveal the missing restoration path.

### Actual behavior

A successful hostname lookup permanently changes the UTS namespace of the operating-system thread that executed the lookup.

### Expected behavior

- The helper must snapshot the calling thread's current UTS namespace before entering the target namespace.
- It must restore that exact namespace before returning the thread to the Go scheduler.
- Hostname lookup errors must not skip restoration.
- A restoration error must be returned together with any hostname error.
- A thread whose namespace could not be restored must never be returned to the scheduler's reusable thread pool.
- All namespace file descriptors must be closed on every path.

## 修复建议或方案

Run the namespace transition in a dedicated goroutine that owns a locked operating-system thread:

1. call `runtime.LockOSThread`;
2. open `/proc/thread-self/ns/uts` to snapshot the namespace of that exact thread;
3. open the target `/proc/<pid>/ns/uts` handle;
4. enter the target namespace and read the hostname;
5. restore the saved namespace in a mandatory cleanup path;
6. unlock the thread only after restoration succeeds;
7. if restoration fails, return the error and let the locked goroutine exit without `runtime.UnlockOSThread`, which causes the Go runtime to terminate that contaminated thread instead of reusing it.

The namespace operations should be isolated behind a small internal test seam so unit tests can deterministically verify operation ordering, restoration after hostname failures, joined errors, descriptor closure, and the restoration-failure path without requiring `CAP_SYS_ADMIN`. The existing Linux integration-style test can remain for the real syscall path.

## 其他

This is not a cosmetic hostname mismatch. It violates process-wide runtime isolation by returning a namespace-mutated operating-system thread to the Go scheduler. The defect is triggered on the normal successful path whenever iotracing resolves a process in a different UTS namespace.

I can submit a focused fix with deterministic unit tests, the existing syscall test, race-enabled tests, vet, configured lint, and repository diff checks.

tju-yxq:
## Summary

- snapshot the calling operating-system thread's UTS namespace before entering a target process namespace
- restore that exact namespace after reading the target hostname, including hostname-error paths
- prevent a thread with a failed namespace restoration from returning to the Go scheduler
- close both namespace descriptors on every path and preserve cleanup errors
- add deterministic unit coverage that does not require `CAP_SYS_ADMIN`

Fixes #629.

## Problem

`HostnameByPid` previously called `runtime.LockOSThread`, entered `/proc/<pid>/ns/uts`, read the hostname, and then called `runtime.UnlockOSThread` without restoring the original namespace.

UTS namespace membership belongs to the Linux task/thread. Unlocking a Go goroutine does not undo `setns(2)`; it only makes the operating-system thread reusable. A successful iotracing hostname lookup could therefore return a thread that still belonged to a container UTS namespace to the scheduler, allowing unrelated goroutines to observe the wrong namespace.

The existing test used the current process PID, so entering the same UTS namespace could not expose this lifecycle defect.

## Implementation

`HostnameByPid` now delegates the transition to a dedicated goroutine that owns a locked operating-system thread.

The worker:

1. opens `/proc/thread-self/ns/uts`, which snapshots the namespace of the exact locked thread;
2. opens the requested `/proc/<pid>/ns/uts` handle;
3. enters the target namespace;
4. reads the hostname;
5. restores the saved namespace even when hostname lookup fails;
6. closes both namespace handles and joins any close errors;
7. unlocks the thread only when restoration succeeds.

If restoration fails, the result is returned to the caller and the worker exits without `runtime.UnlockOSThread`. Go terminates an operating-system thread when its locked goroutine exits, so a namespace-contaminated thread cannot re-enter the scheduler pool.

The production path still uses `os.Open`, `unix.Setns`, and `os.Hostname`. A small unexported operations seam makes lifecycle behavior deterministic in unit tests without changing the public API.

## Error behavior

Errors now identify the failed lifecycle stage:

- opening the current thread namespace;
- opening the target namespace;
- entering the target namespace;
- reading the target hostname;
- restoring the original namespace;
- closing either namespace descriptor.

`errors.Join` preserves simultaneous hostname, restoration, and close failures for `errors.Is` callers.

## Regression coverage

The new tests verify:

- target entry followed by restoration in the correct order;
- restoration after `os.Hostname` failure;
- restoration errors are returned while the resolved hostname remains available;
- target-entry failures do not call the hostname operation;
- both descriptors close after success and failures;
- the current descriptor closes when opening the target namespace fails;
- close errors from both descriptors remain discoverable.

The existing Linux syscall test remains in place for the real `HostnameByPid` path and still skips when the environment lacks the required namespace capability.

## Validation

- `go test ./internal/utils/executil -count=1`
- `go test -race ./internal/utils/executil -count=1`
- `go vet ./internal/utils/executil`
- `golangci-lint run -v ./internal/utils/executil --timeout=5m --config .golangci.yaml`
- `golangci-lint run -v ./... --timeout=10m --config .golangci.yaml`
- `git diff --check`

All commands above passed. The full repository lint used all 26 configured linters and completed with no reportable issues.

### DIFF
--- a/internal/utils/executil/exec.go
+++ b/internal/utils/executil/exec.go
@@ -173,21 +173,113 @@ func RunningDir() (string, error) {
 	return filepath.Dir(exePath), nil
 }
 
+const currentThreadUTSNamespace = "/proc/thread-self/ns/uts"
+
+type namespaceHandle interface {
+	Fd() uintptr
+	Close() error
+}
+
+type utsNamespaceOperations struct {
+	open     func(string) (namespaceHandle, error)
+	setns    func(int, int) error
+	hostname func() (string, error)
+}
+
+type utsHostnameResult struct {
+	hostname      string
+	err           error
+	discardThread bool
+}
+
 func HostnameByPid(pid uint32) (string, error) {
-	var empty string
-	fd, err := os.Open(procfs.Path(fmt.Sprintf("%d", pid), "ns/uts"))
+	targetPath := procfs.Path(fmt.Sprintf("%d", pid), "ns/uts")
+	return hostnameByUTSNamespace(targetPath, utsNamespaceOperations{
+		open: func(path string) (namespaceHandle, error) {
+			return os.Open(path)
+		},
+		setns:    unix.Setns,
+		hostname: os.Hostname,
+	})
+}
+
+func hostnameByUTSNamespace(
+	targetPath string,
+	operations utsNamespaceOperations,
+) (string, error) {
+	resultCh := make(chan utsHostnameResult, 1)
+	go func() {
+		runtime.LockOSThread()
+		result := readHostnameInUTSNamespace(targetPath, operations)
+		resultCh <- result
+		if !result.discardThread {
+			runtime.UnlockOSThread()
+		}
+		// Exiting while still locked makes the runtime terminate a thread
+		// whose original namespace could not be restored.
+	}()
+
+	result := <-resultCh
+	return result.hostname, result.err
+}
+
+func readHostnameInUTSNamespace(
+	targetPath string,
+	operations utsNamespaceOperations,
+) utsHostnameResult {
+	current, err := operations.open(currentThreadUTSNamespace)
 	if err != nil {
-		return empty, err
+		return utsHostnameResult{
+			err: fmt.Errorf("open current thread UTS namespace: %w", err),
+		}
 	}
-	defer fd.Close()
 
-	runtime.LockOSThread()
-	defer runtime.UnlockOSThread()
-
-	if err := unix.Setns(int(fd.Fd()), unix.CLONE_NEWUTS); err != nil {
-		return empty, err
+	target, err := operations.open(targetPath)
+	if err != nil {
+		return utsHostnameResult{
+			err: errors.Join(
+				fmt.Errorf("open target UTS namespace %q: %w", targetPath, err),
+				closeNamespaceHandle(current, "current thread UTS namespace"),
+			),
+		}
 	}
-	return os.Hostname()
+
+	if err := operations.setns(int(target.Fd()), unix.CLONE_NEWUTS); err != nil {
+		return utsHostnameResult{
+			err: errors.Join(
+				fmt.Errorf("enter target UTS namespace %q: %w", targetPath, err),
+				closeNamespaceHandle(target, "target UTS namespace"),
+				closeNamespaceHandle(current, "current thread UTS namespace"),
+			),
+		}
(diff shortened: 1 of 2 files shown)


huhong-web (commented):
Reviewed immutable head `86a1357ccc87` against base `de8dadafdfb0` (merge base `de8dadafdfb0`).

No actionable findings found. Validation included immutable diff and caller inspection of the locked-thread UTS-namespace enter/restore lifecycle and joined errors, plus `git diff --check`. Package tests could not be rerun because the review environment lacked enough disk space to materialize the immutable worktree. Residual risk: The real UTS `setns` restoration path still requires privileged Linux/kernel validation.

---
*Co-authored by: gpt-5.6-sol*

PR #632, opened 2026-08-17
fix(netutil): restore namespace after cookie lookup

### LINKED ISSUE #631: [Bug]: NetNSCookieByPid can return contaminated scheduler threads
## 这个问题是否已经存在？

## 部署环境

- Operating system: Linux 5.14+ for `SO_NETNS_COOKIE`; the namespace restoration defect also exists on older kernels after a successful `setns`
- Affected component: `internal/utils/netutil.NetNSCookieByPid`
- Trigger: container discovery calls this helper for each container init PID
- Required privilege: permission to enter the target network namespace

## 如何复现

`NetNSCookieByPid` locks the current goroutine to an operating-system thread, enters `/proc/<pid>/ns/net`, and defers a best-effort restoration:

```go
runtime.LockOSThread()

curNS, _ := os.Open("/proc/self/ns/net")
...
unix.Setns(int(targetNS.Fd()), unix.CLONE_NEWNET)
defer unix.Setns(int(curNS.Fd()), unix.CLONE_NEWNET) // error ignored
```

There are two correctness problems.

### 1. Restoration failure is ignored

If the deferred `setns` fails, `runtime.UnlockOSThread` still returns the network-namespace-mutated thread to the Go scheduler. Unrelated goroutines can then run in a container network namespace and observe different interfaces, routes, DNS behavior, socket defaults, and network namespace cookies.

### 2. `/proc/self/ns/net` does not explicitly identify the locked calling thread

Namespace transitions are task/thread scoped. `/proc/thread-self/ns/net` is the unambiguous handle for the operating-system thread that was locked and will execute `setns`. The helper should snapshot that exact namespace before entering the target.

A privileged regression can use two network namespaces:

1. lock a goroutine to its operating-system thread;
2. record `/proc/thread-self/ns/net`;
3. call the current cookie lookup for a process in another network namespace;
4. force or simulate restoration failure;
5. observe that the helper returns while the thread remains in the target namespace and is unlocked for scheduler reuse.

The normal container discovery path calls `NetNSCookieByPid` from `kubeletSyncContainers`, so this is not an artificial direct-call scenario.

### Actual behavior

The target namespace is entered correctly, but returning to the original namespace is best effort. A failure is silent and the contaminated thread is reused.

### Expected behavior

- Snapshot `/proc/thread-self/ns/net` for the exact locked worker thread.
- Restore it after all target operations, including socket and `getsockopt` failures.
- Return restoration errors together with the primary operation error.
- Never unlock a thread whose network namespace could not be restored.
- Close the namespace handles and socket on every path while preserving cleanup errors.
- Preserve the existing `0, nil` compatibility result for kernels that return `ENOPROTOOPT` for `SO_NETNS_COOKIE`.

## 修复建议或方案

Run the namespace-sensitive operation in a dedicated goroutine that owns a locked operating-system thread:

1. open `/proc/thread-self/ns/net`;
2. open `/proc/<pid>/ns/net`;
3. enter the target namespace;
4. create the datagram socket and read `SO_NETNS_COOKIE`;
5. restore the saved namespace in a mandatory cleanup path;
6. unlock only after successful restoration;
7. if restoration fails, return the joined error and let the locked goroutine exit so the Go runtime terminates the contaminated thread.

Use a small unexported operations seam for deterministic tests. Unit coverage should verify success, `ENOPROTOOPT`, socket failure, `getsockopt` failure, restoration failure, target-entry failure, and descriptor/socket cleanup without requiring privileged namespaces in CI.

## 其他

A network namespace leak is more severe than an incorrect metric value. It changes the execution environment of an arbitrary reusable Go runtime thread. The impact can extend to later sockets and filesystem reads of network state executed by unrelated collectors.

I can submit the focused lifecycle fix and run unit tests, race tests, vet, configured lint, and repository diff checks.

tju-yxq:
## Summary

- rebase the lifecycle fix onto current `main` and the renamed
  `NetNamespaceCookieByPID` API;
- save the network namespace of the exact locked worker thread;
- restore that namespace before returning the thread to the Go scheduler;
- discard the locked OS thread when restoration fails;
- preserve the existing `ENOPROTOOPT` compatibility result;
- make lock/unlock behavior deterministic and observable in unit tests;
- add a privileged integration regression for real procfs, `setns`, namespace
  cookies, and subsequent scheduler-thread namespace state.

## Problem

`NetNamespaceCookieByPID` previously entered `/proc/<pid>/ns/net` on a locked
OS thread and deferred a best-effort restoration. The restoration error was
ignored, but `runtime.UnlockOSThread` still ran. Because network namespace
membership is thread scoped, a failed restoration could put that mutated
thread back into the scheduler pool. Unrelated goroutines could then execute
inside a container network namespace.

The helper also opened `/proc/self/ns/net`. This change uses
`/proc/thread-self/ns/net` so the saved handle unambiguously belongs to the
exact OS thread performing both `setns` calls.

The branch has been rebased onto current `main` at
`a1e74a8ef58499c98368e61fa892f168e989bb87`. It keeps the current exported API
introduced by the intervening refactor:

```go
NetNamespaceInumByPID
NetNamespaceCookieByPID
```

The production caller in `internal/pod` compiles against that API.

## Implementation

The public `NetNamespaceCookieByPID` API delegates the namespace-sensitive
work to a dedicated goroutine which:

1. locks its OS thread;
2. opens `/proc/thread-self/ns/net` and the target namespace;
3. enters the target namespace;
4. creates the datagram socket and reads `SO_NETNS_COOKIE`;
5. restores the saved namespace on every post-entry result path;
6. unlocks only after successful restoration.

If restoration fails, the result includes that error and the locked goroutine
exits without calling `runtime.UnlockOSThread`. This prevents the contaminated
OS thread from being returned to the scheduler pool.

The worker now sends its result only after the unlock decision. This makes the
lifecycle complete before the caller returns and lets tests deterministically
assert the exact lock/unlock counts.

Small unexported operation seams cover namespace, socket, and thread-lifecycle
operations. Production still uses `runtime.LockOSThread`,
`runtime.UnlockOSThread`, `os.Open`, and the real `unix` syscalls.

## Error behavior

- `ENOPROTOOPT` still returns `0, nil` when all cleanup succeeds.
- Socket creation and `getsockopt` failures still trigger restoration.
- Target-entry failures close both namespace handles without attempting socket
  operations.
- `errors.Join` preserves primary operation, restoration, socket-close, and
  namespace-handle close failures.
- A successful cookie can be returned with a non-nil cleanup error when the
  value was read but lifecycle completion failed.
- Restoration failure never invokes the injected or production unlock
  operation.

## Deterministic regression coverage

The unit tests verify:

- successful target entry, cookie retrieval, restoration order, and cleanup;
- successful restoration locks once and unlocks exactly once;
- restoration failure locks once and never unlocks;
- the Linux <5.14 `ENOPROTOOPT` compatibility path;
- restoration after socket creation failure;
- joined restoration and socket-close failures;
- target `setns` failure without socket creation;
- target namespace open failure and current-handle cleanup;
- preservation of `getsockopt` and both namespace-close errors.

A negative mutation that replaced the conditional unlock with an unconditional
unlock fails `TestNetNamespaceCookieReportsRestoreAndSocketCloseFailures` with
`unlockCalls = 1, want 0`.

## Privileged integration coverage

`integration/test_netnamespace_cookie.sh` uses the root integration harness and
`integration/lib_namespace.sh` to:

1. create a target network namespace with a live process;
2. build an integration-tagged Go probe;
3. read `SO_NETNS_COOKIE` directly from a socket executed inside the target
   namespace;
4. call the public `NetNamespaceCookieByPID` helper from the host namespace;
5. compare the public helper result with the independently read target cookie;
6. run 64 subsequent goroutines locked to OS threads and verify every one still
   observes the host network namespace.

The test skips cleanly when the kernel does not support `SO_NETNS_COOKIE`.

## Validation

Passed on the final rebased head:

```text
go test ./internal/utils/netutil ./internal/pod -count=1
go test -race ./internal/utils/netutil -count=1
go vet ./internal/utils/netutil ./internal/pod
golangci-lint run -v ./internal/utils/netutil ./internal/pod \
  --timeout=5m --config .golangci.yaml
docker run --rm --privileged ... \
  bash integration/run.sh test_netnamespace_cookie.sh
make check
git diff --check
```

The privileged integration test passed twice on the development container,
with target cookies `28673` and `12291`. The full configured lint completed
with zero issues. `make check` passed against an LF-preserving archive of the
current upstream base with this patch staged.

Fixes #631

### DIFF
--- a/internal/utils/netutil/netns.go
+++ b/internal/utils/netutil/netns.go
@@ -24,6 +24,32 @@ import (
 	"golang.org/x/sys/unix"
 )
 
+const currentThreadNetNamespace = "/proc/thread-self/ns/net"
+
+type namespaceHandle interface {
+	Fd() uintptr
+	Close() error
+}
+
+type netNamespaceCookieOperations struct {
+	open       func(string) (namespaceHandle, error)
+	setns      func(int, int) error
+	socket     func(int, int, int) (int, error)
+	getsockopt func(int, int, int) (uint64, error)
+	close      func(int) error
+}
+
+type netNamespaceThreadOperations struct {
+	lock   func()
+	unlock func()
+}
+
+type netNamespaceCookieResult struct {
+	cookie        uint64
+	err           error
+	discardThread bool
+}
+
 // NetNamespaceInumByPID returns the network namespace inum for pid.
 func NetNamespaceInumByPID(pid int) (uint64, error) {
 	netnsStat, err := os.Stat(fmt.Sprintf("/proc/%d/ns/net", pid))
@@ -36,38 +62,121 @@ func NetNamespaceInumByPID(pid int) (uint64, error) {
 // NetNamespaceCookieByPID returns the network namespace cookie for pid.
 // Requires Linux 5.14+ (SO_NETNS_COOKIE). Returns 0, nil on older kernels.
 func NetNamespaceCookieByPID(pid int) (uint64, error) {
-	runtime.LockOSThread()
-	defer runtime.UnlockOSThread()
+	targetPath := fmt.Sprintf("/proc/%d/ns/net", pid)
+	return netNamespaceCookie(targetPath, netNamespaceCookieOperations{
+		open: func(path string) (namespaceHandle, error) {
+			return os.Open(path)
+		},
+		setns:      unix.Setns,
+		socket:     unix.Socket,
+		getsockopt: unix.GetsockoptUint64,
+		close:      unix.Close,
+	})
+}
 
-	curNS, err := os.Open("/proc/self/ns/net")
-	if err != nil {
-		return 0, fmt.Errorf("open self netns: %w", err)
-	}
-	defer curNS.Close()
+func netNamespaceCookie(
+	targetPath string,
+	operations netNamespaceCookieOperations,
+) (uint64, error) {
+	return netNamespaceCookieWithThreadOperations(
+		targetPath,
+		operations,
+		netNamespaceThreadOperations{
+			lock:   runtime.LockOSThread,
+			unlock: runtime.UnlockOSThread,
+		},
+	)
+}
 
-	targetNS, err := os.Open(fmt.Sprintf("/proc/%d/ns/net", pid))
-	if err != nil {
-		return 0, fmt.Errorf("open pid %d netns: %w", pid, err)
-	}
-	defer targetNS.Close()
-
-	if err := unix.Setns(int(targetNS.Fd()), unix.CLONE_NEWNET); err != nil {
-		return 0, fmt.Errorf("setns pid %d: %w", pid, err)
-	}
-	defer unix.Setns(int(curNS.Fd()), unix.CLONE_NEWNET) //nolint:errcheck // best-effort restore
-
-	fd, err := unix.Socket(unix.AF_INET, unix.SOCK_DGRAM|unix.SOCK_CLOEXEC, 0)
-	if err != nil {
-		return 0, fmt.Errorf("socket: %w", err)
-	}
-	defer unix.Close(fd)
-
-	cookie, err := unix.GetsockoptUint64(fd, unix.SOL_SOCKET, unix.SO_NETNS_COOKIE)
-	if err != nil {
-		if errors.Is(err, unix.ENOPROTOOPT) {
-			return 0, nil
+func netNamespaceCookieWithThreadOperations(
+	targetPath string,
+	operations netNamespaceCookieOperations,
+	threadOperations netNamespaceThreadOperations,
+) (uint64, error) {
+	resultCh := make(chan netNamespaceCookieResult, 1)
+	go func() {
+		threadOperations.lock()
+		result := readNetNamespaceCookie(targetPath, operations)
+		if !result.discardThread {
+			threadOperations.unlock()
 		}
-		return 0, fmt.Errorf("getsockopt SO_NETNS_COOKIE: %w", err)
+		resultCh <- result
+		// A goroutine that exits while locked takes a namespace-contaminated
+		// operating-system thread out of the scheduler pool.
+	}()
+
+	result := <-resultCh
+	return result.cookie, result.err
+}
+
+func readNetNamespaceCookie(
+	targetPath string,
+	operations netNamespaceCookieOperations,
+) netNamespaceCookieResult {
+	current, err := operations.open(currentThreadNetNamespace)
+	if err != nil {
+		return netNamespaceCookieResult{
+			err: fmt.Errorf("open current thread network namespace: %w", err),
+		}
 	}
-	return cookie, nil
+
+	target, err := operations.open(targetPath)
+	if err != nil {
+		return netNamespaceCookieResult{err: errors.Join(
+			fmt.Errorf("open target network namespace %q: %w", targetPath, err),
+			closeNamespaceHandle(current, "current thread network namespace"),
+		)}
+	}
+
+	if err := operations.setns(int(target.Fd()), unix.CLONE_NEWNET); err != nil {
+		return netNamespaceCookieResult{err: errors.Join(
+			fmt.Errorf("enter target network namespace %q: %w", targetPath, err),
+			closeNamespaceHandle(target, "target network namespace"),
+			closeNamespaceHandle(current, "current thread network namespace"),
+		)}
+	}
+
+	cookie, operationErr := readCookieFromSocket(operations)
+	restoreErr := operations.setns(int(current.Fd()), unix.CLONE_NEWNET)
+	if restoreErr != nil {
+		restoreErr = fmt.Errorf("restore current thread network namespace: %w", restoreErr)
+	}
+
+	return netNamespaceCookieResult{
+		cookie: cookie,
+		err: errors.Join(
+			operationErr,
+			restoreErr,
+			closeNamespaceHandle(target, "target network namespace"),
+			closeNamespaceHandle(current, "current thread network namespace"),
+		),
+		discardThread: restoreErr != nil,
+	}
+}
+
+func readCookieFromSocket(operations netNamespaceCookieOperations) (uint64, error) {
+	fd, err := operations.socket(unix.AF_INET, unix.SOCK_DGRAM|unix.SOCK_CLOEXEC, 0)
+	if err != nil {
+		return 0, fmt.Errorf("create network namespace cookie socket: %w", err)
+	}
+
+	cookie, cookieErr := operations.getsockopt(fd, unix.SOL_SOCKET, unix.SO_NETNS_COOKIE)
+	if errors.Is(cookieErr, unix.ENOPROTOOPT) {
+		cookieErr = nil
+		cookie = 0
+	} else if cookieErr != nil {
+		cookieErr = fmt.Errorf("read SO_NETNS_COOKIE: %w", cookieErr)
+	}
+	closeErr := operations.close(fd)
+	if closeErr != nil {
+		closeErr = fmt.Errorf("close network namespace cookie socket: %w", closeErr)
+	}
+	return cookie, errors.Join(cookieErr, closeErr)
+}
+
+func closeNamespaceHandle(handle namespaceHandle, description string) error {
+	if err := handle.Close(); err != nil {
+		return fmt.Errorf("close %s: %w", description, err)
+	}
+	return nil
 }
--- a/internal/utils/netutil/netns_test.go
+++ b/internal/utils/netutil/netns_test.go
@@ -15,8 +15,13 @@
 package netutil
 
 import (
+	"errors"
 	"os"
 	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"golang.org/x/sys/unix"
 )
 
 func TestNetNamespaceInumByPID(t *testing.T) {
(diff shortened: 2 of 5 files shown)


huhong-web (commented):
Reviewed immutable head `c714690e17b8` against base `de8dadafdfb0` (merge base `de8dadafdfb0`).

No actionable findings found. Validation included immutable diff and caller inspection of the locked-thread network-namespace enter/restore lifecycle and joined errors, plus `git diff --check`. Package tests could not be rerun because the review environment lacked enough disk space to materialize the immutable worktree. Residual risk: The real `setns` restoration path still requires privileged Linux/kernel validation.

---
*Co-authored by: gpt-5.6-sol*

internal/utils/netutil/netns.go
```diff
@@ -36,38 +57,101 @@ func NetNSInumByPid(pid int) (uint64, error) {
 // NetNSCookieByPid returns the network namespace cookie for the given pid.
 // Requires Linux 5.14+ (SO_NETNS_COOKIE). Returns 0, nil on older kernels.
 func NetNSCookieByPid(pid int) (uint64, error) {
```
huhong-web: [P2] Rebase this fix onto the current netutil API

At review time, `main` is at `7ee5c471`. Commit `9d1d3315` renamed this helper to `NetNamespaceCookieByPID` (and `NetNSInumByPid` to `NetNamespaceInumByPID`), and `internal/pod/container_kubelet_sync.go` now calls that spelling. This head is 42 commits behind `main`, and GitHub reports it as conflicting. Applying the change under the obsolete exported name would either fail during the rebase or regress the current API contract.

Please rebase onto current `main` and apply the lifecycle fix to `NetNamespaceCookieByPID`, keeping the related unexported types/helpers/tests consistent with the current `netNamespace...` naming. As an acceptance check, `go test ./internal/utils/netutil ./internal/pod` should compile and pass on the rebased head.

internal/utils/netutil/netns.go
```diff
@@ -36,38 +57,101 @@ func NetNSInumByPid(pid int) (uint64, error) {
...
+	})
+}
+
+func netNSCookie(targetPath string, operations netNSCookieOperations) (uint64, error) {
+	resultCh := make(chan netNSCookieResult, 1)
+	go func() {
+		runtime.LockOSThread()
+		result := readNetNSCookie(targetPath, operations)
+		resultCh <- result
+		if !result.discardThread {
```
huhong-web, line 100: [P2] Make the no-unlock guarantee observable in tests

This branch is the safety boundary that prevents a thread left in the target network namespace from re-entering the Go scheduler. `TestNetNSCookieReportsRestoreAndSocketCloseFailures` simulates a restore failure, but it only checks returned errors. In a temporary review checkout, replacing this conditional with an unconditional `runtime.UnlockOSThread()` still left every `internal/utils/netutil` test passing, even though that recreates the namespace-contamination bug.

Please add a deterministic lifecycle seam/assertion around the worker so tests prove both outcomes: successful restoration unlocks exactly once, while restoration failure never unlocks. A focused privileged integration case using the existing namespace harness would additionally validate real `/proc/thread-self` and `setns` behavior, but the no-unlock invariant itself should be covered deterministically.

huhong-web (commented):
## Code Review

I found two P2 issues; both are attached as inline comments.

### Architecture & Design

**Verdict: needs changes.**

The affected call chain remains appropriately bounded: `internal/pod.kubeletUpdateContainer` calls the netutil helper, which owns a dedicated locked worker thread while it enters the target network namespace, reads `SO_NETNS_COOKIE`, restores the original namespace, closes its resources, and transfers one result back to the caller. Using a dedicated goroutine is justified here: if restoration fails, allowing that goroutine to exit while still locked prevents the namespace-contaminated OS thread from returning to the scheduler. The small operations abstraction is also a proportionate test seam, and `errors.Join` preserves independent operation, restoration, and cleanup failures.

Two cross-layer contract gaps remain:

1. The branch is 42 commits behind the current `main` tip and conflicts with the already-renamed netutil API consumed by `internal/pod`. The lifecycle fix must be rebased and applied to the current API rather than restoring obsolete exported names.
2. The defining safety invariant—never calling `runtime.UnlockOSThread` after namespace restoration fails—is not observable in the tests. A mutation that made the unlock unconditional still passed the entire target package test suite.

No storage, configuration, migration, or external dependency contracts are changed by this PR.

### Other review dimensions

- **Concurrency/lifecycle:** the one-shot buffered result channel has clear ownership, the worker has a finite exit path, and the production discard-thread approach is sound; the missing assertion above prevents regression protection.
- **Errors:** wrapping and joined-error semantics preserve causes and let the caller handle the result once.
- **Security:** no injection, path traversal, secret, cryptographic, or unsafe input-construction issue was found. The PID is typed as `int`, syscall arguments are fixed, and the socket uses `SOCK_CLOEXEC`.
- **Style/licensing:** both modified files retain the required 2026 Apache 2.0 header; no line-length, whitespace, or conflict-marker issue was found.
- **Compatibility:** `/proc/thread-self/ns/net` predates the repository's supported Linux kernel floor, and the existing `ENOPROTOOPT` compatibility behavior is preserved.

### Verification

Reviewed PR head `c714690e17b8f6f330b0828731c77ae8410ddfc7` against merge base `de8dadafdfb052dbdccc789003c2564db866c85c`, then checked the current `main` tip `7ee5c47142af416b5bc3ceb721dc3955c985c581`.

- `go test ./internal/utils/netutil -count=1`: pass
- `go test -race ./internal/utils/netutil -count=1`: pass
- `go vet ./internal/utils/netutil`: pass
- `git diff --check de8dadaf..c714690e`: pass
- Negative mutation (unconditional unlock): tests still pass, confirming finding 2
- Repository lint was not reproducible locally because the installed golangci-lint v2.5.0 cannot parse this repository's older config format
- GitHub reports no completed check suite for the branch; the three workflow runs ended as `action_required`, so they provide no build/lint/QEMU evidence

---
*Co-authored by: gpt-5.6-sol*

internal/utils/netutil/netns_test.go
```diff
@@ -54,3 +59,260 @@ func TestNetNSInumByPid(t *testing.T) {
...
+func (h *fakeNamespaceHandle) Fd() uintptr {
+	return h.fd
+}
+
+func (h *fakeNamespaceHandle) Close() error {
+	h.closed = true
+	return h.closeErr
+}
+
+func TestNetNSCookieRestoresCallingThread(t *testing.T) {
```
huhong-web: [P2] Add a real namespace-boundary regression

This test injects fake namespace handles and a fake `setns`, so it verifies call order but never exercises the contract this patch depends on: `/proc/thread-self/ns/net` must identify the locked worker, the real target fd must be accepted by `setns`, and a successful call must leave scheduler threads in the original namespace. A wrong proc path or real syscall interaction could therefore ship while every new unit test passes.

Please add the smallest privileged Linux case to the existing root integration harness (`integration/run.sh` and `integration/lib_namespace.sh`): create a target netns with a live PID, call the rebased public helper, compare its cookie with `SO_NETNS_COOKIE` read inside that netns, and verify subsequent locked-thread work still observes the host netns. Keep the fake tests for deterministic restore-failure/no-unlock coverage; the two layers prove different contracts.

huhong-web (commented):
## Supplemental Code Review

One additional P2 verification gap is attached inline. Together with the two existing open findings, this PR has three actionable issues.

### Architecture & Design

**Verdict: needs changes.**

The production boundary is clear: `internal/pod.kubeletUpdateContainer` calls netutil, and netutil owns a dedicated goroutine whose locked OS thread enters the target network namespace, reads `SO_NETNS_COOKIE`, restores the saved namespace, and returns one result. The dedicated worker is justified: doing this on the caller's thread cannot safely return a restoration error while also preventing a contaminated thread from re-entering the scheduler. The small operations seam and `errors.Join` are proportionate.

The remaining contract gaps are:

1. Rebase the branch onto current `main` and port the fix to the renamed `NetNamespaceCookieByPID` API.
2. Make the restore-failure/no-unlock branch deterministic and observable in unit tests.
3. Exercise the real privileged Linux namespace boundary; the current fakes do not prove `/proc/thread-self`, fd, `setns`, or scheduler-restoration behavior.

No storage, configuration, migration, or security contract is changed. No injection, traversal, secret, cryptographic, or unsafe-input finding was identified.

### Verification

Reviewed head `c714690e17b8f6f330b0828731c77ae8410ddfc7` against merge base `de8dadafdfb052dbdccc789003c2564db866c85c` and current `main` `7ee5c47142af416b5bc3ceb721dc3955c985c581`.

- `go test -p=1 -vet=off ./internal/utils/netutil -count=1`: pass
- `go test -p=1 -vet=off -race ./internal/utils/netutil -count=1`: pass
- `go vet -p=1 ./internal/utils/netutil`: pass
- `gofmt -d internal/utils/netutil/netns.go internal/utils/netutil/netns_test.go`: pass
- `git diff --check de8dadaf...c714690e`: pass
- Configured golangci-lint could not run because installed v2.5.0 rejects the repository's legacy config format.
- GitHub reports no check suite; the three workflow runs ended as `action_required`.

---
*Co-authored by: gpt-5.6-sol*

internal/utils/netutil/netns.go
tju-yxq: Addressed in `228435b3110ef2b71e34ad3e401bb5ab4de8322f`.

The worker has an unexported thread-lifecycle seam backed by `runtime.LockOSThread` and `runtime.UnlockOSThread` in production. The result is sent only after the unlock decision, so the caller cannot return before the lifecycle assertion is observable.

Tests now prove both branches:

- successful restoration: lock once, unlock exactly once;
- restoration failure: lock once, unlock zero times.

I also ran the negative mutation described in the review. Replacing the conditional unlock with an unconditional call fails `TestNetNamespaceCookieReportsRestoreAndSocketCloseFailures` because `unlockCalls` becomes 1 instead of 0. Restoring the production conditional makes the test pass again.

internal/utils/netutil/netns.go
tju-yxq: Addressed in `228435b3110ef2b71e34ad3e401bb5ab4de8322f`.

The branch is rebased onto current `main` at `a1e74a8ef58499c98368e61fa892f168e989bb87`. The implementation and all related unexported helpers/tests now use the current `NetNamespaceCookieByPID` / `netNamespace...` naming; the obsolete exported spelling is not reintroduced.

Acceptance checks pass on the rebased head:

```text
go test ./internal/utils/netutil ./internal/pod -count=1
go vet ./internal/utils/netutil ./internal/pod
```

GitHub now reports the PR as `MERGEABLE` rather than conflicting.

internal/utils/netutil/netns_test.go
tju-yxq: Addressed in `228435b3110ef2b71e34ad3e401bb5ab4de8322f`.

Added `integration/test_netnamespace_cookie.sh`, an integration-tagged Go probe, and reusable single-process namespace setup/cleanup in `integration/lib_namespace.sh`.

The privileged test:

1. creates a target network namespace with a live PID;
2. reads `SO_NETNS_COOKIE` directly from a socket executed inside that namespace;
3. calls the public `NetNamespaceCookieByPID` helper from the host namespace;
4. compares the two cookie values;
5. runs 64 subsequent goroutines locked to OS threads and verifies each still observes the host network namespace.

It passed twice in the privileged development container, with target cookies `28673` and `12291`. It skips cleanly when `SO_NETNS_COOKIE` is unavailable. The fake unit tests remain in place for deterministic restoration-failure/no-unlock coverage.

PR #634, opened 2026-08-17
fix(config): sync directory after atomic replace

### LINKED ISSUE #633: [Bug]: Config Sync returns before rename is durably committed
## 这个问题是否已经存在？

## 部署环境

- Operating system: Linux on a filesystem that requires a directory `fsync` to make rename metadata durable, including common ext4 and XFS deployments
- Affected component: `internal/config.Sync`
- Production caller: `cmd/huatuo-bamai/config.UpdateAndSync`, used to persist runtime configuration updates

## 如何复现

`internal/config.Sync` correctly stages the TOML in the destination directory, copies the existing permissions, flushes and closes the temporary file, and atomically renames it over the configured path:

```go
if err := f.Sync(); err != nil {
    return fmt.Errorf("syncing temporary config file: %w", err)
}
if err := f.Close(); err != nil {
    return fmt.Errorf("closing temporary config file: %w", err)
}
if err := os.Rename(tempPath, path); err != nil {
    return fmt.Errorf("replacing config file: %w", err)
}
return nil
```

The file data is durable before the rename, but the rename itself is directory metadata. On Linux, successful `rename(2)` only makes the replacement visible atomically to running processes. It does not guarantee that the directory entry survives a power loss or kernel crash. The containing directory must be opened and synchronized after the rename.

A crash-window reproduction is:

1. start HUATUO with a configuration file on a local filesystem;
2. submit an update that reaches `UpdateAndSync`;
3. observe the request succeed and the replacement file become visible;
4. crash or power-cycle the host before unrelated writeback synchronizes the parent directory;
5. after recovery, observe that the old directory entry may remain or the new entry may be absent despite `Sync` having returned success.

This outcome depends on filesystem and crash timing, so it is not a reliable ordinary unit test. The missing durability barrier is directly visible in the syscall sequence: `fsync(temp)`, `close(temp)`, `rename(temp, path)`, return, with no `open(parent)` and `fsync(parent)`.

### Actual behavior

`Sync` reports success after `os.Rename`. The caller then publishes the new runtime configuration. A crash can therefore leave the live state before the crash and the persisted state after reboot inconsistent, even though the persistence call returned nil.

### Expected behavior

After a successful rename, `Sync` must synchronize the containing directory before reporting success. If opening or syncing the directory fails, the error must be returned with enough context to identify that the replacement is visible but its directory entry could not be confirmed durable.

The existing guarantees must remain intact:

- the target is replaced atomically, never rewritten in place;
- the temporary file is flushed before rename;
- existing target permissions are preserved;
- pre-rename failures leave the original file unchanged and remove the temporary file;
- no temporary file remains after success;
- the public `Sync(path, src)` API does not change.

## 修复建议或方案

Add a small unexported filesystem-operations seam around temporary-file creation, stat, rename, and directory open so the durability sequence and failure behavior can be tested without inducing a machine crash.

The successful sequence should be:

1. create the temporary file in `filepath.Dir(path)`;
2. encode, `fsync`, and close it;
3. rename it over `path`;
4. open the containing directory read-only;
5. call `Sync` on the directory handle;
6. close the directory handle and report any sync/close error.

Tests should verify exact ordering and the following boundaries:

- directory sync runs only after a successful rename;
- directory-open, directory-sync, and directory-close errors are preserved;
- a rename failure does not attempt directory sync and still removes the temporary file;
- a directory-sync failure does not remove the already-renamed destination;
- the real filesystem integration test still reloads the complete TOML, preserves permissions, and leaves no temporary files.

## 其他

This is a persistence correctness issue rather than a request for stronger global synchronization. PR #532 addresses atomic publication of runtime updates after validation and persistence; it does not add the post-rename directory durability barrier inside `internal/config.Sync`.

The patch should stay focused on the persistence primitive and its tests. No storage schema, configuration option, or public API change is required.

tju-yxq:
## Summary

- synchronize the containing directory after atomically replacing a persisted
  configuration file;
- return contextual directory open, sync, and close failures;
- add deterministic durability-order and cleanup tests;
- retain real-filesystem coverage for content, permissions, and temporary-file
  cleanup.

## Problem

`internal/config.Sync` flushed and closed its staged file before atomically
renaming it over the destination, but returned immediately after `os.Rename`.
The file data was durable, while the rename remained directory metadata that
had not been synchronized.

On Linux filesystems, rename atomicity and crash durability are separate
properties. A process sees the replacement atomically after a successful
rename, but a power loss or kernel crash can occur before that directory-entry
update reaches stable storage. `UpdateAndSync` could therefore report success
and publish the new runtime configuration even though the persisted replacement
was not yet durably committed.

## Implementation

`Sync(path, src)` keeps its public API and existing staging behavior. After the
temporary file is encoded, flushed, closed, and renamed, it now:

1. opens `filepath.Dir(path)`;
2. calls `Sync` on the directory handle;
3. closes the directory handle;
4. returns joined, contextual sync and close errors.

The sequence is now:

```text
create temp -> chmod -> encode -> file fsync -> file close -> rename
-> directory open -> directory fsync -> directory close
```

A small unexported operations seam makes the ordering and post-rename failures
deterministic in unit tests. Production operations still delegate directly to
`os.CreateTemp`, `os.Stat`, `os.Rename`, `os.Remove`, and `os.Open`.

## Error behavior

- A rename failure does not open or synchronize the directory.
- Pre-rename failures continue to close and remove the staged temporary file.
- Directory open, sync, and close failures are returned with distinct context.
- Simultaneous directory sync and close failures are preserved with
  `errors.Join`.
- After a successful rename, cleanup only addresses the old temporary path; it
  never removes the replacement destination.
- Existing target permissions remain preserved.

## Regression coverage

The tests verify:

- exact successful operation order, including directory sync after rename;
- directory-open failure propagation;
- directory-sync failure propagation;
- directory-close failure propagation;
- preservation of simultaneous sync and close errors;
- no directory operation after rename failure;
- temporary-file cleanup after rename and durability failures;
- successful TOML reload from the real filesystem;
- preservation of an existing `0640` target mode;
- absence of temporary files after successful replacement.

## Validation

Passed:

```text
go test ./internal/config -count=1
go test -race ./internal/config -count=1
go vet ./internal/config
golangci-lint run -v ./internal/config --timeout=5m --config .golangci.yaml
git diff --check
```

Both configured lint runs completed with zero issues. `make check` was also
executed in a container-local copy: generation and lint completed, while its
final `git diff --exit-code` exposed checkout-wide CRLF conversions introduced
by copying the Windows working tree into that container. Those unrelated
line-ending changes are not present in this branch.

Fixes #633

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -19,6 +19,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"reflect"
@@ -33,6 +34,27 @@ var (
 	// CoreBpfDir is the directory where BPF object files are stored.
 	CoreBpfDir = ""
 )
+
+type syncFile interface {
+	io.Writer
+	Name() string
+	Chmod(mode os.FileMode) error
+	Sync() error
+	Close() error
+}
+
+type syncDirectory interface {
+	Sync() error
+	Close() error
+}
+
+type syncOperations struct {
+	createTemp    func(string, string) (syncFile, error)
+	stat          func(string) (os.FileInfo, error)
+	rename        func(string, string) error
+	remove        func(string) error
+	openDirectory func(string) (syncDirectory, error)
+}
 
 func init() {
 	if exePath, err := os.Executable(); err == nil {
@@ -55,8 +77,22 @@ func Load(path string, dst any) error {
 // Sync atomically replaces path with the TOML encoding of src. Callers must
 // prevent concurrent mutation of src until Sync returns.
 func Sync(path string, src any) (retErr error) {
+	return syncConfig(path, src, syncOperations{
+		createTemp: func(dir, pattern string) (syncFile, error) {
+			return os.CreateTemp(dir, pattern)
+		},
+		stat:   os.Stat,
+		rename: os.Rename,
+		remove: os.Remove,
+		openDirectory: func(path string) (syncDirectory, error) {
+			return os.Open(path)
+		},
+	})
+}
+
+func syncConfig(path string, src any, operations syncOperations) (retErr error) {
 	dir := filepath.Dir(path)
-	f, err := os.CreateTemp(dir, "."+filepath.Base(path)+".tmp-*")
+	f, err := operations.createTemp(dir, "."+filepath.Base(path)+".tmp-*")
 	if err != nil {
 		return fmt.Errorf("creating temporary config file: %w", err)
 	}
@@ -68,13 +104,13 @@ func Sync(path string, src any) (retErr error) {
 		if err := f.Close(); err != nil && !errors.Is(err, os.ErrClosed) {
 			retErr = errors.Join(retErr, fmt.Errorf("closing temporary config file: %w", err))
 		}
-		if err := os.Remove(tempPath); err != nil && !errors.Is(err, os.ErrNotExist) {
+		if err := operations.remove(tempPath); err != nil && !errors.Is(err, os.ErrNotExist) {
 			retErr = errors.Join(retErr, fmt.Errorf("removing temporary config file: %w", err))
 		}
 	}()
 
 	mode := os.FileMode(0o600)
-	info, err := os.Stat(path)
+	info, err := operations.stat(path)
 	switch {
 	case err == nil:
 		mode = info.Mode().Perm()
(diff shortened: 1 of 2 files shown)


huhong-web (commented):
Reviewed immutable head `58b41a7af337` against base `de8dadafdfb0` (merge base `de8dadafdfb0`).

No actionable findings found. Validation included immutable diff and caller inspection of the temp-file/fsync/rename sequence, error paths, and directory synchronization, plus `git diff --check`. Package tests could not be rerun because the review environment lacked enough disk space to materialize the immutable worktree. Residual risk: Filesystem crash-consistency behavior remains platform/filesystem dependent and should be exercised by CI or fault-injection coverage.

---
*Co-authored by: gpt-5.6-sol*

PR #636, opened 2026-08-17
fix(storage): restore shutdown cleanup

### LINKED ISSUE #635: [Bug]: Storage setup no longer drains backends during shutdown
## 这个问题是否已经存在？

## 部署环境

- Architecture: all supported architectures
- Operating system: all Linux deployments using a configured storage backend
- Affected component: `cmd/huatuo-bamai.setupStorage`
- Most visible backend: Elasticsearch, whose `BulkIndexer` buffers writes asynchronously
- Trigger: normal SIGINT/SIGTERM shutdown, dry-run completion, or a later daemon setup failure

## 如何复现

The daemon lifecycle records every non-nil cleanup returned by a setup step and runs those cleanups in reverse order:

```go
cleanup, err := setup(d)
...
if cleanup != nil {
    cleanups = append(cleanups, cleanup)
}
```

`setupStorage` currently initializes and registers storage stores but always returns a nil cleanup:

```go
func setupStorage(d *Daemon) (func(context.Context) error, error) {
    if d.opts.DisableStorage {
        return nil, nil
    }

    return nil, initStorage(d.opts.Region, config.Get())
}
```

`pkg/tracing.CloseStores` already exists specifically to close each configured store once and aggregate errors. Commit `25b3bba222bde9225a0aa287dc3e2099509464fc` originally called it on process shutdown to drain the Elasticsearch `BulkIndexer` before BPF teardown. During the later setup/cleanup-stack refactor, the storage step was converted to the new signature but returned `nil` instead of registering `tracing.CloseStores`.

Reproduction:

1. configure Elasticsearch storage and start `huatuo-bamai`;
2. generate tracing, task, or profiling documents so the bulk indexer has pending work;
3. send SIGTERM;
4. observe that the daemon runs other cleanup functions but never calls `tracing.CloseStores`;
5. pending asynchronous documents can be lost and backend resources are not released through `Backend.Close`.

The same omission occurs if a setup step after storage fails: the daemon unwinds earlier cleanup functions, but storage has not contributed one to the stack.

There is a second lifecycle boundary inside `initStorage`. It constructs up to three stores sequentially and publishes tracing/task writers before constructing the profiling store. If a later constructor fails, earlier stores are neither closed nor kept out of global writer state. The setup returns an error, but the failed setup step has no cleanup recorded, so those partially initialized resources remain live.

### Actual behavior

- Successful storage setup returns `cleanup == nil`.
- Graceful shutdown does not drain or close configured stores.
- A later daemon setup failure does not close stores initialized earlier.
- A mid-storage initialization failure can leak already-created stores and leave partially configured global writers.

### Expected behavior

- Disabled storage continues to return `(nil, nil)`.
- Successful storage setup returns a cleanup that closes each configured store exactly once with the daemon shutdown context.
- Store construction is transactional: no tracing/task/profile writer is published until every required store has been created.
- If any constructor fails, all stores created by that setup attempt are closed before the error is returned.
- Initialization and cleanup errors are both preserved.
- Cleanup remains idempotent with respect to a store registered under both tracing and task writers.

## 修复建议或方案

Keep store construction local until all enabled backends and collections are ready:

1. create the tracing Elasticsearch store when enabled;
2. create the tracing local-file store when configured;
3. create the profiling Elasticsearch store when enabled;
4. on any failure, close every store created in steps 1-3 and join cleanup errors with the constructor error;
5. only after successful construction, publish the tracing, task, and profile writers;
6. return `tracing.CloseStores` from `setupStorage` as the daemon cleanup.

Use an unexported store-constructor seam for deterministic tests. Regression coverage should verify:

- disabled storage returns no cleanup;
- successful setup returns a cleanup and closes all distinct stores;
- the Elasticsearch tracing store shared by tracing and task writers is closed once;
- a later constructor failure closes all earlier stores;
- partial writers are not published on failure;
- constructor and cleanup failures are joined;
- the caller-provided shutdown context reaches each backend close.

## 其他

This restores an existing shutdown guarantee rather than introducing a new feature. `Backend.Close`, `Store.Close`, and `tracing.CloseStores` were added to prevent buffered document loss, but the current daemon lifecycle no longer invokes them.

The issue is distinct from PR #577, which closes a backend when `storage.NewStore` itself fails during mapper or collection initialization. This issue covers stores that were successfully constructed by `initStorage` and then leaked because the overall storage setup or daemon lifecycle did not close them.

tju-yxq:
## Summary

- return storage cleanup to the daemon lifecycle after successful setup;
- give backend shutdown a single owner so every store closes exactly once;
- stop tracing collectors before draining stores, and drain stores before BPF
  teardown;
- construct all required stores transactionally before publishing writers;
- close already-created stores when a later constructor fails;
- preserve constructor and rollback-close errors together;
- add deterministic shutdown-order and partial-initialization regression tests.

## Problem

The daemon records each non-nil cleanup returned by a setup step and executes
those cleanups in reverse order during normal shutdown and startup rollback.
`setupStorage` initialized and registered stores but returned a nil cleanup:

```go
return nil, initStorage(d.opts.Region, config.Get())
```

As a result, storage did not own a cleanup in the daemon stack. At the same
time, `startTracing` closed both the tracing manager and all stores. That made
store lifetime depend on tracing startup: if a setup step failed after storage
but before tracing, the created stores were never closed.

Simply registering `tracing.CloseStores` in `setupStorage` without changing
`startTracing` would introduce a second owner. On normal shutdown the tracing
cleanup would close the Elasticsearch bulk indexer, and the later storage
cleanup would close it again. The vendored bulk indexer closes its queue
channel in `Close`, so the second close can panic with `close of closed channel`.

There was also a partial-initialization boundary inside `initStorage`. Tracing
and task writers were published before the profiling store was constructed. If
that later constructor failed, earlier stores remained open and global writers
could point at a setup attempt that had already returned an error.

### Single shutdown owner

When storage is disabled, `setupStorage` still returns `(nil, nil)`. After a
successful enabled setup, it now returns `tracing.CloseStores`. Storage is the
only daemon component that closes stores.

`startTracing` now owns only the tracing manager. Its cleanup stops and waits
for collectors, but does not close storage backends.

The daemon setup order places BPF before storage and tracing after storage:

```text
setup:   BPF -> storage -> tracing
cleanup: tracing -> storage -> BPF
```

This preserves the required lifecycle:

1. stop collectors so no new records are produced;
2. drain and close storage backends exactly once;
3. tear down BPF resources.

### Transactional construction

`initStorageWithFactory` keeps newly constructed stores in local slices while
creating:

1. the Elasticsearch tracing/task store, when enabled;
2. the local-file tracing store, when configured;
3. the Elasticsearch profiling store, when enabled.

Tracing, task, and profile writers are published only after all required
constructors succeed. A profiling-store failure can therefore no longer expose
partially configured global writers.

### Failure rollback

A deferred rollback closes every distinct store created by the failed setup
attempt. It deduplicates store pointers and joins close failures with the
original constructor error. This preserves both the reason initialization
failed and any resource-release failure.

The production constructor remains `storage.NewFromConfig`. The unexported
factory and manager-closer seams exist only to make lifecycle boundaries
deterministic in unit tests.

## Error and lifecycle behavior

- Disabled storage creates no stores and returns no cleanup.
- Successful enabled setup returns one storage cleanup function.
- Tracing cleanup stops collectors but never closes stores.
- The same Elasticsearch store registered for tracing and task output is
  closed once by `tracing.CloseStores`.
- Distinct tracing, local-file, and profiling stores are all closed even if one
  close operation fails.
- The daemon-provided shutdown context reaches every backend `Close` call.
- A later constructor failure closes all stores created earlier in that setup.
- Constructor and rollback-close errors remain discoverable through
  `errors.Is`.
- No tracing, task, or profile writer is published by a failed setup attempt.

## Regression coverage

The tests verify:

- the disabled path returns no cleanup;
- enabled setup returns an invokable cleanup;
- construction of both tracing stores and the profiling store;
- successful publication only after complete construction;
- shutdown closes all three distinct backends;
- the tracing Elasticsearch store shared by two writers is closed exactly
  once;
- tracing cleanup and storage cleanup execute in order without double-closing
  the backend;
- daemon setup order produces tracing, storage, then BPF cleanup order;
- close errors are returned without preventing other stores from closing;
- the exact shutdown context is propagated to every backend;
- profiling-constructor failure rolls back both earlier stores;
- constructor and rollback errors are joined;
- failed setup does not route tracing, task, or profiling writes to the rolled
  back backends.

## Validation

Passed after the shutdown-ownership revision:

```text
go test ./cmd/huatuo-bamai -count=1
go test -race ./cmd/huatuo-bamai -count=1
go vet ./cmd/huatuo-bamai
golangci-lint run -v ./cmd/huatuo-bamai --timeout=5m --config .golangci.yaml
make check
git diff --check
```

`make check` was executed against an LF-preserving archive of `upstream/main`
with this patch staged, inside the repository development image. Generation,
formatting, the configured full-repository lint, and the final diff check all
completed successfully.

Fixes #635

### DIFF
--- a/cmd/huatuo-bamai/storage.go
+++ b/cmd/huatuo-bamai/storage.go
@@ -16,6 +16,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 
 	"huatuo-bamai/cmd/huatuo-bamai/config"
@@ -27,21 +28,54 @@ import (
 	"huatuo-bamai/pkg/tracing"
 )
 
+type documentStoreFactory func(
+	context.Context,
+	*driver.Config,
+	string,
+	driver.Mapper[*tracing.Document],
+) (*storage.Store[*tracing.Document], error)
+
 func setupStorage(d *Daemon) (func(context.Context) error, error) {
 	if d.opts.DisableStorage {
 		log.Infof("storage backends disabled by --disable-storage")
 		return nil, nil
 	}
 
-	return nil, initStorage(d.opts.Region, config.Get())
+	if err := initStorage(d.opts.Region, config.Get()); err != nil {
+		return nil, err
+	}
+	return tracing.CloseStores, nil
 }
 
 func initStorage(storageRegion string, cfg *config.Config) error {
+	return initStorageWithFactory(
+		storageRegion,
+		cfg,
+		storage.NewFromConfig[*tracing.Document],
+	)
+}
+
+func initStorageWithFactory(
+	storageRegion string,
+	cfg *config.Config,
+	newStore documentStoreFactory,
+) (retErr error) {
 	var esStore *storage.Store[*tracing.Document]
+	var profileStore *storage.Store[*tracing.Document]
 
 	tracingMetadataStores := make([]*storage.Store[*tracing.Document], 0, 2)
+	initializedStores := make([]*storage.Store[*tracing.Document], 0, 3)
+	defer func() {
+		if retErr != nil {
+			retErr = errors.Join(
+				retErr,
+				closeInitializedStores(context.Background(), initializedStores),
+			)
+		}
+	}()
+
 	if cfg.Storage.Elasticsearch.Enabled() {
-		store, err := storage.NewFromConfig[*tracing.Document](context.Background(), &driver.Config{
+		store, err := newStore(context.Background(), &driver.Config{
 			Driver:      "elasticsearch",
 			ESAddresses: strutil.SplitCommaList(cfg.Storage.Elasticsearch.Address),
 			ESUsername:  cfg.Storage.Elasticsearch.Username,
@@ -53,10 +87,11 @@ func initStorage(storageRegion string, cfg *config.Config) error {
 		}
 		esStore = store
 		tracingMetadataStores = append(tracingMetadataStores, esStore)
+		initializedStores = append(initializedStores, esStore)
 	}
 
 	if cfg.Storage.LocalFile.Path != "" {
-		localFileStore, err := storage.NewFromConfig[*tracing.Document](context.Background(), &driver.Config{
+		localFileStore, err := newStore(context.Background(), &driver.Config{
 			Driver:                "localfile",
 			LocalFilePath:         cfg.Storage.LocalFile.Path,
 			LocalFileMaxRotation:  cfg.Storage.LocalFile.MaxRotatedFiles,
@@ -66,22 +101,11 @@ func initStorage(storageRegion string, cfg *config.Config) error {
 			return fmt.Errorf("new tracing document store (localfile): %w", err)
 		}
 		tracingMetadataStores = append(tracingMetadataStores, localFileStore)
-	}
-
-	if len(tracingMetadataStores) > 0 {
-		tracing.SetTracingStore(
-			tracingMetadataStores,
-			tracing.DocumentOptions{
-				Region: storageRegion,
-			},
-		)
-	}
-	if esStore != nil {
-		tracing.SetTaskStore([]*storage.Store[*tracing.Document]{esStore}, tracing.DocumentOptions{Region: storageRegion})
+		initializedStores = append(initializedStores, localFileStore)
 	}
 
 	if cfg.Storage.Elasticsearch.Enabled() {
-		profileStore, err := storage.NewFromConfig[*tracing.Document](context.Background(), &driver.Config{
+		store, err := newStore(context.Background(), &driver.Config{
 			Driver:      "elasticsearch",
 			ESAddresses: strutil.SplitCommaList(cfg.Storage.Elasticsearch.Address),
 			ESUsername:  cfg.Storage.Elasticsearch.Username,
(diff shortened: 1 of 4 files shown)


cmd/huatuo-bamai/storage.go
```diff
@@ -27,21 +28,54 @@ import (
...
 	if d.opts.DisableStorage {
 		log.Infof("storage backends disabled by --disable-storage")
 		return nil, nil
 	}
 
-	return nil, initStorage(d.opts.Region, config.Get())
+	if err := initStorage(d.opts.Region, config.Get()); err != nil {
+		return nil, err
+	}
+	return tracing.CloseStores, nil
```
huhong-web, line 47: [P1] Keep a single owner for store shutdown. `startTracing` already calls `tracing.CloseStores` in its cleanup, and `Daemon.Run` executes that tracing cleanup before this storage cleanup. With Elasticsearch configured, the second call reaches `esutil.BulkIndexer.Close`, which closes `bi.queue` again and panics (`close of closed channel`) on every normal shutdown. Remove the store close from `startTracing` (and add a daemon lifecycle test that executes both cleanups), or make `CloseStores` genuinely idempotent before registering this second cleanup.

huhong-web (commented):
Reviewed immutable head `fb7b75246ecc` against base `de8dadafdfb0` (merge base `de8dadafdfb0`).

Found one actionable shutdown defect; see the inline comment. Validation included lifecycle/caller tracing through `Daemon.Run`, `startTracing`, `tracing.CloseStores`, and the vendored Elasticsearch bulk indexer, plus `git diff --check`. Package tests could not be rerun because the review environment lacked enough disk space to materialize the immutable worktree.

---
*Co-authored by: gpt-5.6-sol*

cmd/huatuo-bamai/storage.go
tju-yxq: Addressed in `8fd7eb1a219b324769b5b0be41c98de2e8e2afc8`.

The review is correct: registering `tracing.CloseStores` in `setupStorage` while retaining the call in `startTracing` created two shutdown owners. The second Elasticsearch bulk-indexer close could close its queue channel twice.

The revision now:

- removes store closure from `startTracing`; its cleanup only stops and waits for the tracing manager;
- keeps `tracing.CloseStores` exclusively in the storage cleanup;
- orders setup as BPF -> storage -> tracing, which produces tracing -> storage -> BPF cleanup order;
- adds a regression test that executes both tracing and storage cleanups and asserts the backend is closed exactly once;
- asserts the daemon step order preserves collector stop, store drain, then BPF teardown.

Revalidated with package tests, race tests, vet, package lint, full-repository lint, `make check`, and `git diff --check`; all pass.

huhong-web (commented):
Re-reviewed immutable head `8fd7eb1a219b` against merge base `de8dadafdfb0`; the prior review targeted `fb7b75246ecc`.

## Architecture & Design

**Verdict: reasonable.** Storage now owns backend shutdown, tracing owns only collector shutdown, and `daemonSetupSteps` produces the required tracing -> storage -> BPF cleanup order. `initStorageWithFactory` also constructs all stores before publishing the global writers and rolls back partial construction, so the lifecycle has one owner and a failure-atomic publication boundary. That is the smallest coherent fix; making every backend tolerate duplicate close calls would leave ownership ambiguous.

Verification covers shared-store deduplication, all configured stores, rollback, joined errors, context propagation, and cleanup order. Repository search finds one production `CloseStores` registration. No new security issue was found. GitHub has no checks recorded for this updated head, so the focused tests in the patch remain the available automated evidence.

**Difference from ISSUE-333:** the previous P1 double-close finding is fixed; actionable findings changed from 1 to 0.

---
*Co-authored by: gpt-5.6-sol*

PR #641, opened 2026-08-18
feat(mthread): add MooreThreads GPU monitoring metrics

dowzyx:
**Description**
Implement Moore Threads GPU collector via purego-bound MTML C library. Collect static device info (model/brand/serial/PCI/BIOS/MUSA/memory spec) and dynamic metrics (temperature/power/voltage/utilization/clocks/fans/ pstate/VPU/PCIe link/MtLink interconnect) per device. Static info is cached and only rebuilt on re-init. Health, PCIe, and MtLink metric groups are independently toggleable via config.

**Usage**
1. Check Environment: 
    Ensure that the /usr/lib/libmtml.so file exists.
2. Modify Configuration: 
    Edit /opt/huatuo-bamai/conf/huatuo-bamai.conf with the following adjustments:
    2.1 Under the [BlackList] section, ensure that mthreads_gpu is not added to the blacklist.
    2.2 Under the [MetricCollector.Mthreads] section, configure the desired metric groups as needed. Set the value to true to enable monitoring, or false to disable it.
3. Verify Metrics:
    After starting the huatuo-bamai service, access the Prometheus metrics endpoint. You should be able to see mthreads_gpu related metrics, such as huatuo_bamai_mthreads_gpu_device_info.
Note: If you do not see any mthreads_gpu_ related metrics after startup, check the huatuo-bamai logs for the error message: "mthreads: GPU collector disabled".

### DIFF
--- a/core/metrics/mthreads/mtml/device.go
+++ b/core/metrics/mthreads/mtml/device.go
@@ -1,0 +1,709 @@
+// Copyright 2026 The HuaTuo Authors
+// Copyright 2026 The Mthreads Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package mtml
+
+// LibraryVersion returns the MTML library version string.
+func LibraryVersion() (string, error) {
+	if !IsReady() {
+		return "", errNotReady
+	}
+	if mtmlLibraryGetVersion == nil {
+		return "", errNotSupportedSymbol("mtmlLibraryGetVersion")
+	}
+	return getString("mtmlLibraryGetVersion", libmtml.Lib(), libraryVersionBufferSize,
+		func(buf []byte) Return {
+			return mtmlLibraryGetVersion(libmtml.Lib(), &buf[0], uint32(len(buf)))
+		})
+}
+
+// CountDevice returns the number of MTML-visible GPU devices.
+func CountDevice() (uint32, error) {
+	if !IsReady() {
+		return 0, errNotReady
+	}
+	var count uint32
+	if err := checkReturnCode("mtmlLibraryCountDevice", mtmlLibraryCountDevice(libmtml.Lib(), &count)); err != nil {
+		return 0, err
+	}
+	return count, nil
+}
+
+// Device is a live MtmlDevice* handle. Close it with Close when done.
+type Device struct{ handle uintptr }
+
+// InitDeviceByIndex returns the device at the given index. Close it when done.
+func InitDeviceByIndex(index uint32) (*Device, error) {
+	if !IsReady() {
+		return nil, errNotReady
+	}
+	var dev uintptr
+	if err := checkReturnCode("mtmlLibraryInitDeviceByIndex",
+		mtmlLibraryInitDeviceByIndex(libmtml.Lib(), index, &dev)); err != nil {
+		return nil, err
+	}
+	return &Device{handle: dev}, nil
+}
+
+// Close frees the device handle (mtmlLibraryFreeDevice).
+func (d *Device) Close() error {
+	if d.handle == 0 {
+		return nil
+	}
+	err := checkReturnCode("mtmlLibraryFreeDevice", mtmlLibraryFreeDevice(d.handle))
+	if err == nil {
+		d.handle = 0
+	}
+	return err
+}
+
+// Index returns the device index.
+func (d *Device) Index() (uint32, error) {
+	if mtmlDeviceGetIndex == nil {
+		return 0, errNotSupportedSymbol("mtmlDeviceGetIndex")
+	}
+	var idx uint32
+	if err := checkReturnCode("mtmlDeviceGetIndex", mtmlDeviceGetIndex(d.handle, &idx)); err != nil {
+		return 0, err
+	}
+	return idx, nil
+}
+
+// Name returns the device product name.
+func (d *Device) Name() (string, error) {
+	if mtmlDeviceGetName == nil {
+		return "", errNotSupportedSymbol("mtmlDeviceGetName")
+	}
+	return getString("mtmlDeviceGetName", d.handle, deviceNameBufferSize,
+		func(buf []byte) Return {
+			return mtmlDeviceGetName(d.handle, &buf[0], uint32(len(buf)))
+		})
+}
+
+// UUID returns the device UUID.
+func (d *Device) UUID() (string, error) {
+	if mtmlDeviceGetUUID == nil {
+		return "", errNotSupportedSymbol("mtmlDeviceGetUUID")
+	}
+	return getString("mtmlDeviceGetUUID", d.handle, deviceUUIDBufferSize,
+		func(buf []byte) Return {
+			return mtmlDeviceGetUUID(d.handle, &buf[0], uint32(len(buf)))
+		})
+}
+
+func (d *Device) Brand() (Brand, error) {
+	if mtmlDeviceGetBrand == nil {
+		return 0, errNotSupportedSymbol("mtmlDeviceGetBrand")
+	}
+	var brand uint32
+	if err := checkReturnCode("mtmlDeviceGetBrand", mtmlDeviceGetBrand(d.handle, &brand)); err != nil {
+		return 0, err
+	}
+	return Brand(brand), nil
+}
+
+// SerialNumber returns the device serial number.
+// NOTE: MTML's signature is (dev, length, buf) — length before buffer.
+func (d *Device) SerialNumber() (string, error) {
+	if mtmlDeviceGetSerialNumber == nil {
+		return "", errNotSupportedSymbol("mtmlDeviceGetSerialNumber")
+	}
+	return getStringSerialLenFirst("mtmlDeviceGetSerialNumber", d.handle, deviceSerialNumberBufferSize,
+		func(buf []byte) Return {
+			return mtmlDeviceGetSerialNumber(d.handle, uint32(len(buf)), &buf[0])
+		})
+}
+
+// BiosVersion returns the device BIOS version.
+func (d *Device) BiosVersion() (string, error) {
+	if mtmlDeviceGetMtBiosVersion == nil {
+		return "", errNotSupportedSymbol("mtmlDeviceGetMtBiosVersion")
+	}
+	return getString("mtmlDeviceGetMtBiosVersion", d.handle, biosVersionBufferSize,
+		func(buf []byte) Return {
+			return mtmlDeviceGetMtBiosVersion(d.handle, &buf[0], uint32(len(buf)))
+		})
+}
+
+// GpuCores returns the number of GPU cores.
+func (d *Device) GpuCores() (uint32, error) {
+	if mtmlDeviceCountGpuCores == nil {
+		return 0, errNotSupportedSymbol("mtmlDeviceCountGpuCores")
+	}
+	var cores uint32
+	if err := checkReturnCode("mtmlDeviceCountGpuCores", mtmlDeviceCountGpuCores(d.handle, &cores)); err != nil {
+		return 0, err
+	}
+	return cores, nil
+}
+
+// PciSbdf returns the PCI SBDF string (e.g. "0000:01:00.0").
+func (d *Device) PciSbdf() (string, error) {
+	if mtmlDeviceGetPciInfo == nil {
+		return "", errNotSupportedSymbol("mtmlDeviceGetPciInfo")
+	}
+	var info mtmlPciInfo
+	if err := checkReturnCode("mtmlDeviceGetPciInfo", mtmlDeviceGetPciInfo(d.handle, &info)); err != nil {
+		return "", err
+	}
+	return cString(info.sbdf[:]), nil
+}
+
+// PciInfo returns the parsed PCIe link info (current/max speed in GT/s and width in lanes).
+func (d *Device) PciInfo() (*PciLinkInfo, error) {
+	if mtmlDeviceGetPciInfo == nil {
+		return nil, errNotSupportedSymbol("mtmlDeviceGetPciInfo")
+	}
+	var info mtmlPciInfo
+	if err := checkReturnCode("mtmlDeviceGetPciInfo", mtmlDeviceGetPciInfo(d.handle, &info)); err != nil {
+		return nil, err
+	}
+	return &PciLinkInfo{
+		CurSpeed: float64(info.pciCurSpeed),
+		MaxSpeed: float64(info.pciMaxSpeed),
+		CurWidth: info.pciCurWidth,
+		MaxWidth: info.pciMaxWidth,
+	}, nil
+}
+
+// PcieReplayCounter returns the PCIe replay (retransmission) counter.
+func (d *Device) PcieReplayCounter() (uint32, error) {
+	if mtmlDeviceGetPcieReplayCounter == nil {
+		return 0, errNotSupportedSymbol("mtmlDeviceGetPcieReplayCounter")
+	}
+	var count uint32
+	if err := checkReturnCode("mtmlDeviceGetPcieReplayCounter",
+		mtmlDeviceGetPcieReplayCounter(d.handle, &count)); err != nil {
+		return 0, err
+	}
+	return count, nil
+}
+
+// PciLinkInfo holds PCIe link speed (GT/s) and width (lanes).
+type PciLinkInfo struct {
+	CurSpeed float64 // current link speed, GT/s
+	MaxSpeed float64 // max link speed, GT/s
+	CurWidth uint32  // current link width, lanes
+	MaxWidth uint32  // max link width, lanes
+}
+
+// PowerUsage returns the device power consumption in milliwatts.
+func (d *Device) PowerUsage() (uint32, error) {
+	if mtmlDeviceGetPowerUsage == nil {
+		return 0, errNotSupportedSymbol("mtmlDeviceGetPowerUsage")
+	}
+	var power uint32
+	if err := checkReturnCode("mtmlDeviceGetPowerUsage", mtmlDeviceGetPowerUsage(d.handle, &power)); err != nil {
+		return 0, err
+	}
+	return power, nil
+}
+
+// PerformanceState returns the device performance state (P-state) as a numeric
+// value (0 = P0, etc.).
+func (d *Device) PerformanceState() (int32, error) {
+	if mtmlDeviceGetPerformanceState == nil {
+		return 0, errNotSupportedSymbol("mtmlDeviceGetPerformanceState")
+	}
+	var pstate int32
+	if err := checkReturnCode("mtmlDeviceGetPerformanceState",
+		mtmlDeviceGetPerformanceState(d.handle, &pstate)); err != nil {
+		return 0, err
+	}
+	return pstate, nil
+}
+
+// MusaComputeCapability returns the MUSA compute capability major/minor version
+// (e.g. major=3, minor=0).
+func (d *Device) MusaComputeCapability() (major, minor int32, err error) {
+	if mtmlDeviceGetMusaComputeCapability == nil {
+		return 0, 0, errNotSupportedSymbol("mtmlDeviceGetMusaComputeCapability")
+	}
+	if err := checkReturnCode("mtmlDeviceGetMusaComputeCapability",
+		mtmlDeviceGetMusaComputeCapability(d.handle, &major, &minor)); err != nil {
+		return 0, 0, err
+	}
+	return major, minor, nil
+}
+
+// FanCount returns the number of fans on the device.
+func (d *Device) FanCount() (uint32, error) {
+	if mtmlDeviceCountFan == nil {
+		return 0, errNotSupportedSymbol("mtmlDeviceCountFan")
+	}
+	var count uint32
+	if err := checkReturnCode("mtmlDeviceCountFan", mtmlDeviceCountFan(d.handle, &count)); err != nil {
+		return 0, err
+	}
+	return count, nil
+}
+
+// FanSpeed returns the fan speed as a percentage of the max noise-tolerance
+// speed (may exceed 100%).
+func (d *Device) FanSpeed(fanIndex uint32) (uint32, error) {
+	if mtmlDeviceGetFanSpeed == nil {
+		return 0, errNotSupportedSymbol("mtmlDeviceGetFanSpeed")
+	}
+	var speed uint32
+	if err := checkReturnCode("mtmlDeviceGetFanSpeed", mtmlDeviceGetFanSpeed(d.handle, fanIndex, &speed)); err != nil {
+		return 0, err
+	}
+	return speed, nil
+}
+
+// FanRpm returns the fan speed in RPM.
+func (d *Device) FanRpm(fanIndex uint32) (uint32, error) {
+	if mtmlDeviceGetFanRpm == nil {
+		return 0, errNotSupportedSymbol("mtmlDeviceGetFanRpm")
+	}
+	var rpm uint32
+	if err := checkReturnCode("mtmlDeviceGetFanRpm", mtmlDeviceGetFanRpm(d.handle, fanIndex, &rpm)); err != nil {
+		return 0, err
+	}
+	return rpm, nil
+}
+
+// Gpu is a live MtmlGpu* handle.
+type Gpu struct{ handle uintptr }
+
+// InitGpu returns the Gpu sub-object of the device. Close it when done.
+func (d *Device) InitGpu() (*Gpu, error) {
+	if mtmlDeviceInitGpu == nil {
+		return nil, errNotSupportedSymbol("mtmlDeviceInitGpu")
+	}
+	var gpu uintptr
+	if err := checkReturnCode("mtmlDeviceInitGpu", mtmlDeviceInitGpu(d.handle, &gpu)); err != nil {
+		return nil, err
+	}
+	return &Gpu{handle: gpu}, nil
+}
+
+// Close frees the gpu handle.
+func (g *Gpu) Close() error {
+	if g.handle == 0 {
+		return nil
+	}
+	if mtmlDeviceFreeGpu == nil {
+		g.handle = 0
+		return errNotSupportedSymbol("mtmlDeviceFreeGpu")
+	}
+	err := checkReturnCode("mtmlDeviceFreeGpu", mtmlDeviceFreeGpu(g.handle))
+	g.handle = 0
+	return err
+}
+
+// Temperature returns the GPU temperature in degrees Celsius.
+func (g *Gpu) Temperature() (int32, error) {
+	if mtmlGpuGetTemperature == nil {
+		return 0, errNotSupportedSymbol("mtmlGpuGetTemperature")
+	}
+	var temp int32
+	if err := checkReturnCode("mtmlGpuGetTemperature", mtmlGpuGetTemperature(g.handle, &temp)); err != nil {
+		return 0, err
+	}
+	return temp, nil
+}
+
+// Utilization returns the GPU utilization as a percentage (0-100).
+func (g *Gpu) Utilization() (uint32, error) {
+	if mtmlGpuGetUtilization == nil {
+		return 0, errNotSupportedSymbol("mtmlGpuGetUtilization")
+	}
+	var util uint32
+	if err := checkReturnCode("mtmlGpuGetUtilization", mtmlGpuGetUtilization(g.handle, &util)); err != nil {
+		return 0, err
+	}
+	return util, nil
+}
+
+// Clock returns the GPU clock frequency in MHz.
+func (g *Gpu) Clock() (uint32, error) {
+	if mtmlGpuGetClock == nil {
+		return 0, errNotSupportedSymbol("mtmlGpuGetClock")
+	}
+	var clock uint32
+	if err := checkReturnCode("mtmlGpuGetClock", mtmlGpuGetClock(g.handle, &clock)); err != nil {
+		return 0, err
+	}
+	return clock, nil
+}
+
+// MaxClock returns the GPU maximum clock frequency in MHz.
+func (g *Gpu) MaxClock() (uint32, error) {
+	if mtmlGpuGetMaxClock == nil {
+		return 0, errNotSupportedSymbol("mtmlGpuGetMaxClock")
+	}
+	var clock uint32
+	if err := checkReturnCode("mtmlGpuGetMaxClock", mtmlGpuGetMaxClock(g.handle, &clock)); err != nil {
+		return 0, err
+	}
+	return clock, nil
+}
+
+// Voltage returns the GPU voltage in millivolts.
+func (g *Gpu) Voltage() (int32, error) {
+	if mtmlGpuGetVoltage == nil {
+		return 0, errNotSupportedSymbol("mtmlGpuGetVoltage")
+	}
+	var v int32
+	if err := checkReturnCode("mtmlGpuGetVoltage", mtmlGpuGetVoltage(g.handle, &v)); err != nil {
+		return 0, err
+	}
+	return v, nil
+}
+
+// EnforcedPowerLimit returns the enforced GPU power limit in milliwatts.
+func (g *Gpu) EnforcedPowerLimit() (int32, error) {
+	if mtmlGpuGetEnforcedPowerLimit == nil {
+		return 0, errNotSupportedSymbol("mtmlGpuGetEnforcedPowerLimit")
+	}
+	var limit int32
+	if err := checkReturnCode("mtmlGpuGetEnforcedPowerLimit",
+		mtmlGpuGetEnforcedPowerLimit(g.handle, &limit)); err != nil {
+		return 0, err
+	}
+	return limit, nil
+}
+
+// PowerManagementDefaultLimit returns the default GPU power management limit in milliwatts.
+func (g *Gpu) PowerManagementDefaultLimit() (int32, error) {
+	if mtmlGpuGetPowerManagementDefaultLimit == nil {
+		return 0, errNotSupportedSymbol("mtmlGpuGetPowerManagementDefaultLimit")
+	}
+	var limit int32
+	if err := checkReturnCode("mtmlGpuGetPowerManagementDefaultLimit",
+		mtmlGpuGetPowerManagementDefaultLimit(g.handle, &limit)); err != nil {
+		return 0, err
+	}
+	return limit, nil
+}
+
+// TemperatureThreshold returns a GPU temperature threshold in degrees Celsius.
+// thresholdType is a TempThreshold* constant (0=shutdown, 1=slowdown).
+func (g *Gpu) TemperatureThreshold(thresholdType int32) (int32, error) {
+	if mtmlGpuGetTemperatureThreshold == nil {
+		return 0, errNotSupportedSymbol("mtmlGpuGetTemperatureThreshold")
+	}
+	var temp int32
+	if err := checkReturnCode("mtmlGpuGetTemperatureThreshold",
+		mtmlGpuGetTemperatureThreshold(g.handle, thresholdType, &temp)); err != nil {
+		return 0, err
+	}
+	return temp, nil
+}
+
+// Memory is a live MtmlMemory* handle.
+type Memory struct{ handle uintptr }
+
+// InitMemory returns the Memory sub-object of the device. Close it when done.
+func (d *Device) InitMemory() (*Memory, error) {
+	if mtmlDeviceInitMemory == nil {
+		return nil, errNotSupportedSymbol("mtmlDeviceInitMemory")
+	}
+	var mem uintptr
+	if err := checkReturnCode("mtmlDeviceInitMemory", mtmlDeviceInitMemory(d.handle, &mem)); err != nil {
+		return nil, err
+	}
+	return &Memory{handle: mem}, nil
+}
+
+// Close frees the memory handle.
+func (m *Memory) Close() error {
+	if m.handle == 0 {
+		return nil
+	}
+	if mtmlDeviceFreeMemory == nil {
+		m.handle = 0
+		return errNotSupportedSymbol("mtmlDeviceFreeMemory")
+	}
+	err := checkReturnCode("mtmlDeviceFreeMemory", mtmlDeviceFreeMemory(m.handle))
+	m.handle = 0
+	return err
+}
+
+// Total returns the total VRAM in bytes.
+func (m *Memory) Total() (uint64, error) {
+	if mtmlMemoryGetTotal == nil {
+		return 0, errNotSupportedSymbol("mtmlMemoryGetTotal")
+	}
+	var total uint64
+	if err := checkReturnCode("mtmlMemoryGetTotal", mtmlMemoryGetTotal(m.handle, &total)); err != nil {
+		return 0, err
+	}
+	return total, nil
+}
+
+// Used returns the used VRAM in bytes.
+func (m *Memory) Used() (uint64, error) {
+	if mtmlMemoryGetUsed == nil {
+		return 0, errNotSupportedSymbol("mtmlMemoryGetUsed")
+	}
+	var used uint64
+	if err := checkReturnCode("mtmlMemoryGetUsed", mtmlMemoryGetUsed(m.handle, &used)); err != nil {
+		return 0, err
+	}
+	return used, nil
+}
+
+// Utilization returns the memory utilization as a percentage (0-100).
+func (m *Memory) Utilization() (uint32, error) {
+	if mtmlMemoryGetUtilization == nil {
+		return 0, errNotSupportedSymbol("mtmlMemoryGetUtilization")
+	}
+	var util uint32
+	if err := checkReturnCode("mtmlMemoryGetUtilization", mtmlMemoryGetUtilization(m.handle, &util)); err != nil {
+		return 0, err
+	}
+	return util, nil
+}
+
+// Temperature returns the memory temperature in degrees Celsius.
+func (m *Memory) Temperature() (int32, error) {
+	if mtmlMemoryGetTemperature == nil {
+		return 0, errNotSupportedSymbol("mtmlMemoryGetTemperature")
+	}
+	var temp int32
+	if err := checkReturnCode("mtmlMemoryGetTemperature", mtmlMemoryGetTemperature(m.handle, &temp)); err != nil {
+		return 0, err
+	}
+	return temp, nil
+}
+
+// Clock returns the memory clock frequency in MHz.
+func (m *Memory) Clock() (uint32, error) {
+	if mtmlMemoryGetClock == nil {
+		return 0, errNotSupportedSymbol("mtmlMemoryGetClock")
+	}
+	var clock uint32
+	if err := checkReturnCode("mtmlMemoryGetClock", mtmlMemoryGetClock(m.handle, &clock)); err != nil {
+		return 0, err
+	}
+	return clock, nil
+}
+
+// MaxClock returns the memory maximum clock frequency in MHz.
+func (m *Memory) MaxClock() (uint32, error) {
+	if mtmlMemoryGetMaxClock == nil {
+		return 0, errNotSupportedSymbol("mtmlMemoryGetMaxClock")
+	}
+	var clock uint32
+	if err := checkReturnCode("mtmlMemoryGetMaxClock", mtmlMemoryGetMaxClock(m.handle, &clock)); err != nil {
+		return 0, err
+	}
+	return clock, nil
+}
+
+// BusWidth returns the memory bus width in bits.
+func (m *Memory) BusWidth() (uint32, error) {
+	if mtmlMemoryGetBusWidth == nil {
+		return 0, errNotSupportedSymbol("mtmlMemoryGetBusWidth")
+	}
+	var width uint32
+	if err := checkReturnCode("mtmlMemoryGetBusWidth", mtmlMemoryGetBusWidth(m.handle, &width)); err != nil {
+		return 0, err
+	}
+	return width, nil
+}
+
+// Bandwidth returns the memory bandwidth in GB/s.
+func (m *Memory) Bandwidth() (uint32, error) {
+	if mtmlMemoryGetBandwidth == nil {
+		return 0, errNotSupportedSymbol("mtmlMemoryGetBandwidth")
+	}
+	var bw uint32
+	if err := checkReturnCode("mtmlMemoryGetBandwidth", mtmlMemoryGetBandwidth(m.handle, &bw)); err != nil {
+		return 0, err
+	}
+	return bw, nil
+}
+
+// Speed returns the memory speed in Mbps.
+func (m *Memory) Speed() (uint32, error) {
+	if mtmlMemoryGetSpeed == nil {
+		return 0, errNotSupportedSymbol("mtmlMemoryGetSpeed")
+	}
+	var speed uint32
+	if err := checkReturnCode("mtmlMemoryGetSpeed", mtmlMemoryGetSpeed(m.handle, &speed)); err != nil {
+		return 0, err
+	}
+	return speed, nil
+}
+
+// MemoryType returns the memory type code (0 = LPDDR4, 1 = GDDR6).
+func (m *Memory) MemoryType() (uint32, error) {
+	if mtmlMemoryGetType == nil {
+		return 0, errNotSupportedSymbol("mtmlMemoryGetType")
+	}
+	var typ uint32
+	if err := checkReturnCode("mtmlMemoryGetType", mtmlMemoryGetType(m.handle, &typ)); err != nil {
+		return 0, err
+	}
+	return typ, nil
+}
+
+// Vendor returns the memory vendor string.
+// NOTE: MTML's signature is (mem, length, buf) — length before buffer.
+func (m *Memory) Vendor() (string, error) {
+	if mtmlMemoryGetVendor == nil {
+		return "", errNotSupportedSymbol("mtmlMemoryGetVendor")
+	}
+	return getStringSerialLenFirst("mtmlMemoryGetVendor", m.handle, memoryVendorBufferSize,
+		func(buf []byte) Return {
+			return mtmlMemoryGetVendor(m.handle, uint32(len(buf)), &buf[0])
+		})
+}
+
+// Vpu is a live MtmlVpu* handle.
+type Vpu struct{ handle uintptr }
+
+// InitVpu returns the VPU sub-object of the device. Close it when done.
+func (d *Device) InitVpu() (*Vpu, error) {
+	if mtmlDeviceInitVpu == nil {
+		return nil, errNotSupportedSymbol("mtmlDeviceInitVpu")
+	}
+	var vpu uintptr
+	if err := checkReturnCode("mtmlDeviceInitVpu", mtmlDeviceInitVpu(d.handle, &vpu)); err != nil {
+		return nil, err
+	}
+	return &Vpu{handle: vpu}, nil
+}
+
+// Close frees the vpu handle.
+func (v *Vpu) Close() error {
+	if v.handle == 0 {
+		return nil
+	}
+	if mtmlDeviceFreeVpu == nil {
+		v.handle = 0
+		return errNotSupportedSymbol("mtmlDeviceFreeVpu")
+	}
+	err := checkReturnCode("mtmlDeviceFreeVpu", mtmlDeviceFreeVpu(v.handle))
+	v.handle = 0
+	return err
+}
+
+// CodecUtil returns the VPU (codec) utilization. encUtil/decUtil are the
+// encoder/decoder utilization rates as percentages.
+type CodecUtil struct {
+	Util    uint32 // overall utilization, %
+	Period  uint32 // sampling period, microseconds
+	EncUtil uint32 // encoder utilization, %
+	DecUtil uint32 // decoder utilization, %
+}
+
+// Utilization returns the VPU codec utilization.
+func (v *Vpu) Utilization() (*CodecUtil, error) {
+	if mtmlVpuGetUtilization == nil {
+		return nil, errNotSupportedSymbol("mtmlVpuGetUtilization")
+	}
+	var raw mtmlCodecUtil
+	if err := checkReturnCode("mtmlVpuGetUtilization", mtmlVpuGetUtilization(v.handle, &raw)); err != nil {
+		return nil, err
+	}
+	return &CodecUtil{
+		Util:    raw.util,
+		Period:  raw.period,
+		EncUtil: raw.encUtil,
+		DecUtil: raw.decUtil,
+	}, nil
+}
+
+// Clock returns the VPU clock frequency in MHz.
+func (v *Vpu) Clock() (uint32, error) {
+	if mtmlVpuGetClock == nil {
+		return 0, errNotSupportedSymbol("mtmlVpuGetClock")
+	}
+	var clock uint32
+	if err := checkReturnCode("mtmlVpuGetClock", mtmlVpuGetClock(v.handle, &clock)); err != nil {
+		return 0, err
+	}
+	return clock, nil
+}
+
+// MaxClock returns the VPU maximum clock frequency in MHz.
+func (v *Vpu) MaxClock() (uint32, error) {
+	if mtmlVpuGetMaxClock == nil {
+		return 0, errNotSupportedSymbol("mtmlVpuGetMaxClock")
+	}
+	var clock uint32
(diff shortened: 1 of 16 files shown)


core/metrics/mthreads_gpu.go
```diff
@@ -0,0 +1,552 @@
...
+func collectMthreadsDeviceMetrics(ctx context.Context, gpu uint32) []*metric.Data {
+	dev, err := mtml.InitDeviceByIndex(gpu)
+	if err != nil {
+		return nil
+	}
+	defer dev.Close()
+
+	var out []*metric.Data
+
+	if cfg.Mthreads.EnableHealth {
```
huhong-web, line 449: [P1] This file does not compile: `cfg` is not declared in this function, file, or package (the same undefined identifier appears at lines 236, 315, and 323). An isolated compiler check reports all four sites. Take one `cfg := configSnapshot()` at the start of a scrape and pass that immutable snapshot through the helpers so a live update cannot split one scrape across configurations. Acceptance: `go test ./core/metrics/...` and `make check` compile successfully.

core/metrics/mthreads/mtml/lib.go
```diff
@@ -0,0 +1,179 @@
...
+
+	if l.refcount > 0 {
+		return nil
+	}
+
+	if err := l.dl.Open(); err != nil {
+		return err
+	}
+
+	l.registerSymbols(l.dl.Handle())
```
huhong-web: [P1] A driver library missing any one of these symbols will panic the entire `huatuo-bamai` process here. `registerSymbols` uses `purego.RegisterLibFunc`, whose implementation panics when `Dlsym` fails (`vendor/github.com/ebitengine/purego/func.go:24-30`); all optional and disassembly-inferred symbols are registered even when their metric groups are disabled. Resolve with the error-returning `purego.Dlsym` plus `RegisterFunc`, distinguish required symbols from optional capabilities, map absent optional symbols to `NotSupported`, and close the handle on any required-symbol failure. Please cover both cases with an intentionally incomplete fake shared library.

core/metrics/mthreads_gpu.go
```diff
@@ -0,0 +1,552 @@
...
+	// Per-device static info.
+	for _, gpu := range cache.devices {
+		if ctx.Err() != nil {
+			log.Debugf("mthreads: scrape timeout exceeded before static info for device %d, skipping", gpu)
+			break
+		}
+		cache.staticInfo = append(cache.staticInfo, collectMthreadsDeviceInfo(ctx, gpu)...)
+	}
+
+	c.cache.Store(cache)
```
huhong-web: [P2] Do not publish this cache after a partial static-info pass. The loop above breaks when the deadline expires, while `collectMthreadsDeviceInfo` converts device-init and individual field failures into empty/partial slices; this cache is then reused until an unrelated MTML re-init. A slow or transiently failing first scrape therefore permanently suppresses metrics for the process lifetime. Store only after every device's static pass completes, or represent incomplete entries and retry them on later scrapes.

core/metrics/mthreads_gpu.go
```diff
@@ -0,0 +1,552 @@
...
+	if sbdf, err := dev.PciSbdf(); err == nil {
+		out = append(out, metric.NewGaugeData("device_pci_info", 1, "GPU PCI info.", map[string]string{
+			"gpu":  gpuLabel,
+			"sbdf": sbdf,
+		}))
+	}
+
+	// pcie max link capability (static: hardware max speed/width, cached).
+	// The negotiated cur values are collected dynamically on each scrape.
+	if cfg.Mthreads.EnablePCIe {
```
huhong-web: [P2] Config-gated output cannot be baked into this one-time cache. Runtime configuration is hot-published by `config.publishConfig -> collector.Set`, but this cache is not invalidated on a config update. If PCIe starts disabled and is later enabled, these max metrics never appear; if it starts enabled and is later disabled, they remain exported. Cache only config-independent facts and apply the gate while assembling each scrape, or key/invalidate the cache by configuration generation.

core/metrics/mthreads/mtml/lib.go
```diff
@@ -0,0 +1,179 @@
...
+func newLibrary() *library {
+	return &library{
+		dl: dl.New(defaultMtmlLibraryPath(), purego.RTLD_NOW|purego.RTLD_GLOBAL),
+	}
+}
+
+func defaultMtmlLibraryPath() string {
+	switch runtime.GOOS {
+	case "linux":
+		return "/usr/lib/libmtml.so"
```
huhong-web: [P2] Requiring exactly `/usr/lib/libmtml.so` rejects installations that expose only the versioned SONAME or place the library on the configured dynamic-loader path. The official Moore Threads binding tries `libmtml.so.2` and then `libmtml.so` through the system loader: https://github.com/MooreThreads/mthreads-ml-py/blob/eb91b67641134ef65377225c58f7247e0fa41eeb/pymtml.py#L927-L950. Prefer that discovery order and report every attempted name. If an exact override is needed, make it a per-backend, startup-only `MetricCollector.Mthreads.LibraryPath` with SONAME discovery as the empty/default behavior; do not use one global path for all GPU vendors.

huatuo-bamai.conf
```diff
@@ -3,7 +3,7 @@
 # - BlackList
 # Global blacklist for tracing and metrics.
 #
-BlackList = ["netdev_hw", "netdev_qdisc", "metax_gpu", "ascend_npu", "diskio", "tcp_retransmit"]
+BlackList = ["netdev_hw", "netdev_qdisc", "metax_gpu", "ascend_npu", "diskio", "tcp_retransmit", "mthreads_gpu"]
```
huhong-web, line 6: [P2] Please update the canonical user contract together with this sample. `docs/configuration/huatuo-bamai-configuration_{en,zh}.md` still documents the old default blacklist, and `docs/key-feature/kernel-wide-insight_{en,zh}.md` still lists only MetaX in the GPU metrics section. Add MThreads enablement, library discovery/minimum-version expectations, toggle semantics, and the emitted metric inventory; document restart semantics if a path override is introduced.

huhong-web (changes requested):
## Code Review

Requesting changes on `de8dadaf..37a388ae`: 6 actionable findings are attached inline (2 P1, 4 P2). The current head does not compile, and the native-library compatibility and cache/config contracts need to be made failure-safe before merge.

## Architecture & Design

**Verdict: needs changes.**

The intended dependency direction is sensible: `core/metrics/mthreads_gpu.go` owns metric policy, `mthreads/mtml` owns the vendor API, and `mthreads/dl` owns dynamic loading. The `dynamicLibrary` interface and library mutex also provide reasonable seams for lifecycle isolation. However, the actual startup path is `newMthreadsGpuCollector -> mtml.Init -> library.load -> registerSymbols`, where every public and undocumented ABI symbol is eagerly required and a missing symbol panics instead of returning through the collector's `ErrNotSupported` contract. That makes one optional hardware backend capable of terminating the whole daemon.

The other cross-layer break is configuration. Runtime updates are atomically published by `cmd/huatuo-bamai/config.publishConfig -> collector.Set`, but MThreads reads feature switches from helpers whose results are stored in a one-time static cache. The cache therefore cannot faithfully implement live enable/disable semantics, and a timeout or transient device error during its first build is persisted indefinitely.

**`.so` discovery decision:** MetaX and Ascend currently hard-code vendor-specific absolute paths (`/opt/mxdriver/lib/libmxsml.so` and `/usr/local/dcmi/libdcmi.so`), and there is no shared GPU-library-path setting. That precedent should not force MTML to require `/usr/lib/libmtml.so`: the official Moore Threads binding resolves `libmtml.so.2` first and falls back to `libmtml.so` through the system loader ([source](https://github.com/MooreThreads/mthreads-ml-py/blob/eb91b67641134ef65377225c58f7247e0fa41eeb/pymtml.py#L927-L950)). The smallest robust design is the same SONAME search with clear attempted-name diagnostics. If deployments also need an exact mounted path, add a per-backend, startup-only `MetricCollector.Mthreads.LibraryPath` override whose empty default retains SONAME discovery; a single global GPU path would not model different vendors. Because `libmtml` is currently constructed at package initialization, such an override should be injected when the collector/client is constructed rather than read into mutable global state. Consolidating the three near-identical dynamic-loader packages can be a follow-up and is not required for this fix.

The configuration and observable metric contract is also only partially updated: the sample config changed, but the canonical English/Chinese configuration docs and GPU metric inventory still omit MThreads. No unit, contract, or hardware integration tests accompany the 1,763-line native ABI addition. At minimum, add config-snapshot/cache tests, required-vs-optional symbol tests using a deliberately incomplete fake `.so`, lifecycle cleanup tests, and one MTML driver/GPU integration matrix entry.

## Other checks

- New Go files contain the Apache 2.0 header; `gofmt` and `git diff --check` are clean.
- `go test ./core/metrics/mthreads/...` passes, but both packages report `[no test files]`.
- An isolated compiler check reports `undefined: cfg` at lines 236, 307, 315, and 323. A full `go test ./core/metrics/...` could not finish in this workspace after generated BPF prerequisites because the filesystem ran out of space; the source-level compile blocker is independently reproducible.
- GitHub exposes no status checks for this head.
- Security review found no new SQL/command injection, secret, path-traversal, authentication, or cryptography issue. The native symbol-resolution availability risk is covered inline.

---
*Co-authored by: gpt-5.6-sol*

core/metrics/mthreads_gpu.go
dowzyx: 已增加 scrape 入口取一次cfg快照的处理

core/metrics/mthreads_gpu.go
dowzyx: buildCache修改为如果超时即不store缓存，那么下次Prometheus scrape时，会重试直到所有device的静态信息全部采集全，缓存才会store。

core/metrics/mthreads/mtml/lib.go
dowzyx: 已修改，修改点包括：1、将  purego.RegisterLibFunc 调用替换为 purego.Dlsym + purego.RegisterFunc，对于必要符号缺失,加载返回错误并关闭句柄，对于可选符号缺失，会保持其函数指针为nil，后续数据采集遇到nil会跳过这些指标；2、上层的 mthreads_gpu.go采集指标时，为所有使用可选符号的方法添加 nil 检查

core/metrics/mthreads_gpu.go
dowzyx: 修改为静态数据固定采集，collect中根据配置决定是否上报

huatuo-bamai.conf
dowzyx: 已经在docs/configuration/huatuo-bamai-configuration_{zh|en}.md文档中补充说明了Mthreads配置介绍；在docs/key-feature/kernel-wide-insight_{zh|en}.md文档中补充了mthreads gpu的指标清单

core/metrics/mthreads/mtml/lib.go
dowzyx: 已修改，将硬编码的单一路径修改为多个候选名字，load时按顺序尝试动态加载

core/metrics/mthreads/mtml/lib.go
```diff
@@ -0,0 +1,336 @@
...
+		if err := l.registerSymbols(d.Handle()); err != nil {
+			_ = d.Close()
+			errs = append(errs, fmt.Sprintf("%s: registerSymbols: %v", name, err))
+			continue
+		}
+
+		// mtmlLibraryInit(MtmlLibrary** lib)
+		var handle uintptr
+		if err := checkReturnCode("mtmlLibraryInit", mtmlLibraryInit(&handle)); err != nil {
+			_ = d.Close()
```
huhong-web: [P1] Format this MTML error before unloading the handle. `checkReturnCode` returns `*mtml.Error`; when line 106 formats it, `Error()` calls `Return.String()`, which calls the package-global `mtmlErrorString` callback registered from this handle. Because `d.Close()` has already unloaded that code, an ordinary `mtmlLibraryInit` failure can SIGSEGV the daemon while building its diagnostic. The same global callbacks survive candidate fallback and `Shutdown`. Keep the `.so` loaded while callbacks exist, or store callbacks in a per-open function table and retire/clear it before `dlclose`. Add a fake-library test whose init returns an error and a fallback with a different symbol set.

core/metrics/mthreads/mtml/device.go
```diff
@@ -0,0 +1,721 @@
...
+	return &Gpu{handle: gpu}, nil
+}
+
+// Close frees the gpu handle.
+func (g *Gpu) Close() error {
+	if g.handle == 0 {
+		return nil
+	}
+	if mtmlDeviceFreeGpu == nil {
+		g.handle = 0
```
huhong-web, line 299: [P1] This drops a live native handle without freeing it. `mtmlDeviceInitGpu` and `mtmlDeviceFreeGpu` are resolved independently, so a library that exports the constructor but not the destructor reaches this branch on every scrape; setting `g.handle = 0` makes the leak permanent. `Memory` and `Vpu` have the same failure mode. Treat each init/free pair as one capability during registration (or require both), and test each incomplete pair with a fake `.so`.

core/metrics/mthreads/mtml/types.go
```diff
@@ -0,0 +1,243 @@
...
+	mtmlDeviceGetMtBiosVersion func(uintptr, *byte, uint32) Return
+	// mtmlDeviceCountGpuCores(MtmlDevice*, unsigned int* numCores)
+	mtmlDeviceCountGpuCores func(uintptr, *uint32) Return
+	// mtmlDeviceGetPciInfo(MtmlDevice*, MtmlPciInfo* pci)
+	mtmlDeviceGetPciInfo func(uintptr, *mtmlPciInfo) Return
+	// mtmlDeviceGetPowerUsage(MtmlDevice*, unsigned int* power) — milliwatts
+	mtmlDeviceGetPowerUsage func(uintptr, *uint32) Return
+	// mtmlDeviceGetPcieReplayCounter(MtmlDevice*, unsigned int* replayCounter)
+	// Undocumented in mtml.h; signature inferred from disassembly (2 args).
+	mtmlDeviceGetPcieReplayCounter func(uintptr, *uint32) Return
```
huhong-web, line 107: [P1] Do not invoke a disassembly-inferred ABI from the long-lived agent without a supported version contract. This function and eight other inferred symbols are absent from the vendor's public MTML 2.2 header; `Dlsym` validates only the symbol name, not the parameter layout. A present symbol with a different private signature will corrupt registers/memory or SIGSEGV during a normal scrape, and this PR already disables two related calls after observing process crashes. Remove inferred metrics for now, or gate them to a vendor-supported MTML version and prove every call in a real driver/GPU integration test. Header: https://github.com/MooreThreads/mthreads-ml-py/blob/144369f45c29db693e6469dec0e327e25be533e0/mtml_2.2.0.h

core/metrics/mthreads_gpu.go
```diff
@@ -0,0 +1,553 @@
...
+// collectMthreadsDeviceMetrics collects dynamic metrics for a single device.
+// It opens the device once and dispatches to per-category collectors (health /
+// PCIe / MtLink), each of which manages its own sub-handles. All groups are
+// best-effort: individual NotSupported errors are logged at debug level and do
+// not abort the scrape. The EnableHealth/EnablePCIe/EnableMTLink switches are
+// independent (P0-4).
+func collectMthreadsDeviceMetrics(ctx context.Context, cfg *Config, gpu uint32) []*metric.Data {
+	dev, err := mtml.InitDeviceByIndex(gpu)
+	if err != nil {
+		return nil
```
huhong-web: [P1] Propagate non-`NotSupported` MTML errors instead of converting them to an empty successful result. This discard, plus the same pattern in the category helpers, makes `Update`'s `errors.As(..., *mtml.Error)` shutdown/re-init path unreachable for most driver failures. After a device reset, metrics can silently disappear while `huatuo_bamai_scrape_collector_success` remains 1 until the driver recovers externally or the process restarts. Return `(metrics, error)`, suppress only `mtml.IsNotSupported`, and propagate operational errors with any partial metrics; test that a driver failure triggers one recovery attempt and a failed scrape.

core/metrics/mthreads_gpu.go
```diff
@@ -0,0 +1,553 @@
...
+	// Per-device static info.
+	for _, gpu := range cache.devices {
+		if ctx.Err() != nil {
+			log.Debugf("mthreads: scrape timeout exceeded before static info for device %d, skipping", gpu)
+			break
+		}
+		cache.staticInfo = append(cache.staticInfo, collectMthreadsDeviceInfo(ctx, cfg, gpu)...)
+	}
+
+	c.cache.Store(cache)
```
huhong-web: [P2] This still publishes an incomplete cache. The loop above breaks when the deadline expires, while `collectMthreadsDeviceInfo` turns a transient device-init or field failure into a partial/nil slice; this `Store` then makes those missing static metrics permanent until an unrelated MTML re-init. Return without storing on timeout and represent per-device completion so failed entries are retried. A deterministic test should fail/time out the first build and verify that the next scrape rebuilds the missing device data.

core/metrics/mthreads_gpu.go
```diff
@@ -0,0 +1,553 @@
...
+	if sbdf, err := dev.PciSbdf(); err == nil {
+		out = append(out, metric.NewGaugeData("device_pci_info", 1, "GPU PCI info.", map[string]string{
+			"gpu":  gpuLabel,
+			"sbdf": sbdf,
+		}))
+	}
+
+	// pcie max link capability (static: hardware max speed/width, cached).
+	// The negotiated cur values are collected dynamically on each scrape.
+	if cfg.Mthreads.EnablePCIe {
```
huhong-web: [P2] PCIe max metrics are still baked into the one-time cache using the first scrape's config. `publishConfig -> collector.Set` hot-publishes snapshots and `collect` now reads one per scrape, so `false -> true` never adds these max metrics and `true -> false` leaves them exported even though current PCIe metrics disappear. Cache `PciInfo` max capability independently of config and apply the gate when assembling each scrape (or invalidate by config generation). Cover `false -> true -> false`; the new restart-only documentation does not match the hot behavior of the other two switches.

docs/key-feature/kernel-wide-insight_en.md
```diff
@@ -1254,3 +1254,47 @@ huatuo_bamai_hungtask_total{host="hostname",region="dev"} 0
...
+
+- Moore Threads (MThreads)
+
+MThreads GPU monitoring via the MTML (Moore Threads Management Library). The library is discovered automatically at startup using SONAME search (`libmtml.so.2` then `libmtml.so`) through the system dynamic linker. Static device info is cached and only rebuilt on re-init; dynamic metrics are collected on every scrape.
+
+Configuration: remove `mthreads_gpu` from `BlackList` to enable. Metric groups are controlled by `[MetricCollector.Mthreads]` toggles (`EnableHealth`, `EnablePCIe`, `EnableMTLink`).
+
+|Metric|Description|Unit|Labels|Source|
+|----|---|---|---|---|
+|library_info|MTML library info.|-|version|mtml.LibraryVersion|
```
huhong-web: [P2] This inventory lists internal `metric.Data` names, not the Prometheus names users can query. `pkg/metric.(*Data).prometheusMetric` calls `BuildFQName(DefaultNamespace, collector, name)`, so this row is emitted as `huatuo_bamai_mthreads_gpu_library_info` (and every following row has the same namespace/collector prefix), matching the PR's own usage example. Update both English and Chinese tables and add a registry-gather contract test so the documented names cannot drift.

core/metrics/mthreads/mtml/lib.go
```diff
@@ -0,0 +1,336 @@
...
+	Close() error
+	Handle() uintptr
+}
+
+// library coordinates reference counting and symbol registration for libmtml.so.
+type library struct {
+	sync.Mutex
+	refcount int32
+	dl       dynamicLibrary
+	loadedAs string // name that successfully loaded
```
huhong-web: [P2] This file is not gofmt-clean: `gofmt`, `gofumpt`, and `goimports` all align this comment with two spaces. Since `make check` runs the formatters and then `git diff --exit-code`, the required check will fail on a clean checkout. Run `make import-fmt` (or at least `gofmt`) and commit the result.

huhong-web (changes requested):
## Follow-up Code Review

Requesting changes on `37a388ae..4b7e181e`: 8 actionable findings are attached inline (4 P1, 4 P2), and the native-library boundary still has no tests. The previous undefined-`cfg` compile blocker and SONAME discovery issue are fixed, but the cache/config fixes are incomplete and the new optional-symbol implementation introduces failure-path crashes and handle leaks.

## Architecture & Design

**Verdict: needs changes.**

The module boundary remains directionally sound: `core/metrics/mthreads_gpu.go` owns scrape policy, `mthreads/mtml` owns the vendor API, and `mthreads/dl` owns `dlopen`/`dlclose`. The framework also serializes calls to each collector through `pkg/metric/collector.go:143-149`, so a mutex around every scrape is unnecessary. The failure is inside the MTML lifecycle contract. The actual path is `newMthreadsGpuCollector -> mtml.Init -> library.load -> registerSymbols`; callbacks are stored in package globals, while fallback and recovery unload the code those callbacks target. The init-error path even unloads before formatting an `*mtml.Error`, whose `Error` method calls the now-stale `mtmlErrorString`. Optional init/free symbols are also treated independently, so a supported constructor with a missing destructor leaks a native object on every scrape.

The smallest robust lifecycle design is to keep the shared object loaded for the process lifetime and reinitialize only the MTML interface/handle, matching the vendor binding's shutdown model ([official binding](https://github.com/MooreThreads/mthreads-ml-py/blob/144369f45c29db693e6469dec0e327e25be533e0/pymtml.py#L1002-L1024)). If unloading is a hard requirement, bind every callback into a per-open immutable function table, validate capability pairs, and atomically retire/clear that table before `dlclose`; package-global callbacks must not outlive their library.

The configuration boundary is still split. `cmd/huatuo-bamai/config.publishConfig -> collector.Set` hot-publishes an immutable snapshot, and one snapshot is now correctly taken per scrape. However, `collectMthreadsDeviceInfo` still uses `EnablePCIe` while building a one-time cache, so later updates affect current PCIe metrics but not max-capability metrics. Cache only config-independent facts, then apply all three gates while assembling each scrape (or invalidate by a documented generation). Stating restart-only behavior in the docs does not make the mixed runtime behavior coherent.

Error semantics also need to cross the layer boundary. `Update` has a typed MTML recovery path, and `metric.Collector` explicitly permits partial metrics with an error, but the per-device helpers discard every non-`NotSupported` error. A driver reset therefore produces silent metric disappearance and a successful scrape instead of recovery or `collector_success=0`. Suppress only capability absence; propagate operational failures with partial data.

Compatibility evidence is insufficient for the remaining native calls. Nine signatures are marked as disassembly-inferred and are absent from the vendor's public MTML 2.2 header ([header](https://github.com/MooreThreads/mthreads-ml-py/blob/144369f45c29db693e6469dec0e327e25be533e0/mtml_2.2.0.h)); `Dlsym` proves only that a name exists, not that the guessed ABI is safe. This PR already documents two related calls that SIGSEGV in the daemon. Use only published APIs unless a supported version contract and real driver/GPU test prove the inferred calls.

**Verification gate:** no `*_test.go` file is added by this 2,254-line change. Add (1) fake shared-library contract tests for required/optional symbols, init failure, fallback, and init/free pairs; (2) collector tests for incomplete-cache retry, `false -> true -> false` config transitions, and propagation of non-`NotSupported` errors; and (3) one hardware integration smoke test that gathers the registry, checks the actual metric names, and exercises shutdown/re-init. These tests observe the highest-risk boundaries and cannot be replaced by source-only unit tests.

## Other checks

- All seven new Go files have the required 2026 Apache 2.0 header; `git diff --check` passes.
- `gofmt`, `gofumpt`, and `goimports` all report the same formatting change in `mtml/lib.go`; `make check` would therefore dirty the tree and fail its final diff check.
- `go test -mod=vendor ./core/metrics/mthreads/...` and `go vet -mod=vendor ./core/metrics/mthreads/...` pass on both the PR head and GitHub's generated merge ref, but both packages report `[no test files]`. A full generated-BPF `make check` was not safe in this workspace with less than 0.5 GB free; the installed `golangci-lint` also rejects the repository's configuration version. GitHub reports no status checks for this head.
- Security review found no new SQL/command injection, secret, path-traversal, authentication, or cryptography issue. The native ABI and stale-callback availability risks are covered inline.
- No performance blocker is claimed without profiling; MTML calls dominate this scrape path. The existing 10-second context is only cooperative between synchronous C calls and does not bound a hung call, which the code documents.

---
*Co-authored by: gpt-5.6-sol*

core/metrics/mthreads/mtml/types.go
huhong-web: Still unresolved on `cb690553`. Comparing the current `Dlsym` set with the vendor's public MTML 2.2 header leaves eight invoked symbols with no published declaration: `mtmlDeviceGetMusaComputeCapability`, `mtmlDeviceGetPcieReplayCounter`, `mtmlDeviceGetPerformanceState`, `mtmlGpuGetEnforcedPowerLimit`, `mtmlGpuGetPowerManagementDefaultLimit`, `mtmlGpuGetTemperatureThreshold`, `mtmlGpuGetVoltage`, and `mtmlMemoryGetTemperature`. Removing the earlier "disassembly-inferred" source comments does not establish an ABI contract; `Dlsym` still proves only that a name exists, not that the parameter and return layouts match. Please remove these metrics for now, or gate them to a vendor-supported MTML/driver version and cover every call with a real driver/GPU integration test.

core/metrics/mthreads_gpu.go
```diff
@@ -0,0 +1,711 @@
...
+		return metrics, err
+	}
+
+	// Recoverable MTML error: try exactly one Shutdown + Init + cache
+	// invalidate + retry cycle.
+	log.Warnf("mthreads: recoverable error, re-initing MTML: %v", err)
+	if serr := mtml.Shutdown(); serr != nil {
+		return nil, fmt.Errorf("recoverable error %v; shutdown failed: %w", err, serr)
+	}
+	if ierr := mtml.Init(); ierr != nil {
```
huhong-web, line 114: [P1] Keep the collector out of `collect` after a failed re-init. `Shutdown` sets the library handle to zero and `dlclose`s the shared object, but all package-global purego callbacks still target that mapping. If this `Init` fails, the collector remains registered and its cache is invalidated only after success; the next `Update -> collect -> InitDeviceByIndex` can therefore invoke an unmapped callback and SIGSEGV the daemon. A successful reload can also retain an old optional callback because `registerSymbols` never clears optionals that are absent in the new library. The vendor binding deliberately leaves the shared object loaded and resets only the interface handle ([source](https://github.com/MooreThreads/mthreads-ml-py/blob/144369f45c29db693e6469dec0e327e25be533e0/pymtml.py#L1002-L1024)); prefer that lifecycle. If unloading is required, bind/clear an immutable per-open function table and guard every API call with a ready state. Add fake-library tests where recovery init fails and where an optional symbol disappears across reload; the following scrape must return an error without invoking stale code.

core/metrics/mthreads_gpu.go
```diff
@@ -0,0 +1,711 @@
...
+	)
+	eg, subCtx := errgroup.WithContext(ctx)
+	// Go 1.22+ scopes the loop variable per-iteration, so the closure
+	// below captures the correct `gpu` without the `gpu := gpu` rebind
+	// that was required pre-1.22. (go.mod: `go 1.24.0`.)
+	for _, gpu := range cache.devices {
+		eg.Go(func() error {
+			if subCtx.Err() != nil {
+				log.Debugf("mthreads: scrape timeout exceeded before device %d, stopping collection", gpu)
+				return nil
```
huhong-web: [P2] Report this deadline instead of treating the skipped device as a successful collection. A worker that starts after the parent deadline returns `nil`; because every worker also returns `nil` and `eg.Wait()` is discarded, `allErrs` may remain empty and `collect` returns cached/partial metrics with no error, causing `huatuo_bamai_scrape_collector_success` to be 1. Return `subCtx.Err()` here and join the `eg.Wait()` result into the scrape error (or append the context error under the mutex). Add a deterministic pre-expired-context test.

core/metrics/mthreads_gpu.go
```diff
@@ -0,0 +1,711 @@
...
+		return nil
+	}
+	defer dev.Close()
+
+	gpuLabel := strconv.Itoa(int(gpu))
+
+	// device_info: name/brand/serial. All three are static per-device fields
+	// fetched once during buildCache; their MTML calls rarely fail, so we
+	// silently default to "" on error rather than abort the whole cache build.
+	name, _ := dev.Name()
```
huhong-web: [P2] Do not cache a transient MTML failure as an empty label for the rest of the process. `buildCache` rejects only a `nil` slice, while this function unconditionally appends `device_info` after device initialization; therefore failures from `Name`, `Brand`, `SerialNumber`, and the other adjacent static reads still produce a non-nil partial snapshot that is stored until an unrelated re-init. Return `(metrics, error)`, suppress only `mtml.IsNotSupported`, and refuse to publish the cache on operational errors. Test a first-call timeout/driver failure followed by a successful read and verify the next scrape rebuilds the field.

core/metrics/mthreads_gpu.go
```diff
@@ -0,0 +1,711 @@
...
+	brandStr := "MTT"
+	if b, err := dev.Brand(); err == nil {
+		brandStr = "MTT" + strconv.Itoa(int(b))
```
huhong-web: [P2] Map the vendor enum instead of concatenating its numeric value. The public header defines `MTML_BRAND_MTT = 0` and `MTML_BRAND_UNKNOWN = 1` ([source](https://github.com/MooreThreads/mthreads-ml-py/blob/144369f45c29db693e6469dec0e327e25be533e0/mtml_2.2.0.h#L324-L330)), so this emits `brand="MTT0"` for an MTT device and `brand="MTT1"` for the unknown case. Use an explicit switch (for example `MTT`, `UNKNOWN`, and a stable unknown-value fallback) and cover both published values.

core/metrics/mthreads_gpu.go
```diff
@@ -0,0 +1,711 @@
...
+	if cfg.Mthreads.EnableHealth {
+		if ctx.Err() != nil {
+			record("health ctx", ctx.Err())
+		} else {
+			m, g, mem, vpu, err := collectMthreadsHealth(ctx, dev, gpu)
+			// Sub-handles must be closed before the device. Register in
+			// reverse so LIFO fires them in vpu → mem → gpu order.
+			defer func() {
+				if vpu != nil {
+					_ = vpu.Close()
```
huhong-web, line 458: [P2] Propagate native free failures instead of discarding them; the same applies to `mem.Close()`, `g.Close()`, and the deferred `dev.Close()`. Each wrapper currently zeros its handle even when MTML returns timeout, busy, or driver failure, so a possibly-live native object cannot be retried and the error never reaches `Update`'s recovery path. Perform cleanup explicitly and join cleanup errors into the returned scrape error. Only retire the Go handle after release is confirmed, or encode a documented non-retryable ownership rule if MTML guarantees consumption on failure. Add fake free functions that fail and verify cleanup order plus error propagation.

docs/configuration/huatuo-bamai-configuration_en.md
```diff
@@ -1012,6 +1012,52 @@ This section defines collection rules for various system and network metrics. Al
...
+
+- **EnablePCIe**: Controls collection of PCIe link metrics.
+
+  Default: false. When enabled, collects: current PCIe link speed and width (`pcie_link_speed_gt_per_sec`, `pcie_link_width_lanes`), max capability (`pcie_link_max_speed_gt_per_sec`, `pcie_link_max_width_lanes`), and replay counter (`pcie_replay_total`).
+
+- **EnableMTLink**: Controls collection of MtLink interconnect metrics.
+
+  Default: false. When enabled, collects: per-link state (`mtlink_state`) and bandwidth (`mtlink_bandwidth_gbps`).
```
huhong-web: [P2] Keep this contract aligned with the emitted metrics. `collectMthreadsMtLink` now exports `mtlink_link_bandwidth_gb_s` and `mtlink_link_count`; it never emits `mtlink_bandwidth_gbps`. Update this line and the Chinese mirror at line 1065 to list both actual names (and retain the distinction that bandwidth/count are device-level static specs, while `mtlink_state` is per-link).

huhong-web (changes requested):
## Follow-up Code Review

Requesting changes on `4b7e181e..cb690553`: six new actionable findings are attached inline (one P1, five P2). I also rechecked the existing P1 ABI thread; eight invoked symbols still lack public declarations. The previous compile, SONAME discovery, formatting, config-gating, and dynamic-error propagation findings were addressed, but the PR remains blocked by native lifecycle/ABI safety and untested boundary behavior.

## Architecture & Design

**Verdict: needs changes (blocked).**

The dependency direction is appropriate: `core/metrics/mthreads_gpu.go` owns scrape policy, `core/metrics/mthreads/mtml` owns the vendor API, and `core/metrics/mthreads/dl` owns loader mechanics. Configuration also has a clear path through `cmd/huatuo-bamai/config.publishConfig -> collector.Set -> configSnapshot`, and `pkg/metric/collector.go` serializes each collector's `Update`. The added `dynamicLibrary` seam is a reasonable test boundary.

The lifecycle contract across those layers is still unsafe. The recovery chain is `Update -> collect -> mtml.Shutdown -> mtml.Init -> collect`. `Shutdown` unloads the shared object while callbacks remain in package globals; a failed `Init` leaves the registered collector and its old cache available to the next scrape, so `InitDeviceByIndex` can jump to unmapped code. Optional callbacks are also not cleared before rebinding. The simplest design is to keep the shared object mapped for process lifetime and reinitialize only the MTML interface handle, matching the [official binding](https://github.com/MooreThreads/mthreads-ml-py/blob/144369f45c29db693e6469dec0e327e25be533e0/pymtml.py#L1002-L1024). If unloading is a hard requirement, use an immutable per-open function table, atomically retire it before `dlclose`, and reject every API call unless the client is ready.

The error/cache contract also crosses layers incompletely. Static collection currently treats operational failures as absent fields and publishes the partial result forever; deadline and native-free errors can disappear before `Update`, so `CollectorManager` reports a successful scrape. Return partial metrics together with typed errors, suppress only `NotSupported`, and cache only a fully completed static snapshot. The user-facing data contract also needs the vendor brand enum and the documented MtLink metric names to match the emitted labels/series.

**Verification gate:** this 2,416-line change still adds no `*_test.go` file. Add fake-shared-library contract tests covering required/optional symbols, optional-symbol removal across reload, failed recovery initialization, init/free pairs, and free failures; collector tests covering transient static-field failure, pre-expired deadlines, configuration transitions, and partial-metric error reporting; and a real MTML driver/GPU integration smoke test that gathers the registry and exercises shutdown/re-init. Native ABI and lifecycle behavior cannot be established by compilation alone.

## Other checks

- All new Go files contain the required 2026 Apache 2.0 header; `gofmt`, `gofumpt`, `goimports`, and `git diff --check` are clean.
- `make bpf-build` and `go run ./build/bpfabi-tool` pass. `go test -mod=vendor ./core/metrics ./core/metrics/mthreads/...` and `go vet -mod=vendor ./core/metrics` pass; the two new MThreads packages report `[no test files]`. `go test -race -mod=vendor ./core/metrics/mthreads/...` likewise passes without exercising any test.
- The installed `golangci-lint v2.5.0` cannot load this repository's empty-version configuration, so that check could not run; GitHub exposes no status checks for this head.
- Security review found no new injection, secret, path-traversal, authentication, or cryptography issue. The native stale-callback and unsupported-ABI memory-safety/availability risks are covered in the inline threads.

---
*Co-authored by: gpt-5.6-sol*

core/metrics/mthreads_gpu.go
dowzyx: 已修改，参考官方实现，仅dlopen一次，中途不会dlclose

core/metrics/mthreads_gpu.go
dowzyx: 已修改

core/metrics/mthreads_gpu.go
dowzyx: 已修改

core/metrics/mthreads_gpu.go
dowzyx: 已修改

core/metrics/mthreads_gpu.go
dowzyx: device close之前会调用接口，增加调用结果检查，只有检查全部资源释放才执行句柄赋值为0

docs/configuration/huatuo-bamai-configuration_en.md
dowzyx: 已修改

core/metrics/mthreads/mtml/errors.go
```diff
@@ -0,0 +1,109 @@
...
+	return errors.As(err, &e) && e.code == ErrorNotSupported
+}
+
+// IsRecoverable reports whether err indicates a transient driver/library
+// state that a Shutdown+Init cycle may fix. These codes are typically returned
+// after a driver upgrade, device hot-plug, or when the driver state has been
+// torn down externally (e.g. modprobe -r mthreads).
+func IsRecoverable(err error) bool {
+	var e *Error
+	if !errors.As(err, &e) {
```
huhong-web: [P1] Inspect every branch of the joined scrape error. `collect` returns `errors.Join(allErrs...)`, and device goroutines append errors in completion order. `errors.As` stops at the first matching `*Error`; if that error is `ErrorInvalidArgument` while a later sibling is `ErrorDriverFailure` or `ErrorTimeout`, this function returns false and `Update` skips its only re-init attempt. Traverse `Unwrap() []error` recursively (or give return codes sentinel/`Is` semantics and use `errors.Is` for each recoverable code). Add deterministic tests for `errors.Join(nonRecoverable, recoverable)` in both orders.

core/metrics/mthreads_gpu.go
```diff
@@ -0,0 +1,732 @@
...
+		out = append(out, metric.NewGaugeData("device_info", 1, "GPU info.", map[string]string{
+			"gpu":    gpuLabel,
+			"name":   name,
+			"brand":  brandStr,
+			"serial": serial,
+		}))
+	}
+
+	// pci_info
+	if sbdf, err := dev.PciSbdf(); err == nil {
```
huhong-web, line 352: [P2] The identity reads now call `record`, but this read and the static reads through `InitMemory`/`BusWidth` still discard non-`NotSupported` errors. A transient driver failure therefore leaves `errs` empty, `collectMthreadsDeviceInfo` reports success, and `buildCache` stores permanently incomplete static metrics. Call `record` for every failed static read (including memory initialization and child fields), retain the centralized `NotSupported` suppression, and publish the cache only after all operational reads succeed. Test a first-call `DriverFailure` followed by a successful rebuild.

docs/configuration/huatuo-bamai-configuration_en.md
```diff
@@ -1012,6 +1012,52 @@ This section defines collection rules for various system and network metrics. Al
...
+
+  Default: false. When enabled, collects: current PCIe link speed and width (`pcie_link_speed_gt_per_sec`, `pcie_link_width_lanes`), max capability (`pcie_link_max_speed_gt_per_sec`, `pcie_link_max_width_lanes`), and replay counter (`pcie_replay_total`).
+
+- **EnableMTLink**: Controls collection of MtLink interconnect metrics.
+
+  Default: false. When enabled, collects: device-level static specs (per-link bandwidth `mtlink_link_bandwidth_gb_s` and link count `mtlink_link_count`) and per-link state (`mtlink_state`).
+
+**Library discovery**: At startup, the collector searches for `libmtml.so.2` then `libmtml.so` via the system dynamic linker (respecting `LD_LIBRARY_PATH` and `/etc/ld.so.cache`). If no library is found, the collector logs a warning and is disabled for the lifetime of the process.
+
+**Restart semantics**: Configuration changes in `[MetricCollector.Mthreads]` take effect after restarting `huatuo-bamai`. Library path discovery is performed only at startup; changing `LD_LIBRARY_PATH` or installing a new MTML version also requires a restart.
```
huhong-web: [P2] This contradicts the implemented hot-configuration contract. `config.Update` calls `publishConfig`, which calls `collector.Set`, and the collector takes `configSnapshot()` on every `Update`; for an already initialized collector, `EnableHealth`, `EnablePCIe`, and `EnableMTLink` therefore take effect on the next scrape. Only library discovery (and enabling a collector that failed startup initialization) requires a process restart. Rewrite this paragraph and the Chinese mirror accordingly, and cover a `false -> true -> false` transition with a collector contract test.

huhong-web (changes requested):
## Follow-up Code Review

Requesting changes on current head `0ddc1730` (following the prior review of `cb690553`). Three new actionable findings are attached inline (one P1, two P2). This head fixes the stale-callback risk by retaining the shared object for the process lifetime, pairs optional constructors with their destructors, propagates pre-expired worker deadlines, reports identity-field failures, maps the brand enum, and aligns the MtLink names. The PR is still blocked by recovery correctness, the unsupported native ABI surface, resource-release error handling, and the absence of tests at the new FFI boundary.

## Architecture & Design

**Verdict: needs changes (blocked).**

The dependency direction remains appropriate: `core/metrics/mthreads_gpu.go` owns scrape policy, `core/metrics/mthreads/mtml` owns the vendor API, and `core/metrics/mthreads/dl` owns loader mechanics. Keeping `libmtml.so` mapped for the process lifetime is also the right lifecycle choice for package-global purego callbacks, and capability-pair validation improves ownership safety.

The recovery contract across those layers is still incomplete. Per-device workers return nested `errors.Join` values, but `mtml.IsRecoverable` inspects only the first matching `*mtml.Error`; concurrent completion order can therefore decide whether the collector performs its only recovery attempt. In addition, `Update -> Shutdown -> Init` leaves the registered collector with a zero MTML interface handle if re-init fails. The next scrape enters native calls anyway, and an `InvalidArgument` result is not recoverable, so the collector may never retry initialization. Model the MTML client as an explicit ready/not-ready lifecycle, reject calls before readiness, and make a later scrape able to retry initialization without invoking native APIs with a zero handle.

The native API contract remains a merge blocker. Eight symbols still invoked by this collector have no declaration in the vendor's public MTML 2.2 header: `mtmlDeviceGetMusaComputeCapability`, `mtmlDeviceGetPcieReplayCounter`, `mtmlDeviceGetPerformanceState`, `mtmlGpuGetEnforcedPowerLimit`, `mtmlGpuGetPowerManagementDefaultLimit`, `mtmlGpuGetTemperatureThreshold`, `mtmlGpuGetVoltage`, and `mtmlMemoryGetTemperature`. `Dlsym` establishes only name availability, not parameter or return layout. Remove these calls, or gate them behind a vendor-supported version contract and prove every call on real driver/GPU combinations.

The cache and configuration contracts also need one source of truth. Identity-field failures now prevent cache publication, but the remaining static PCI/spec/BIOS/MUSA/memory reads still convert operational failures into permanently missing cached series. Conversely, the documentation says every MThreads option is restart-only even though `config.Update -> publishConfig -> collector.Set -> configSnapshot` applies the three metric-group toggles on the next scrape. Preserve `NotSupported` as capability absence, propagate other static errors, and document live option changes separately from startup-only library discovery.

**Verification gate:** this 2,458-line change adds no `*_test.go` file. Add fake-library contract tests for required/optional symbols, failed re-initialization, joined-error classification, constructor/destructor pairs, and cleanup failures; collector tests for transient static-field errors and `false -> true -> false` configuration transitions; plus a real MTML driver/GPU smoke test that gathers the registry and exercises shutdown/re-init.

Existing blocker threads remain at [the ABI contract](https://github.com/ccfos/huatuo/pull/641#discussion_r3829812106), [failed re-init readiness](https://github.com/ccfos/huatuo/pull/641#discussion_r3842933351), and [native cleanup errors](https://github.com/ccfos/huatuo/pull/641#discussion_r3842933383).

## Other checks

- All new Go files contain the required 2026 Apache 2.0 header; `gofmt`, `gofumpt`, `goimports`, and `git diff --check` are clean.
- `go test -vet=off -mod=vendor ./core/metrics/mthreads/...` passes, but both new packages report `[no test files]`.
- The broader `core/metrics` test compiled all Go packages but its final native link exited with `ld returned 1` and no linker diagnostic. Earlier full-check attempts were also blocked by workspace disk pressure; `make golangci-lint` could not complete with this environment's read-only default Go caches and missing Cap'n Proto `/std` include. These are verification-environment limitations, not PR diagnostics.
- GitHub exposes no status checks for this head.
- Security review found no SQL/command injection, hardcoded secret, path-traversal, authentication, or cryptography issue. Native ABI memory-safety/availability risk is covered above.
- No performance blocker is claimed without profiling. Device fan-out is bounded by detected hardware, and synchronous MTML calls dominate the scrape path; the documented timeout remains cooperative rather than an in-flight C-call bound.

---
*Co-authored by: gpt-5.6-sol*

core/metrics/mthreads_gpu.go
huhong-web: The current head fixes the unmapped-callback portion of this thread by keeping `libmtml.so` loaded for the process lifetime. The readiness portion remains: after a successful `Shutdown`, a failed `Init` leaves `libmtml.lib == 0` while the collector remains registered. The next scrape calls `CountDevice`/device APIs with that zero interface handle; if MTML returns `InvalidArgument`, `IsRecoverable` rejects it and no later scrape retries initialization. Please gate all API calls on an explicit ready state and let a not-ready `Update` retry `Init` before entering collection. A fake-library test should fail the first re-init, then succeed on the following scrape without any native call receiving a zero interface handle.

core/metrics/mthreads/mtml/errors.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	if joiner, ok := err.(interface{ Unwrap() []error }); ok {
+		for _, e := range joiner.Unwrap() {
+			if hasRecoverableCode(e) {
+				return true
+			}
+		}
+		return false
+	}
+	var e *Error
```
huhong-web, line 64: [P1] Recurse through single-error wrappers before classifying the leaf. The production path wraps a joined collector error with `fmt.Errorf("health: %w", err)` (and can wrap it again while building the cache). At that wrapper, `errors.As` returns only the first `*Error`; if it is `InvalidArgument`, a later `DriverFailure` in the join is never examined and `Update` skips recovery. Walk both `Unwrap() error` and `Unwrap() []error`, and add tests for `fmt.Errorf("health: %w", errors.Join(nonRecoverable, recoverable))` in both leaf orders.

core/metrics/mthreads/mtml/init.go
```diff
@@ -0,0 +1,89 @@
...
+// but shutdown the interface"): all purego-registered mtml* function
+// pointers (types.go) keep pointing into the .so code section, so
+// subsequent mtml.* calls stay valid. A subsequent Init() re-issues
+// mtmlLibraryInit to obtain a fresh MtmlLibrary* handle without
+// reloading the .so or re-registering any symbol.
+func Shutdown() error {
+	libmtml.Lock()
+	defer libmtml.Unlock()
+
+	libmtml.ready = false
```
huhong-web: [P1] Do not transition to not-ready until `mtmlLibraryShutDown` succeeds. On a busy/timeout/failure return, this currently leaves the old nonzero `lib` owned by the process but advertises `ready=false`; the next `Init` can then overwrite or zero that still-live interface pointer. Preserve the ready/handle state on shutdown failure (or model an explicit failed lifecycle state), never initialize over a retained handle, and cover shutdown-failure followed by retry with fake callbacks.

core/metrics/mthreads_gpu.go
```diff
@@ -0,0 +1,766 @@
...
+		} else {
+			record("memory vendor", err)
+		}
+		if w, err := mem.BusWidth(); err == nil {
+			labels["bus_width"] = strconv.Itoa(int(w))
+		} else {
+			record("memory bus width", err)
+		}
+		if len(labels) > 1 {
+			out = append(out, metric.NewGaugeData("memory_info", 1, "GPU memory info.", labels))
```
huhong-web: [P1] Emit a stable label schema for every `memory_info` sample. `NotSupported` is suppressed, so different devices can reach this line with `gpu+type`, `gpu+vendor`, or all four keys. `pkg/metric/data.go` caches the descriptor by metric name only: different label counts make `MustNewConstMetric` panic, while equal counts with different keys silently map a value to the first descriptor's label name. Always include `gpu`, `type`, `vendor`, and `bus_width` with documented fallback values (while still surfacing operational errors), and add a registry-gather test with heterogeneous device capabilities.

core/metrics/mthreads_gpu.go
```diff
@@ -0,0 +1,766 @@
...
+	// Double-check: another goroutine may have completed the build while we
+	// were waiting for the lock.
+	if cache := c.cache.Load(); cache != nil {
+		return cache, nil
+	}
+	cache, err := c.buildCache(ctx)
+	if err != nil {
+		return nil, err
+	}
+	c.cache.Store(cache)
```
huhong-web, line 217: [P2] Do not permanently cache an empty or stale device enumeration. A successful `CountDevice()==0` build is stored here before `collect` returns “no devices”; all later scrapes take the cache fast path and never discover a device that becomes available. Added devices are also never seen, and removed indices do not necessarily produce one of the configured recoverable codes. At minimum, reject an empty cache before storing it; preferably refresh topology and cache static data by stable device identity. Cover `0→1`, `1→2`, and `2→1` transitions.

core/metrics/mthreads_gpu.go
```diff
@@ -0,0 +1,766 @@
...
+// collectMthreadsVersions emits the library and driver version info metrics.
+func collectMthreadsVersions(ctx context.Context) []*metric.Data {
+	var out []*metric.Data
+
+	if ctx.Err() != nil {
+		log.Debug("mthreads: scrape timeout exceeded before collecting library version, skipping")
+		return nil
+	}
+
+	if v, err := mtml.LibraryVersion(); err == nil {
```
huhong-web, line 279: [P2] Distinguish optional capability absence from operational failure here. Any `LibraryVersion` error is currently discarded, and the resulting cache is then published permanently without `library_info`; a transient driver/library error is never retried until an unrelated re-init. Return the error from this helper, suppress only `NotSupported`/missing-symbol behavior if the metric is intentionally optional, and avoid publishing the cache on other failures. Add a first-call-fails/second-call-succeeds test.

core/metrics/mthreads/mtml/init_test.go
```diff
@@ -0,0 +1,193 @@
...
+	}
+}
+
+func TestInit_WithoutLoadedSo_FailsAndLeavesReadyFalse(t *testing.T) {
+	// No .so on this test box (CI), or loadLocked has no candidates:
+	// Init must surface the error and leave ready=false. We do not assert
+	// on the error text — the candidate list depends on the host.
+	resetLibrary(t)
+	err := Init()
+	if err == nil {
```
huhong-web, line 165: [P2] Register native cleanup before taking this skip path. On an MTML-equipped host `Init` has already acquired a reference-counted interface, but the test skips without `Shutdown`; later reset helpers only zero the Go fields and cannot release that interface. This can leak native state and invalidate the subsequent shutdown/re-init lifecycle tests. Use `t.Cleanup` immediately after successful `Init` and do not erase a live handle without releasing it.

core/metrics/mthreads_gpu.go
```diff
@@ -0,0 +1,766 @@
...
+	metrics, err := c.collect(ctx)
+	if err == nil || !mtml.IsRecoverable(err) {
+		return metrics, err
+	}
+
+	// Recoverable MTML error: try exactly one Shutdown + Init + cache
+	// invalidate + retry cycle.
+	log.Warnf("mthreads: recoverable error, re-initing MTML: %v", err)
+	if serr := mtml.Shutdown(); serr != nil {
+		return nil, fmt.Errorf("recoverable error %v; shutdown failed: %w", err, serr)
```
huhong-web: [P2] Preserve both errors in the returned error tree. The configured `errorlint` check fails here and at line 104 because the triggering recoverable error is formatted with `%v`; callers can only unwrap the shutdown/re-init error. Go supports multiple `%w` operands, so use `fmt.Errorf("recoverable error: %w; shutdown failed: %w", err, serr)` and the equivalent re-init form (or `errors.Join` with contextual wrappers).

core/metrics/mthreads/mtml/lib.go
```diff
@@ -0,0 +1,306 @@
...
+		strings.Join(errs, "\n  "))
+}
+
+// closeLocked is a no-op kept for API symmetry. The vendor binding
+// (pymtml.py: "Leave the library loaded, but shutdown the interface")
+// never dlcloses libmtml.so — only the MtmlLibrary* interface handle
+// is reset. We follow the same lifecycle: once loadLocked succeeds,
+// the dlopen mapping is kept for the process lifetime, and so are
+// all purego-registered function pointers.
+func (l *library) closeLocked() {
```
huhong-web: [P2] Remove this unused no-op. The current repository lint job fails `unused` on `closeLocked`; the lifecycle explanation can remain on `Shutdown`/`loadLocked` without retaining an uncalled method.

core/metrics/mthreads/mtml/types.go
```diff
@@ -0,0 +1,265 @@
...
+	// mtmlDeviceGetMtLinkState(MtmlDevice*, unsigned int linkId, MtmlMtLinkState* state)
+	// MtmlMtLinkState is an enum (int): 0=DOWN, 1=UP, 2=DOWNGRADE.
+	mtmlDeviceGetMtLinkState func(uintptr, uint32, *uint32) Return
+)
+
+// mtmlPciInfo mirrors the C struct MtmlPciInfo layout for purego.
+// sbdf is a fixed char[32]; followed by unsigned ints, two floats (speed), and
+// rsvd[6] int padding. Field order/type must match the C definition exactly.
+//
+// nolint:unused // fields are placeholders for C struct layout; purego reads by memory offset,
```
huhong-web: [P2] Fix the suppression syntax at this line and lines 185/197. `nolintlint` does not recognize `// nolint:unused` because of the space and fails the current CI job. Prefer removing an unnecessary suppression; if it is required for ABI padding fields, use the accepted `//nolint:unused // ...` form with the existing justification.

core/metrics/mthreads_gpu.go
```diff
@@ -0,0 +1,766 @@
...
+	}
+
+	spec, err := dev.MtLinkSpec()
+	if err != nil {
+		record("mtlink spec", err)
+		return out, errors.Join(errs...)
+	}
+	// Bandwidth/count are per-device static specs (not per-link live throughput),
+	// so they don't get a `link` label.
+	out = append(out, metric.NewGaugeData("mtlink_link_bandwidth_gb_s", float64(spec.BandWidth),
```
huhong-web: [P2] Combine this append with the immediately following `mtlink_link_count` append. The configured `appendCombine` linter reports these adjacent appends and currently fails the required lint job.

huhong-web (changes requested):
## Follow-up code review

Reviewed current head `11a6172d0f0ca2dcee3ed553d51cd33f11d72f9b` against the base and the unresolved review threads. I am requesting changes: the current revision still has lifecycle and FFI-contract blockers, and the repository lint job is red.

### Architecture & Design

**Verdict: needs changes.**

The broad dependency direction is reasonable: `mthreads_gpu.go` owns scrape/cache policy, `mthreads/mtml` isolates the vendor API, and `mthreads/dl` owns dynamic loading. Keeping the shared object mapped for the process lifetime is also consistent with the package-global purego callbacks. The latest revision improves SONAME discovery, configuration gating, constructor/destructor pairing, static-field error propagation, brand mapping, and metric documentation.

The remaining cross-boundary contracts are not safe enough to merge:

- `memory_info` can produce one metric name with different label schemas, while `pkg/metric` caches descriptors by metric name alone. That can panic during collection or silently attach values to the wrong label names.
- Error recovery does not traverse a join hidden behind the production `fmt.Errorf("%w")` wrappers, so recovery remains dependent on error ordering.
- A failed native shutdown leaves contradictory ownership state and permits reinitialization over a retained interface handle.
- Private MTML entry points outside the public 2.2 header remain in production paths without a supported driver/library compatibility contract or an integration test that invokes them: https://github.com/ccfos/huatuo/pull/641#discussion_r3829812106
- Native cleanup failures are still discarded and sub-handles surrender ownership after failed frees: https://github.com/ccfos/huatuo/pull/641#discussion_r3842933383

The inline comments give concrete fixes and regression-test shapes. Before merge, please also resolve the two linked existing blockers and exercise every retained private ABI on an explicitly supported MooreThreads driver/GPU matrix.

### Review perspectives and verification

Three independent passes covered architecture/cross-layer contracts, correctness/security, and Go engineering quality; the findings below were then rechecked against the same BASE...HEAD diff.

- `go test -mod=vendor ./core/metrics/mthreads/...`: passed (hardware smoke cases skipped without MTML).
- `go vet -mod=vendor ./core/metrics/mthreads/...`: passed.
- `gofmt`, `gofumpt`, `goimports`, and `git diff --check`: passed.
- Apache-2.0 copyright headers: present on the new Go files.
- GitHub **Build, Lint and Vendor**: failed on the current head with seven lint errors called out inline; compilation completed before lint failed: https://github.com/ccfos/huatuo/actions/runs/32927403971
- The separate `sync` failure is repository/infrastructure-related (missing checkout token), while VM jobs were still pending during review.
- No SQL/command injection, hard-coded secret, path-traversal, authentication, or cryptography issue was found. The material safety concern is the unsupported native ABI above.
- Full `make check` was not completed locally because the workspace lacked sufficient temporary disk; targeted checks and the authoritative CI result are reported above.

---
*Co-authored by: gpt-5.6-sol*

PR #643, opened 2026-08-18
fix(symbol): demangle C++ function names

pioneerlfn:
## What

User-space stack frames are reported with their raw Itanium-ABI mangled names, so
C++ profiles are unreadable:

```
_ZN5doris6Thread16supervise_threadEPv
```

This resolves them through `github.com/ianlancetaylor/demangle`:

```
doris::Thread::supervise_thread(void*)
```

Kernel frames and plain C symbols are untouched; anything that fails to demangle
keeps its original name.

## Why lazy

Demangling every symbol while loading the ELF is too expensive on real targets.
A Doris BE binary carries ~439k `.dynsym` and ~476k `.symtab` FUNC symbols, and
demangling them all up front stretched a 30-second capture to ~66 seconds.

Instead the symbol is demangled when a frame actually resolves to it, and the
display name is cached on the resolver. Measured cost per unique symbol:

| symbol | cost |
|---|---|
| plain C name (`malloc`, early-out) | 4.25 ns, 0 allocs |
| simple C++ | 805 ns, 30 allocs |
| heavily templated | 27.5 us, 400 allocs |

The resolver walks every frame of every sample, so the cache matters; the
early-out keeps pure-C stacks free.

## Why template parameters are kept

`demangle.NoTemplateParams` would cut the longest observed symbol from 1435 to
175 characters, but it is wrong here. Doris drives much of its work through
`std::function` callbacks, where the only thing naming the callback lives inside
the template argument. Measured on a real 30-second Doris BE capture, 14 of 195
`doris::`-bearing symbols lose the Doris name entirely when template parameters
are stripped, and all 14 collapse to the same string:

```
before: std::_Function_handler<void (), doris::WalManager::_init_wal_dirs_info()::$_0>::_M_invoke(std::_Any_data const&)
after:  std::_Function_handler::_M_invoke(std::_Any_data const&)
```

Since the aggregation key is the frame string, that would merge unrelated
background threads into a single flamegraph bar.

`demangle.NoRust` is passed so old-style Rust names are not mistaken for C++;
new-style `_R...` names still demangle.

## Concurrency

`UsymResolver` is used from a single goroutine at all four call sites
(`cmd/profiler/provider`, `cmd/perf`, `internal/profiler/memray`), and its five
existing maps have never been synchronised. No lock is added for the new cache —
a lock on only the new map would suggest the type is goroutine-safe when
`exeCache`/`procmaps` would still race — and the type now documents the
constraint instead.

## Testing

Built and run on Linux/x86_64:

- `go vet ./internal/symbol/` — clean
- `go test ./internal/symbol/` — pass
- `go test -race ./internal/symbol/` — pass
- `make check` (goimports + gofumpt + golangci-lint, 26 linters) — exit 0, no findings
- The cache test is mutation-verified: deleting the cache-read fast path in
  `displayName` makes `TestUsymResolverDisplayName` fail, and restoring it turns
  the test green again.

End-to-end against a live Doris BE (`--type cpu --language c++ --thread-group`,
30-second capture, results read back from Elasticsearch):

| | before | after |
|---|---|---|
| unique symbols | 467 | 753 |
| symbols still starting with `_Z` | 73 | **0** |
| symbols containing `doris::` | 0 | 85 |

Capture wall time was 30s for a 30s configured duration, i.e. no regression.

## Not addressed

`unknown lib-no-sym/...` frames are unrelated to demangling and remain. They come
from frame-pointer-based unwinding (`bpf_get_stackid` with `BPF_F_USER_STACK`)
walking into code built without frame pointers — sampling 400 functions in the
target's `libc-2.17.so` found only 3.8% with a frame-pointer prologue, versus
96.5% in `doris_be`. The symbol tables are complete; the addresses are not real
return addresses. Fixing that needs DWARF/`.eh_frame`-based unwinding and belongs
in its own change.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -25,6 +25,7 @@ require (
 	github.com/grafana/grafana-plugin-sdk-go v0.251.0
 	github.com/grafana/pyroscope v1.7.1
 	github.com/grafana/pyroscope/api v0.4.0
+	github.com/ianlancetaylor/demangle v0.0.0-20260724033716-83e58baca724
 	github.com/jsimonetti/rtnetlink v1.4.2
 	github.com/mdlayher/netlink v1.7.2
 	github.com/opencontainers/runtime-spec v1.2.0
--- a/go.sum
+++ b/go.sum
@@ -311,6 +311,8 @@ github.com/hashicorp/serf v0.10.1 h1:Z1H2J60yRKvfDYAOZLd2MU0ND4AH/WDz7xYHDWQsIPY
 github.com/hashicorp/serf v0.10.1/go.mod h1:yL2t6BqATOLGc5HF7qbFkTfXoPIY0WZdWHfEvMqbG+4=
 github.com/hashicorp/yamux v0.1.1 h1:yrQxtgseBDrq9Y652vSRDvsKCJKOUD+GzTS4Y0Y8pvE=
 github.com/hashicorp/yamux v0.1.1/go.mod h1:CtWFDAQgb7dxtzFs4tWbplKIe2jSi3+5vKbgIO0SLnQ=
+github.com/ianlancetaylor/demangle v0.0.0-20260724033716-83e58baca724 h1:QixF8Mcbe87ET7pK/fPbBJ9GXFddmEY8yYMepzMzo30=
+github.com/ianlancetaylor/demangle v0.0.0-20260724033716-83e58baca724/go.mod h1:gx7rwoVhcfuVKG5uya9Hs3Sxj7EIvldVofAWIUtGouw=
 github.com/invopop/yaml v0.2.0 h1:7zky/qH+O0DwAyoobXUqvVBwgBFRxKoQ/3FjcVpjTMY=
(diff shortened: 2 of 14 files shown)


pioneerlfn:
extra explanation:
https://claude.ai/code/artifact/d9e22066-d2c2-4b55-bc78-054672d24dbf

hao022:
@pioneerlfn 

发现的问题:
[P2] symbols.go:44 — 注释与行为不符: 代码使用 demangle.NoRust 禁用了 Rust demangling，但注释写的是 "C++/Rust"。请修正注释为 "C++ symbol" 或改用无选项。
[P2] symbols.go:44 — 函数位置: demangleSymbolName 定义在 symbols.go 中但只在 usymbols.go 中使用，建议移到 usymbols.go。
[P2] usymbols.go:41 — 并发安全: 新增的 names map 未加锁。注释已说明非并发安全，但请确认全部调用方确实是串行的。

PR #669, opened 2026-08-25
fix(metrics): refresh RDMA links per scrape

yyqdbngt:
## Summary

- discover RDMA links on every collector update instead of only at construction
- allow later scrapes to recover from transient link-discovery failures
- add regression coverage for device hotplug and discovery errors

## Why

The collector cached the RDMA device list when it was created. Devices added afterward never appeared, while removed devices stayed in the list. Refreshing the kernel link view per scrape keeps the exported series aligned with current hardware.

## Validation

- `gofmt` completed for the changed Go files
- `git diff --check` passed
- the targeted Go test was attempted, but this metrics package depends on Linux-only files that are excluded by the Windows build; the repository CI should run the Linux build and tests

### DIFF
--- a/core/metrics/netdev_rdma_link.go
+++ b/core/metrics/netdev_rdma_link.go
@@ -1,4 +1,4 @@
-// Copyright 2025 The HuaTuo Authors
+// Copyright 2025, 2026 The HuaTuo Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


core/metrics/netdev_rdma_link.go
```diff
@@ -29,26 +29,30 @@ func init() {
 }
 
 type rdmaLink struct {
-	rdmaList []*netlink.RdmaLink
+	listLinks  func() ([]*netlink.RdmaLink, error)
+	statistics func(*netlink.RdmaLink) (*netlink.RdmaDeviceStatistic, error)
 }
 
 func newRdmaLink() (*tracing.EventTracingAttr, error) {
-	lists, err := netlink.RdmaLinkList()
```
hao022, line 36: rdma 热插拔的场景在真实的生产环境存在吗？在new的时候调用list 主要考虑到性能。因此没有强烈需求，不建议修改。

PR #675, opened 2026-08-29
fix(tracing): reject path traversal in task binary name

unbridled-41:
## Problem

`POST /tasks` forwards the caller-supplied `tracer_name` to `exec.CommandContext` via `path.Join(TaskBinDir, name)` (`pkg/tracing/task.go`) with no validation. `path.Join` cleans dot-dot segments away, so `../../bin/bash` resolves outside the tool directory. The agent HTTP API has no authentication (`huatuo-bamai` registers no `RequireAuth`), so this is unauthenticated arbitrary-command execution on a privileged node (the agent runs BPF), with output readable back through `GET /tasks/:id` (arbitrary file read, e.g. `../../bin/cat /etc/shadow`).

## Reproduction (before fix)

Unit PoC (verified locally, prior to the fix):

```
attacker tracer_name="../002/escape.sh" joins to "/tmp/.../002/escape.sh"
status=completed data="PWNED-ESCAPE\n"   <- executed binary outside TaskBinDir
```

## Fix

`validateTaskBinary` at the single task creation gate (`NewTaskWithIDLimit`): the name must be a bare file name inside the tool directory (`filepath.Base(name) == name`, rejecting empty/`.`/`..`/separators/absolute paths).

## Tests

- `pkg/tracing`: `TestValidateTaskBinary` (8 cases), `TestNewTaskWithIDLimitRejectsPathTraversal` (end-to-end rejection, asserts nothing executed).
- Fail-first verified: removing the validation makes the traversal test fail.

## End-to-end (real agent + curl, combined build)

| request | result |
|---|---|
| `{"tracer_name":"../../etc/passwd"}` | HTTP 400 `must be a bare file name within the tool directory` |
| `{"tracer_name":"/bin/echo"}` | HTTP 400 |
| `{"tracer_name":"sub/dir"}` | HTTP 400 |
| `{"tracer_name":"ok.sh"}` (legit) | task completed, output returned — no false positive |

## Verification

### DIFF
--- a/pkg/tracing/task.go
+++ b/pkg/tracing/task.go
@@ -22,6 +22,7 @@ import (
 	"math/big"
 	"os/exec"
 	"path"
+	"path/filepath"
 	"sort"
 	"sync"
 	"time"
@@ -178,6 +179,9 @@ func NewTaskWithIDLimit(
 	if taskID == "" {
 		return "", errors.New("task id is required")
 	}
+	if err := validateTaskBinary(execBinary); err != nil {
+		return "", err
+	}
 	taskCreateMu.Lock()
 	defer taskCreateMu.Unlock()
 	if _, loaded := taskLifeTmpCache.Load(taskID); loaded {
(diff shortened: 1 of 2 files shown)


hao022:
@unbridled-41 未来会删除这个接口，不在使用；最新的代码很快会提交上来。

PR #679, opened 2026-08-29
fix(metrics): convert tcp memory pages with the kernel page size

unbridled-41:
## Problem

`tcp_memory` hardcoded 4 KiB when converting the sockstat TCP `mem` field to bytes (`*p.Mem * 4096`). The kernel reports `mem` in native pages, so on hosts whose kernel uses larger pages (64 KiB is common on aarch64: RHEL/Amazon Linux arm, Kunpeng, Phytium) `usage_bytes` is wrong by 16x — silently, corrupting dashboards and alerts.

## Fix

Convert with `os.Getpagesize()` (hoisted into a package var for tests), matching the pending sockstat fix. Also converts to float64 before multiplying, avoiding int overflow on 32-bit builds.

## Tests

- `core/metrics`: `TestParseTcpMemoryUsesKernelPageSize` — fixture procfs tree (`tcp_mem` + `sockstat`), page size overridden to 64 KiB; asserts pages/bytes/limit and that bytes track the configured page size, not 4096.

## Verification

### DIFF
--- a/core/metrics/net_tcp_memory.go
+++ b/core/metrics/net_tcp_memory.go
@@ -76,7 +76,7 @@ func parseTcpMemory() (*tcpMemoryStat, error) {
 		if p.Mem != nil {
 			return &tcpMemoryStat{
 				memoryPages: float64(*p.Mem),
-				memoryBytes: float64(*p.Mem * 4096),
+				memoryBytes: float64(*p.Mem) * float64(defaultHostPageSize),
 				memoryLimit: float64(values[2]), // tcpMemLimit
 			}, nil
 		}


core/metrics/net_tcp_memory.go
```diff
@@ -18,13 +18,20 @@ import (
...
 	"huatuo-bamai/pkg/metric"
 	"huatuo-bamai/pkg/tracing"
 	"huatuo-bamai/pkg/types"
 )
 
+// pageSize is the kernel page size in bytes. tcp_mem and the sockstat
+// mem field count native pages, so converting with a hardcoded 4 KiB
+// misreports hosts whose kernel uses larger pages (e.g. 64 KiB on
+// aarch64).
+var pageSize = int64(os.Getpagesize())
```
hao022: core/metrics/net_sockstat.go 已经引入了defaultHostPageSize 因此可以直接使用；删除单元测试（这些测试意义不大）；

unbridled-41: 已按建议调整，现已直接复用 defaultHostPageSize，并移除了新增的单元测试。最终 PR diff 仅保留 TCP memory page-to-bytes 转换的一行修改。感谢 review！

PR #680, opened 2026-08-29
fix(toolstream): close connection when client handshake fails

unbridled-41:
## Problem

`NewClient` returned the handshake error without closing the dialed UDS connection, leaking one fd per failed dial whenever the server went away between dial and connect frame. A probe loop (300 failed handshakes against a closing server) showed +158 fds held at measurement time on the unfixed code — under a fast retry loop the process can exhaust its fd limit before GC finalizers release them.

## Fix

Close the connection on the handshake error path.

## Tests

- `internal/toolstream/transport`: `TestNewClientClosesConnOnHandshakeFailure` — a `dialUDS` test seam returns a recording conn whose writes fail; the test asserts `Close` was invoked. Deterministic (no GC timing involved).
- Fail-first verified: reverting the close fails the test.

## Verification


internal/toolstream/transport/client.go
```diff
@@ -132,15 +132,20 @@ func (c *Client) End() error {
...
 func NewClient(path, toolName, version, taskID string) (*Client, error) {
-	conn, err := DialUDS(path)
+	conn, err := dialUDS(path)
 	if err != nil {
 		return nil, err
 	}
 
 	c := &Client{encoder: capnp.NewEncoder(conn), conn: conn}
 	if err := c.handshake(toolName, version, taskID); err != nil {
+		// The caller never sees the connection, so it must not leak.
```
hao022, line 144: 这是一个非常不错的修复；建议直接使用DialUDS，不需要增加DialUDS单元测试；直接修复问题 即可。

unbridled-41: 已按建议调整，现已直接复用 DialUDS，并移除了额外的 test seam 和相关测试。核心修复保持为 handshake 失败时关闭连接。感谢 review！

PR #685, opened 2026-08-29
fix(profiler): propagate off-CPU drain failures instead of spinning

unbridled-41:
## Problem

The off-CPU profiler drain loop retried `ReadBatch` forever on any error that was neither lost-samples nor the exit marker — without logging. A permanently broken reader or an ABI-drift decode failure produced an empty profile with no signal. `ReadBatch` joins real failures with the lost-samples count, so `errors.As(err, &lostErr)` is true even for real failures and the check was insufficient.

## Fix

Return the error unless it carries nothing but lost-sample information (`isOnlyLostSamples` helper unwraps joined errors).

## Tests

- `cmd/profiler/provider`: `TestIsOnlyLostSamples` — nil, pure loss, real failure, and failure-joined-with-loss cases.

## Verification

### DIFF
--- a/cmd/profiler/provider/native_offcpu_profiler.go
+++ b/cmd/profiler/provider/native_offcpu_profiler.go
@@ -154,6 +154,12 @@ func (p *cpuNativeProfiler) readOffCPUDataLoop(
 			if errors.Is(err, types.ErrExitByCancelCtx) {
 				return nil
 			}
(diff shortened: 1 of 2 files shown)


hao022:
@unbridled-41 这个问题已经通过其他方式优化。区分了事件丢失，和具体错误。

PR #662, opened 2026-08-24
fix(pod): bound kubelet response bodies

tju-yxq:
## Summary

- bound kubelet `/pods` and `/configz` response reads with a fixed 128 MiB
  request-layer limit;
- derive the limit from reproducible 200-400 Pod serialization profiles instead
  of a per-Pod arithmetic assumption;
- reject declared oversized responses before consuming their payload;
- enforce the decoded-size limit on unknown-length, chunked, and automatically
  decompressed streams;
- emit warning logs when declared or observed response size exceeds the limit;
- cap non-OK and malformed-JSON diagnostic previews at 8 KiB;
- keep the safeguard internal to the kubelet HTTP layer, without adding Pod
  configuration or documentation surface.

Fixes #661.

## Problem

Kubelet discovery called `io.ReadAll(resp.Body)` for every response before
checking the HTTP status. Neither successful `/pods` and `/configz` responses
nor non-OK diagnostic bodies had a byte boundary.

The five-second HTTP timeout limits elapsed time, not response size. A local
kubelet, proxy, incorrectly routed service, or compromised listener can still
deliver a large body within that interval. Successful responses remain in
memory until JSON decoding completes, while non-OK or malformed bodies can also
be copied into errors and logs.

## Kubelet response format

HuaTuo currently depends on Kubernetes API v0.31.3. Kubernetes v1.31.3 handles
`/pods` by obtaining the complete desired Pod set, copying it into a
`v1.PodList`, and encoding it through the v1 legacy codec:

- source: `pkg/kubelet/server/server.go`, `encodePods` and `getPods`;
- the response therefore includes each Pod's metadata, spec, and status rather
  than a reduced discovery projection.

Kubelet v1.31.3 defaults `maxPods` to 110, so 200-400 Pods is treated here as an
explicit dense-node requirement rather than the default deployment case.
Kubernetes also limits the total annotation key/value bytes on one object to
256 KiB. Both facts are included in the response-budget analysis below.

### Method

A reproducible proof program used Kubernetes v0.31.3's
`LegacyCodec(v1)`—the same encoding family used by kubelet `encodePods`—to
serialize PodLists containing 110, 200, and 400 copies of four profiles.

Profiles:

1. **Representative dense workload**
   - 4 containers and 1 init container;
   - 12 environment variables per container;
   - 8 volumes and mounts;
   - 12 labels;
   - 16 KiB annotations;
   - 1 KiB previous-termination messages.
2. **Spec/status-heavy workload**
[...]

PR #693, opened 2026-08-29
fix(container): bound container API responses

### LINKED ISSUE #692: [Bug]: container client reads API responses without a memory bound
## Problem

`internal/command/container.getContainers` reads container API responses
without a byte limit. Successful responses are decoded directly from the
stream, and non-OK bodies are read in full and copied into an error string.
The three-second timeout limits elapsed time, not bytes, so a misrouted or
malfunctioning local endpoint can cause unbounded allocation.

## Expected behavior

- reject declared successful responses above a fixed practical limit before
  reading them;
- bound streamed successful responses with a limit-plus-one read;
- cap non-OK diagnostic text and add an explicit truncation marker;
- preserve body read errors and normal response decoding behavior.

## Suggested validation

Cover below-limit, exact-limit, streamed overflow, read-error, declared-length
rejection, and bounded error-preview behavior.

yyqdbngt:
## Summary

- cap successful container API responses at 32 MiB
- keep at most 8 KiB of an error response and mark truncation
- add regression coverage for limits, truncation, and read errors

## Why

The local container API response was read without a byte limit. A malfunctioning endpoint could therefore grow the agent's memory use without bound.

## Validation

- `go test -mod=vendor internal/command/container/body.go internal/command/container/body_test.go -count=1`
- `git diff --check`
- Full package tests and `make check` could not run on this Windows host because the generated Linux BPF ABI is absent and `make` is unavailable; GitHub CI is expected to run the repository checks.

Fixes #692

### DIFF
--- a/internal/command/container/body.go
+++ b/internal/command/container/body.go
@@ -1,0 +1,33 @@
+// Copyright 2026 The HuaTuo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package container
+
+import "io"
+
+const (
+	maxContainerResponseBytes = 32 << 20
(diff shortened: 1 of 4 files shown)


hao022:
@yyqdbngt @tju-yxq 当前补丁和 https://github.com/ccfos/huatuo/pull/662 类似，都是修复http请求大小问题，因此可以在https://github.com/ccfos/huatuo/pull/662 的基础上继续优化扩展。我先将这个pr 关闭掉，等 662 合并后，新的pr可以基于这个再次提交。

hao022:
@yyqdbngt @tju-yxq #662 已经合并，大家可以看看如何有共用的地方，形成一套限制规则。

PR #694, opened 2026-08-29
fix(kmsgutil): preserve semicolons in kernel messages

### LINKED ISSUE #691: [Bug]: kmsg formatter truncates messages containing semicolons
## Problem

`internal/utils/kmsgutil.formatKmsgEntry` splits a `/dev/kmsg` record on every
semicolon and keeps only `parts[1]`. The kernel record delimiter is the first
semicolon; semicolons after it belong to the message. A record such as:

```text
6,1001,2026000;device reset; retry scheduled
```

is currently rendered as `device reset`, losing the diagnostic suffix.

## Expected behavior

Split the header from the message at the first semicolon only, preserving the
complete message text. Malformed records without the delimiter should retain
the current error behavior.

## Suggested validation

Add a regression case containing multiple semicolons and assert that the
formatted message is unchanged after the timestamp prefix.

btlqql:
## Summary

- split each `/dev/kmsg` record only at the header delimiter
- preserve semicolons that belong to the kernel message payload
- add regression coverage for a message containing a semicolon

## Why

The first semicolon separates the kmsg header from the message. Splitting at every semicolon silently dropped diagnostic text after the next delimiter.

## Validation

- `GOOS=linux GOARCH=amd64 go test -c -mod=vendor ./internal/utils/kmsgutil`
- `git diff --check`
- The Linux test binary cannot execute on this Windows host, and `make check` is unavailable because `make` is not installed; GitHub CI is expected to run the repository checks.

Fixes #691

### DIFF
--- a/internal/utils/kmsgutil/kmsgutil.go
+++ b/internal/utils/kmsgutil/kmsgutil.go
@@ -1,4 +1,4 @@
-// Copyright 2025 The HuaTuo Authors
+// Copyright 2025, 2026 The HuaTuo Authors
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -112,7 +112,7 @@ func formatKmsgs(kmsgs string) string {
 
 // convert timestamp to human-readable format
 func formatKmsgEntry(entry string) (string, error) {
-	parts := strings.Split(entry, ";")
+	parts := strings.SplitN(entry, ";", 2)
 	if len(parts) < 2 {
 		return "", fmt.Errorf("invalid entry format")
 	}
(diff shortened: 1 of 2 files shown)


internal/utils/kmsgutil/kmsgutil.go
```diff
@@ -112,7 +112,7 @@ func formatKmsgs(kmsgs string) string {
 
 // convert timestamp to human-readable format
 func formatKmsgEntry(entry string) (string, error) {
-	parts := strings.Split(entry, ";")
+	parts := strings.SplitN(entry, ";", 2)
```
hao022, line 115: 这种方式在性能上有优势，但是没有必要增加单元测试。

btlqql: 已按建议删除新增的单元测试，PR 现在只保留 `SplitN` 的必要修改。

PR #706, opened 2026-09-02
fix(tracing): bound task command output

### LINKED ISSUE #702: [Bug]: tracing tasks buffer unbounded combined output
## What happened?

`pkg/tracing.runTask` uses `exec.Cmd.CombinedOutput`, so a task process can make the agent retain unlimited stdout and stderr in memory. The task timeout bounds runtime but not bytes. A noisy or malfunctioning tool can therefore exhaust the agent before the result is stored or returned.

## Expected behavior

Capture combined output through a bounded writer, terminate/reap the task when the limit is exceeded, and report a clear task error. Normal output and timeout/cancellation behavior must remain unchanged.

## Proposed verification

Add focused task tests for output below the limit and output that exceeds it.

btlqql:
## Summary

- capture task stdout and stderr through a shared 64 MiB bounded buffer
- surface a dedicated task error when combined output reaches the limit
- retain existing timeout, cancellation, and successful-output behavior

## Why

`exec.Cmd.CombinedOutput` grows without a byte limit. A noisy or malfunctioning
tracing tool could exhaust the agent before its result is stored or returned.

Closes #702

## Testing

- `goimports` on changed Go files
- focused tests cover partial writes, repeated writes after the limit, and
  combined command output
- full Linux checks deferred to CI because generated BPF ABI packages are not
  available in the Windows checkout

### DIFF
--- a/pkg/tracing/task.go
+++ b/pkg/tracing/task.go
@@ -101,15 +101,26 @@ var (
 )
 
 type boundedBuffer struct {
-	buffer   bytes.Buffer
-	limit    int
-	exceeded bool
+	mu         sync.Mutex
+	buffer     bytes.Buffer
+	limit      int
+	exceeded   bool
+	onExceeded func()
 }
 
 func (b *boundedBuffer) Write(p []byte) (int, error) {
+	b.mu.Lock()
+	defer b.mu.Unlock()
+
+	if b.exceeded {
+		return 0, ErrTaskOutputLimitExceeded
+	}
 	remaining := b.limit - b.buffer.Len()
 	if remaining <= 0 {
 		b.exceeded = true
+		if b.onExceeded != nil {
+			b.onExceeded()
+		}
 		return 0, ErrTaskOutputLimitExceeded
 	}
 	if len(p) <= remaining {
@@ -117,14 +128,21 @@ func (b *boundedBuffer) Write(p []byte) (int, error) {
 	}
 	n, _ := b.buffer.Write(p[:remaining])
 	b.exceeded = true
+	if b.onExceeded != nil {
+		b.onExceeded()
+	}
 	return n, ErrTaskOutputLimitExceeded
 }
 
 func (b *boundedBuffer) Bytes() []byte {
+	b.mu.Lock()
+	defer b.mu.Unlock()
 	return b.buffer.Bytes()
 }
 
 func (b *boundedBuffer) String() string {
+	b.mu.Lock()
+	defer b.mu.Unlock()
 	return b.buffer.String()
 }
 
@@ -298,8 +316,11 @@ func runTask(ctx context.Context, task *task) {
 }
 
 func runTaskCommand(ctx context.Context, binary string, args []string, limit int) ([]byte, error) {
-	output := &boundedBuffer{limit: limit}
-	cmd := exec.CommandContext(ctx, binary, args...)
+	outputCtx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	output := &boundedBuffer{limit: limit, onExceeded: cancel}
+	cmd := exec.CommandContext(outputCtx, binary, args...)
 	cmd.Stdout = output
 	cmd.Stderr = output
 	err := cmd.Run()
--- a/pkg/tracing/task.go
+++ b/pkg/tracing/task.go
@@ -15,6 +15,7 @@
 package tracing
 
 import (
+	"bytes"
 	"context"
 	"crypto/rand"
 	"errors"
@@ -44,6 +45,8 @@ const (
 )
 
 var TaskBinDir = "bin"
+
+const maxTaskOutputBytes = 64 * 1024 * 1024
 
 type TaskStorageType int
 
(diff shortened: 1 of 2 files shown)


pkg/tracing/task.go
```diff
@@ -258,6 +297,18 @@ func runTask(ctx context.Context, task *task) {
 	task.mu.Unlock()
 }
 
+func runTaskCommand(ctx context.Context, binary string, args []string, limit int) ([]byte, error) {
+	output := &boundedBuffer{limit: limit}
+	cmd := exec.CommandContext(ctx, binary, args...)
+	cmd.Stdout = output
+	cmd.Stderr = output
+	err := cmd.Run()
+	if output.exceeded {
```
huhong-web, line 306: [P1] [CR-001] The 64 MiB output limit reports an error only after cmd.Run waits for child exit. A child can exceed the limit then sleep, retaining an active task slot until an unrelated timeout. Cancel or kill the child when the limit is reached and test a noisy-then-sleep process.\n\nFix: apply the stated contract at this boundary and add a focused regression test.

huhong-web (commented):
## Code Review

One blocking finding is included inline: pkg/tracing/task.go:306.

## Architecture & Design
- Verdict: Needs changes.
- Key evidence: pkg/tracing/task.go:305-312 does not bind output-limit detection to subprocess lifecycle cancellation.
- Risk and recommendation: Resolve the inline contract violation before approval.

## Scope & Focus
- Verdict: Focused
- Primary objective: Bound task command output.
- Necessary changes: pkg/tracing/task.go:305-312 and its lifecycle tests.
- Out-of-scope changes: None.
- Split plan: Not required.

## Review Convergence
- Mode: full
- Base SHA: ae0badc0a1fed0a1403ff352edd33723aeafaf3e
- Reviewed HEAD: 7696bc7c19e9827f54c131a7254a2c8035173868
- Scope verdict: Focused
- Coverage: Changed files and direct contracts, correctness/security and architecture/quality perspectives.
- New findings: 1 FIRST_PASS / BLOCKING
- Completeness declaration: At HEAD 7696bc7c19e9827f54c131a7254a2c8035173868, all changed human-authored files and assigned dimensions were covered; all known blockers are included.

## Verification
- Targeted verification: Not required.

---

btlqql:
Closing this one: `pkg/tracing/task.go` was removed by `ccec9dda`, so the diff cannot be rebased and the branch would reintroduce a deleted file.

The review on this PR was still valid, though. The [`[P1]` on `task.go:327`](https://github.com/ccfos/huatuo/pull/706#discussion_r3944533638) asked for the child to be stopped when the output limit is reached, and the relocated code in `internal/exec` had the same gap. That is now fixed and tested in #834 (`fix(exec): stop a command that exceeds its output limit`, `Fixes #833`), with a `noisy-then-sleep` regression that fails on current `main` and passes with the change.

Thanks @huhong-web — the contract question was the right one.

btlqql:
Closing this one: `pkg/tracing/task.go` was removed by `ccec9dda`, so the diff cannot be rebased and the branch would reintroduce a deleted file.

The review on this PR was still valid, though. The [`[P1]` on `task.go:327`](https://github.com/ccfos/huatuo/pull/706#discussion_r3944533638) asked for the child to be stopped when the output limit is reached, and the relocated code in `internal/exec` had the same gap. That is now fixed and tested in #834 (`fix(exec): stop a command that exceeds its output limit`, `Fixes #833`), with a `noisy-then-sleep` regression that fails on current `main` and passes with the change.

Thanks @huhong-web — the contract question was the right one.

pkg/tracing/task.go
btlqql: You were right, and thank you for stating the contract precisely. `Process` bounded the *retained* output but not the command's lifetime: nothing signalled the child when the limit was passed, and the error only appeared after `reap` returned from `cmd.Wait`, so a noisy-then-idle child held its slot and every waiter until it exited on its own.

`pkg/tracing/task.go` no longer exists — `ccec9dda` replaced that subsystem, and the bounded-output path now lives in `internal/exec` (`Spec.MaxOutputBytes` plus the `outputBuffer`/`tailBuffer` pair). I have carried the fix there in #834: the buffer hands the first write past the limit to the process, which stops the command with the existing SIGTERM-then-SIGKILL escalation, and `reap` no longer wraps that kill as `ErrStopped` because no caller asked for a stop. The `noisy-then-sleep` regression you asked for is in that PR: it fails on current `main` with `Wait() did not return: the command kept running after passing the output limit` and passes with the change.

btlqql: You were right, and thank you for stating the contract precisely. `Process` bounded the *retained* output but not the command's lifetime: nothing signalled the child when the limit was passed, and the error only appeared after `reap` returned from `cmd.Wait`, so a noisy-then-idle child held its slot and every waiter until it exited on its own.

`pkg/tracing/task.go` no longer exists — `ccec9dda` replaced that subsystem, and the bounded-output path now lives in `internal/exec` (`Spec.MaxOutputBytes` plus the `outputBuffer`/`tailBuffer` pair). I have carried the fix there in #834: the buffer hands the first write past the limit to the process, which stops the command with the existing SIGTERM-then-SIGKILL escalation, and `reap` no longer wraps that kill as `ErrStopped` because no caller asked for a stop. The `noisy-then-sleep` regression you asked for is in that PR: it fails on current `main` with `Wait() did not return: the command kept running after passing the output limit` and passes with the change.

PR #715, opened 2026-09-04
fix(procfs): sum arp_cache counters across per-CPU rows

unbridled-41:
## Problem / Evidence

`NetArpCache` parses `/proc/net/stat/arp_cache` by reading the header and exactly **one** data row (`scanner.Scan()` once for values). The kernel emits one row **per possible CPU** (`neigh_stat_seq_show` iterates `for_each_possible_cpu`), so on any SMP host the reported stats are CPU 0's counters only.

Live evidence on a 16-CPU host (this repo's dev box):

```
entries  allocs   destroys hash_grows lookups  hits     ...
00000006 00000001 00000000 00000000   00000000 00000000 ...   <- CPU 0 (all zeros)
00000006 00000000 00000000 00000000   00000001 00000000 ...
00000006 00000003 00000000 00000000   00000008 00000001 ...   <- real traffic here
```

The `net_arp` collector (`core/metrics/net_arp.go:57`) therefore published `lookups=0, hits=0` while the host was actively resolving neighbors. The existing open PRs touching this file (#507, #552) only add panic/validation guards and still read a single row; the per-CPU semantics are unaddressed.

## Root cause / Fix

The value loop ran once over the first data row. This change:

- iterates **all** remaining rows and **sums** each column, which is the host-wide total for per-CPU kernel counters;
- checks `scanner.Err()` so a short read no longer silently yields an empty stat map;
- stops indexing `headers[num]` out of range when a data row has more fields than the header (previously a slice panic; existing tests asserted neither behavior).

Single-row and malformed fixtures keep their previous behavior; the existing test cases pass unchanged.

## Priority

PRIORITY 83/100: impact 32 (collector ships wrong values on every SMP host), scope 16 (net_arp collector output), reproducibility 20 (deterministic, fixture-backed; live file on this box shows 17 rows), maintenance 15 (consistent with how per-CPU proc counters must be aggregated). FIX_CONFIDENCE 95.

## Tests

```
go test ./internal/procfs/ -run TestNetArpCache -v
```

- New `PerCpuLinesAreSummed` case fails before the fix (entries=0xa, only CPU 0) and passes after (sum over 3 rows = 0xd).
- Full package: `ok  huatuo-bamai/internal/procfs`. `go build ./...`, `gofmt -l`, `go vet` clean.

## Risk

Low. `entries` is now the sum of per-CPU neighbor-table sizes (host total, the previously intended semantic). Values can only increase relative to the old single-CPU reading; no other consumer of `NetArpCache` exists.

### DIFF
--- a/internal/procfs/net_arpcache.go
+++ b/internal/procfs/net_arpcache.go
@@ -38,22 +38,33 @@ func NetArpCache() (*ArpCacheStats, error) {
 	defer file.Close()
 
 	scanner := bufio.NewScanner(file)
-	scanner.Scan()
 
 	// First string is always a header for stats
 	var headers []string
-	headers = append(headers, strings.Fields(scanner.Text())...)
+	if scanner.Scan() {
+		headers = append(headers, strings.Fields(scanner.Text())...)
+	}
 
 	// Fast path ...
 	cache := &ArpCacheStats{Stats: make(map[string]uint64)}
 
-	scanner.Scan()
-	for num, counter := range strings.Fields(scanner.Text()) {
-		value, err := strconv.ParseUint(counter, 16, 64)
-		if err != nil {
-			return nil, err
+	// The kernel emits one row of counters per possible CPU
+	// (neigh_stat_seq_show), so the host total is the sum over all rows;
+	// reading a single row would report CPU 0 only.
+	for scanner.Scan() {
+		for num, counter := range strings.Fields(scanner.Text()) {
+			if num >= len(headers) {
+				break
+			}
+			value, err := strconv.ParseUint(counter, 16, 64)
+			if err != nil {
+				return nil, err
+			}
+			cache.Stats[headers[num]] += value
 		}
-		cache.Stats[headers[num]] = value
+	}
+	if err := scanner.Err(); err != nil {
+		return nil, err
 	}
 
 	return cache, nil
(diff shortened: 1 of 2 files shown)


unbridled-41:
Closing this draft after CI evidence disproved the premise - my fix was wrong, not the original code.

**What the integration CI revealed** (all 4 VM runs failed on this branch):

```
level=info msg="collector arp failed ...: strconv.ParseUint: parsing \"...\": invalid syntax"
```

Two facts I had wrong:

1. **The `entries` column is not a per-CPU counter.** In `neigh_stat_seq_show` (net/core/neighbour.c, checked against v6.6) the first column printed is `atomic_read(&tbl->entries)` - a table-global value repeated identically on every per-CPU row. The live file on a 16-CPU host and the integration fixture (`integration/fixtures/proc/net/stat/arp_cache`, two rows of `0000000f` plus a literal `...` row) both confirm this. Only the remaining columns (allocs, lookups, hits, ...) are per-CPU. Summing the file therefore corrupts `entries` (6 entries -> 96 on a 16-CPU host).

2. **The only consumer reads exactly that global column.** `core/metrics/net_arp.go` publishes `huatuo_bamai_arp_total` = `cache.Stats["entries"]` and nothing else from `NetArpCache`, so the old single-row read was already correct for every published metric. My claimed impact ("lookups/hits published as 0") was false - those columns are parsed but never exported.

With the premise gone, the remaining defects (per-CPU columns read from one row, unchecked `scanner.Err()`, potential `headers[num]` index panic) have no call path, so per the project contribution rules they do not qualify as a high-value bug on their own. Withdrawing rather than patching a non-bug; the per-CPU summing variant for the counter columns stays recorded in my notes if a future consumer ever needs them.

Apologies for the noise - the integration VM suite caught exactly what my unit fixture (single-row, different values per row) masked.

PR #716, opened 2026-09-04
fix(profiler): split java collapsed stacks into per-frame samples

unbridled-41:
Fixes #722.

## Problem / Evidence

`javaAggregator.Aggregate` (cmd/profiler/provider/java_aggregator.go) built each sample as

```go
frames := []string{fmt.Sprintf("process %d", so.PID), stack}
```

where `stack` is the whole semicolon-joined collapsed line emitted by async-profiler. The `output.Sample.Frames` contract is **one element per frame**, and the raw formatter (`internal/profiler/output/raw/raw.go`, `writeFoldedFrame`) rewrites every `;` inside a frame to `:`. The asprof input format is the one documented in this repo (`internal/profiler/profiler.go`, `ParseCollapsedData`):

```
HotCode.main;HotCode.hotmethod3;java/util/UUID.randomUUID;... 1
```

So every local-file output (collapsed is the CLI default, plus flamegraph/svg) welded the **entire Java call tree into a single bogus frame**. Verbatim regression-test output, before the fix (line produced by the new test):

```
process 1234;HotCode.main:HotCode.hotMethod:java/util/UUID.randomUUID 50
```

after the fix:

```
process 1234;HotCode.main;HotCode.hotMethod;java/util/UUID.randomUUID 50
```

The java upload/pprof path is unaffected: `ParseCollapsedData` splits collapsed stacks itself; only the local formatter path was missed. The python aggregator already does the correct split (`strings.Split(stack, ";")` in python_aggregator.go), and the native aggregator appends one frame per element.

## Root cause / Fix

Split the collapsed stack on `;` before handing it to the formatter, mirroring the python aggregator:

```go
// Frames must be one element per frame: the raw formatter rewrites
// ';' inside a frame to ':', which would weld the whole collapsed
// stack into a single frame.
frames := append([]string{fmt.Sprintf("process %d", so.PID)}, strings.Split(stack, ";")...)
```

## Priority

PRIORITY 80/100: impact 30 (default collapsed/flamegraph output for Java profiling - a flagship use case - renders a useless two-level graph), scope 15 (all local formatter outputs for java profiles), reproducibility 20 (deterministic, pure in-process), maintenance 15 (restores the Frames contract and python-path parity). FIX_CONFIDENCE 90.

## Tests

[...]

Issue #637, opened 2026-08-17
[Bug]: cgroup v2 unlimited memory is reported as a parse failure


yyqdbngt:
## What happened?

`internal/cgroups/v2.CgroupV2.MemoryUsage` reads both `memory.current` and `memory.max` with `parseutil.ReadUint`. On cgroup v2, an unlimited memory limit is represented by the literal value `max`, not a decimal integer. The current code therefore returns a parse error for a valid unlimited cgroup and the OOM snapshot path loses its memory sample.

A minimal fixture containing:

```text
memory.current: 4096
memory.max: max
```

causes `MemoryUsage` to fail while `CpuQuotaAndPeriod` already handles the equivalent `cpu.max` unlimited value explicitly.

## Expected behavior

Treat `memory.max=max` as an unlimited limit, using the same `math.MaxUint64` representation already used for unlimited CPU quota. Decimal limits and malformed input must retain their existing behavior.

## Proposed verification

Add fixture-backed tests for a decimal limit, `max`, malformed content, and missing files, then run the focused cgroup v2 package tests and repository formatting checks.

PR #717, opened 2026-09-04
fix(cgroups): treat v1 unlimited memory limit as unlimited

unbridled-41:
## Problem / Evidence

`CgroupV1.MemoryUsage` (internal/cgroups/v1/cgroups.go) reads `memory.limit_in_bytes` with `parseutil.ReadUint`. On cgroup v1 an **unlimited** memcg holds the literal `-1` in that file - the kernel documents that writing `-1` resets the limit to unlimited (Documentation/admin-guide/cgroup-v1/memory.rst, "We can write '-1' to reset the `*.limit_in_bytes` (unlimited)"), and reading it back yields `-1`. `ReadUint` only knows the cgroup v2 `max` sentinel, so it fails:

```
$ go test ./internal/cgroups/v1/ -run TestMemoryUsageHandlesUnlimitedLimit -v   # before the fix
    cgroups_test.go:96: strconv.ParseUint: parsing "-1": invalid syntax
--- FAIL: TestMemoryUsageHandlesUnlimitedLimit
```

**Call path and consequence** (both verified in-tree): BPF OOM event -> `oomEventLoop` -> `buildTracingData` -> `cgroupMemorySnapshot` (core/events/oom_snapshot.go:76) -> `cgroup.MemoryUsage(container.CgroupPath)`. The returned error aborts the whole cgroup snapshot, so the OOM tracing record loses `current`, `max`, `memory.stat` and `memory.events` - the forensic fields that matter most. On cgroup v1 hosts (Legacy/Hybrid, both supported by `internal/cgroups/cgroups.go`), every container without a memory limit hits this; unlimited memcgs are the default for pods without memory limits.

The v1 CPU path already handles the identical sentinel explicitly (`quota == -1` -> `math.MaxUint64` in `CpuQuotaAndPeriod`), which shows the intended pattern; the memory path was missed. (Related open issue #637 discusses the v2 `max` sentinel; current v2 code already maps it via `ReadUint`, so this PR deliberately touches the v1 `-1` case only.)

## Root cause / Fix

Read the limit with `parseutil.ReadInt` and map `-1` to `math.MaxUint64`, matching the unlimited CPU quota representation and the v2 `max` read:

```go
maxLimited, err := parseutil.ReadInt(paths.Path(subsystem.SubsystemMemory,
	path, "memory.limit_in_bytes"))
if err != nil {
	return nil, err
}
if maxLimited == -1 {
	return &stats.MemoryUsage{Usage: usage, MaxLimited: math.MaxUint64}, nil
}
```

## Priority

PRIORITY 78/100: impact 30 (OOM forensics silently lost on cgroup v1 hosts for the common unlimited-container case), scope 14 (OOM snapshot path; single consumer), reproducibility 18 (deterministic fixture repro), maintenance 16 (sibling `-1` handling already exists in the CPU path). FIX_CONFIDENCE 90.

## Tests

Fixture-backed tests via `paths.RootfsDefaultPath` override (internal/cgroups/v1/cgroups_test.go, `TestMemoryUsageHandlesUnlimitedLimit`, table-driven):

- `unlimited`: `memory.limit_in_bytes = -1` -> `MaxLimited = MaxUint64`, `Usage` passed through. **Fails before the fix** (output above), passes after.
- `decimalLimit`: `memory.limit_in_bytes = 1073741824` -> unchanged passthrough (guards the `ReadUint` -> `ReadInt` switch).

```
$ go test ./internal/cgroups/v1/ -run TestMemoryUsageHandlesUnlimitedLimit -v
--- PASS: TestMemoryUsageHandlesUnlimitedLimit (0.00s)
    --- PASS: TestMemoryUsageHandlesUnlimitedLimit/unlimited (0.00s)
    --- PASS: TestMemoryUsageHandlesUnlimitedLimit/decimalLimit (0.00s)
$ go test ./internal/cgroups/... ./core/events/ ./pkg/tracing/   # consumers
ok  huatuo-bamai/internal/cgroups/v1 ... ok huatuo-bamai/core/events ... ok huatuo-bamai/pkg/tracing
$ go test ./... -timeout=6m      # full suite on this branch: exit 0
```

`gofmt -l` and `go vet` clean.

## CI

First commit: `Build, Lint and Vendor` pass, `pr_name_lint` pass; VM integration pass on ubuntu 20.04 / 24.04 / 26.04. On ubuntu 22.04, `test_tcp_retransmit_fast.sh` failed ("tcpshark produced no retransmit events"), which is unrelated to this change: the commit only touches `internal/cgroups/v1` (no BPF/tcpshark/retransmit code), the same commit passed the identical test on the other three distros, and the same 22.04 suite passed for sibling PRs #716/#718 raised in the same window. Recorded as environment flake. A fresh CI run for the test-only second commit failed in all four VMs during environment setup, before the repository was built: `.github/workflows/scripts/qemu-2-run-in-vm.sh:137` runs `go install mvdan.cc/sh/v3/cmd/shfmt@latest`, and the newly released shfmt v3.14.0 requires go >= 1.26.0 while the VM images ship go 1.24.0 (the toolchain auto-download then also failed on the proxy). This breaks the VM workflow for every branch right now and is unrelated to the code here. The relevant CI signal remains the first commit's run described above; the test-only commit itself passes locally (`go test ./internal/cgroups/v1/`, see Tests).

## Risk

Low. Only the v1 unlimited-limit reading changes (`error` -> `MaxUint64`); decimal limits, usage reading, and the v2 path are untouched (covered by the new `decimalLimit` case).

### DIFF
--- a/internal/cgroups/v1/cgroups.go
+++ b/internal/cgroups/v1/cgroups.go
@@ -175,11 +175,17 @@ func (c *CgroupV1) MemoryUsage(path string) (*stats.MemoryUsage, error) {
 		return nil, err
 	}
 
-	maxLimited, err := parseutil.ReadUint(paths.Path(subsystem.SubsystemMemory,
+	// cgroup v1 reports an unlimited memcg as -1, which ReadUint cannot
+	// parse; return the same MaxUint64 sentinel the unlimited CPU quota
+	// uses in CpuQuotaAndPeriod and that the v2 "max" read produces.
+	maxLimited, err := parseutil.ReadInt(paths.Path(subsystem.SubsystemMemory,
 		path, "memory.limit_in_bytes"))
 	if err != nil {
 		return nil, err
 	}
+	if maxLimited == -1 {
+		return &stats.MemoryUsage{Usage: usage, MaxLimited: math.MaxUint64}, nil
+	}
 
-	return &stats.MemoryUsage{Usage: usage, MaxLimited: maxLimited}, nil
+	return &stats.MemoryUsage{Usage: usage, MaxLimited: uint64(maxLimited)}, nil
 }
--- a/internal/cgroups/v1/cgroups_test.go
+++ b/internal/cgroups/v1/cgroups_test.go
@@ -15,6 +15,7 @@
 package v1
 
 import (
+	"fmt"
 	"math"
 	"os"
 	"path/filepath"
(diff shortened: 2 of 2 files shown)


unbridled-41:
Follow-up on CI: the fresh run for the test-only commit failed in all four VMs during **environment setup**, before the repository was built. Root cause is repo-wide CI breakage, not this change: `qemu-2-run-in-vm.sh:137` installs `shfmt@latest`, and the newly released v3.14.0 requires go >= 1.26.0 while the VM images ship go 1.24.0 (toolchain auto-download via the proxy also failed). Any branch pushing today will hit the same wall. The relevant CI signal is the first commit's run (Build/Lint/Vendor + 20.04/24.04/26.04 pass; 22.04 tcp_retransmit flake, see description).

unbridled-41:
Closing: the premise does not hold against the kernel source. The reviewer is right - I derived "reading returns -1" from "writing -1 means unlimited", and that inference is wrong.

**What the kernel actually does** (v6.6 `mm/memcontrol.c`, fetched and checked):

- **Write path** - the literal `-1` exists only here: `mem_cgroup_write` parses the input with `page_counter_memparse(buf, "-1", &nr_pages)`, which maps the `-1` sentinel to `PAGE_COUNTER_MAX`.
- **Read path** - `mem_cgroup_read_u64` returns `case RES_LIMIT: return (u64)counter->max * PAGE_SIZE;` - always a non-negative, page-aligned decimal. For an unlimited memcg that is `PAGE_COUNTER_MAX * PAGE_SIZE` = `9223372036854771712` on 64-bit (the well-known k8s/Docker "unset" value), which `parseutil.ReadUint` parses without error.

So on real cgroup v1 hosts `CgroupV1.MemoryUsage` does **not** fail for unlimited memcgs, and the claimed OOM-snapshot data loss (current/max/memory.stat/memory.events dropped) does not occur. The `-1 -> MaxUint64` mapping in my patch is dead code against any real kernel, and my fixture (writing a literal `-1` into `memory.limit_in_bytes`) tests a file content the kernel never produces.

Withdrawing per the contribution rules rather than merging a fix for a non-bug. Thanks to the reviewer for catching the write-vs-read conflation - the same distinction my #715 review pass should have applied to the /proc values before claiming an impact.

Side note for anyone touching this area later: unlimited v1 limits do surface as the finite `9223372036854771712` in `OOMCgroupMemorySnapshot.Max` (vs `MaxUint64` on v2). That is a cosmetic inconsistency, not a parse failure, so I am not filing it.

PR #745, opened 2026-09-06
fix(metrics): keep memory_reclaim zero series per container

zjncs:
## Problem

`memory_reclaim` (`core/metrics/memory_reclaim.go`) uploads zero for containers without reclaim events only when the BPF map is **completely empty**:

```go
// if events haven't happened, upload zero for all containers.
if len(items) == 0 { ... }
```

As soon as a single container stalls, the condition is false and every other container's `container_memory_reclaim_directstall` series disappears from the scrape until it also stalls. On a node where one container reclaims occasionally, all the other containers' series flap present/absent on every event, which breaks `absent()`-style alerts and any consumer expecting a stable label set per container.

## Fix

Emit the zero fallback per container instead of only in the empty-map case: track the css addresses reported from the dumped map, then emit zero for the remaining tracked containers. The documented all-zero behaviour when no events happened anywhere is preserved.

The per-item aggregation moves into `buildReclaimMetrics(items, containersCssMem)` so the semantics can be tested without a BPF object or a live pod store; `Update()` keeps its flow unchanged.

## Tests

- `TestMemoryReclaimReportsZeroForContainersWithoutEvents` — one container with events, one without: both series present, sibling at 0. Fails before the fix (`quiet` missing), passes after.
- `TestMemoryReclaimReportsZeroForAllContainersWhenMapEmpty` — empty map: all containers at 0 (existing behaviour preserved).
- `TestMemoryReclaimIgnoresEntriesWithoutContainer` — a stale cgroup entry with no tracked container produces no metric.

## Verification

- `go test ./core/metrics/ -count=1` — full package green
- `gofmt -l` — clean; `go vet` — clean
- Regression verified fail-before/pass-after in a Linux container (golang:1.24)
- Checked open PRs: none touch `core/metrics/memory_reclaim.go`

Signed-off-by: zjncs <18910855655@163.com>

### DIFF
--- a/core/metrics/memory_reclaim.go
+++ b/core/metrics/memory_reclaim.go
@@ -68,10 +68,17 @@ func (c *memoryCgroupReclaim) Update() ([]*metric.Data, error) {
 		return nil, err
 	}
 
+	return buildReclaimMetrics(items, containersCssMem)
+}
+
+// buildReclaimMetrics turns dumped BPF map items keyed by memory-cgroup css
+// address into per-container directstall metrics.
+func buildReclaimMetrics(items []bpf.MapItem, containersCssMem map[uint64]*pod.Container) ([]*metric.Data, error) {
 	var (
 		reclaimVal memoryBpfStruct
 		cssAddr    uint64
 		data       []*metric.Data
+		reported   = make(map[uint64]struct{}, len(items))
 	)
 	for _, v := range items {
 		keyBuf := bytes.NewReader(v.Key)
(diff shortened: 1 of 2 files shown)


huhong-web (commented):
## Code Review

No blocking findings.

## Architecture & Design
- Verdict: Reasonable.
- Evidence:  keeps BPF lease, pod lookup, and map dumping in , while the extracted pure helper owns only item-to-metric projection. The  set is populated only for tracked CSS addresses at , then the fallback loop at  emits exactly one zero series for every remaining tracked container.
- Contract:  defines the memory-CSS container boundary;  publishes the returned data in the same scrape. No API, configuration, storage, lifecycle, or dependency direction changes were introduced.
- Verification: unit-level coverage at  exercises mixed active/quiet, empty-map, and stale-entry cases. No integration-test trigger was identified.

## Scope & Focus
- Verdict: Focused
- Primary objective: Keep  present with zero for tracked containers without reclaim events.
- Necessary changes:  implements the per-container fallback;  verifies the new and preserved cases.
- Out-of-scope changes: None.
- Split plan: Not required.

## Review Convergence
- Mode: full
- Base SHA: ae0badc0a1fed0a1403ff352edd33723aeafaf3e
- Reviewed HEAD: baee46ed981423821670a37f05c53520a3ea4aa9
- Primary objective: Keep a stable zero-valued memory-reclaim series for every tracked container.
- Scope verdict: Focused
- Coverage: 2 changed files; correctness/error handling/security/concurrency, architecture/design, Go quality/style, copyright, tests, and direct metric/pod contracts; two independent review perspectives completed.
- New findings: 0
- Completeness declaration: At HEAD baee46ed981423821670a37f05c53520a3ea4aa9, all changed human-authored files and the listed review dimensions were covered; all currently known blocking findings are included in this review batch. This is a bounded coverage statement, not a guarantee of zero defects.

## Verification
- Targeted verification: Not required.
- CI: missing — GitHub returned no check runs for this HEAD.

---

PR #781, opened 2026-09-07
fix(job): reject malformed lone dash sort field

### LINKED ISSUE #771: [Bug]: Job List queries accept a lone "-" sort and fail only in the storage backend
### What happened?

`internal/job` validates a job query by stripping a leading `-` from `JobQuery.Sort`. A caller can pass the lone sort value `-`, which validates successfully and is later translated into an empty storage sort field. The SQLite backend then rejects it as an invalid field, so a query that should fail fast with `ErrInvalidQuery` fails later with a backend-specific error.

### Expected behavior

`validateJobQuery` should reject a sort value that becomes empty after the descending-order prefix is removed.

### Relevant files

yyqdbngt:
## Summary

Reject the lone descending sort value `"-"` while validating job queries. Previously `validateJobQuery` stripped the leading dash, accepted the resulting empty field, and the error surfaced later from the SQLite backend.

## Root cause

`validateJobQuery` only checked that the field was non-empty before stripping a leading `-`; it never verified that a field remained after stripping.

## Validation

- `gofmt -w internal/job/storage_store.go internal/job/storage_store_test.go`
- `git diff --check`
- `go test ./internal/job -run '^TestValidateJobQueryRejectsLoneDescendingSort$' -count=1`
- `GOOS=linux GOARCH=amd64 go test -c -mod=vendor ./internal/job`

Fixes #771

### DIFF
--- a/internal/job/storage_store.go
+++ b/internal/job/storage_store.go
@@ -169,6 +169,9 @@ func validateJobQuery(query *JobQuery) error {
 	field := query.Sort
 	if field != "" && field[0] == '-' {
 		field = field[1:]
+		if field == "" {
+			return fmt.Errorf("%w: unsupported sort field %q", ErrInvalidQuery, query.Sort)
+		}
 	}
 	switch field {
 	case "", "id", "created_at", "finished_at", "hostname", "container_id", "status", "type":
--- a/internal/job/storage_store_test.go
+++ b/internal/job/storage_store_test.go
@@ -239,3 +239,10 @@ func TestToStorageQueryUsesJobFieldNames(t *testing.T) {
 		t.Errorf("default sorts=%v, want descending created_at", query.Sorts)
 	}
 }
+
+func TestValidateJobQueryRejectsLoneDescendingSort(t *testing.T) {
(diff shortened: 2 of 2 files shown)


yyqdbngt:
Closing this: the behaviour it guards is already on `main`, so only the assertion was missing, and that assertion now lives in #784 together with the other two superseded cases.

Evidence on top of `852d724d` (Ubuntu 22.04, go1.24.13) — a probe test calling `validateQuerySort` the same way this PR's test does:

```
$ go test -count=1 ./internal/job/ -run TestProbeLoneDashSortRejected -v
=== RUN   TestProbeLoneDashSortRejected
    zz_probe_superseded_test.go:13: validateQuerySort(Sort: "-") = invalid job query: unsupported sort field ""
--- PASS: TestProbeLoneDashSortRejected (0.00s)
```

`main` reaches that through `querySort`, which uses `strings.TrimPrefix(field, "-")` and then validates the result against `jobQueryFields`, so a lone `-` is rejected there. The old branch had to add the guard by hand because it predates that refactor.

The test itself is kept in #784 (`TestValidateQuerySortRejectsLoneDescendingField`), adapted to `validateQuerySort`.

PR #782, opened 2026-09-07
fix(job): apply ID filter in job storage queries

### LINKED ISSUE #772: [Bug]: job List queries ignore the JobQuery.ID filter
### What happened?

`storageStore.List`/`Count` translate `JobQuery` to `driver.Query` in `toStorageQuery`, but `JobQuery.ID` is never added as an equality filter. Listing jobs with a non-empty ID therefore returns all rows matching the other filters instead of only that job.

### Expected behavior

When `JobQuery.ID` is non-empty, the storage query should include an `id` equality filter.

### Relevant files

yyqdbngt:
## Summary

Apply `JobQuery.ID` when translating a job query to the generic storage query. A non-empty ID now becomes an `id = ?` filter instead of being silently ignored.

## Root cause

`toStorageQuery` built filters for user, container, host, status, and type, but never read `JobQuery.ID`.

## Validation

- `gofmt -w internal/job/storage_store.go internal/job/storage_store_test.go`
- `git diff --check`
- `go test ./internal/job -run '^TestToStorageQueryIncludesIDFilter$' -count=1`
- `GOOS=linux GOARCH=amd64 go test -c -mod=vendor ./internal/job`

Fixes #772

### DIFF
--- a/internal/job/storage_store.go
+++ b/internal/job/storage_store.go
@@ -255,6 +255,9 @@ func toStorageQuery(q *JobQuery) driver.Query {
 		q = &JobQuery{}
 	}
 	filters := make([]driver.Filter, 0, 6)
+	if q.ID != "" {
+		filters = append(filters, driver.Filter{Field: "id", Op: driver.OpEq, Value: q.ID})
+	}
 	if q.UserID != "" && !q.IsAdmin {
 		filters = append(filters, driver.Filter{Field: "user_id", Op: driver.OpEq, Value: q.UserID})
 	}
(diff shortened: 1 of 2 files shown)


yyqdbngt:
Closing this: the behaviour it guards is already on `main`, so only the assertion was missing, and that assertion now lives in #784 together with the other two superseded cases.

Evidence on top of `852d724d` (Ubuntu 22.04, go1.24.13) — a probe test calling the query builder the same way this PR's test does:

```
$ go test -count=1 ./internal/job/ -run TestProbeIDFilterPresent -v
=== RUN   TestProbeIDFilterPresent
    zz_probe_superseded_test.go:25: filters = [{Field:id Op:eq Value:job-store-alpha}]
--- PASS: TestProbeIDFilterPresent (0.00s)
```

`main` now builds its filters through `appendEqual("id", query.ID)` in `buildStorageQuery`, so `Query.ID` already becomes an `id` equality filter. This branch predates that refactor and patched `toStorageQuery` by hand.

The test itself is kept in #784 (`TestBuildStorageQueryFiltersByID`), adapted to `buildStorageQuery`.

PR #783, opened 2026-09-07
fix(storage): reject empty IDs and aggregation fields

### LINKED ISSUE #773: [Bug]: Generic storage forwards empty IDs and aggregation fields to backends
### What happened?

`Store.Get`, `Store.Delete`, and `Store.Values` validate pagination but not the required identifier/field argument. Empty strings are forwarded to backend implementations, producing backend-specific behavior instead of the storage layer`s `ErrInvalidField`. `Store.record` already rejects empty IDs for Save/Create; the read, delete, and aggregation paths should be consistent.

### Expected behavior

`Get`/`Delete` reject an empty ID and `Values` rejects an empty field before invoking the backend.

### Relevant files

- `internal/storage/store.go`

yyqdbngt:
## Summary

Validate empty IDs and aggregation fields at the generic `Store` boundary. `Get` and `Delete` now reject an empty ID, and `Values` now rejects an empty field, before any backend call is made.

## Root cause

Only the write path (`Store.record`) validated IDs. Read, delete, and aggregation paths forwarded empty strings to backend implementations.

## Validation

- `gofmt -w internal/storage/store.go internal/storage/store_test.go`
- `git diff --check`
- `go test ./internal/storage -run '^TestStoreRejectsEmptyIDAndFieldBeforeBackend$' -count=1`
- `GOOS=linux GOARCH=amd64 go test -c -mod=vendor ./internal/storage`

Fixes #773


yyqdbngt:
The `Build, Lint and Vendor` job caught a real problem in my revision, in `internal/storage/store_test.go`:

```
git diff --exit-code
diff --git a/internal/storage/store_test.go b/internal/storage/store_test.go
@@ -105,7 +105,7 @@ type testBackend struct {
 	saveCalls          int
-	getCalls            int
+	getCalls           int
 	deleteCalls        int
 	deleteByQueryCalls int
```

The new `getCalls` field I added was one space too far right for the struct's existing alignment; `gofumpt` realigned it in CI, `make check`'s `git diff --exit-code` then failed. Fixed in `474ef247` (the commit is amended, so the diff is otherwise unchanged).

Re-ran the CI-equivalent checks on the V100 with the same toolchain (`/usr/local/go1.24.13`, `gofumpt v0.8.0`, `goimports v0.36.0`, `golangci-lint v1.62.2`, `GOFLAGS=-mod=vendor`) against current `main` (`87de2a7`):

- `gofumpt -l ./internal/storage/` — no output
- `goimports -l -local github.com/ccfos/huatuo ./internal/storage/` — no output
- `make check` — exit 0 (gofumpt + goimports + shfmt + golangci-lint + generated-file diff)
- `make build` — exit 0
- `go test -count=1 ./internal/storage/...` — ok in all five packages (`internal/storage`, `.../driver`, `.../elasticsearch`, `.../localfile`, `.../sqlite`)

Pushing again so the job can re-run.

PR #784, opened 2026-09-07
test(storage): cover query building and the index snapshot

yyqdbngt:
## Summary

Add the regression coverage for three behaviours that `main` already implements but does not assert:

- `NewStore` must take one `mapper.Indexes()` snapshot and hand that same slice to `backend.Init`.
- `validateQuerySort` must reject a lone `-`, which names no field.
- `buildStorageQuery` must turn `Query.ID` into an `id` equality filter.

## Why this is now tests only

This PR originally also changed `internal/storage/store.go` to stop calling `mapper.Indexes()` twice. That production change has since landed on `main` through the storage refactor (`1b270fc9`), so after rebasing onto `852d724d` the code change is a no-op and only the regression test remains. I have retitled it accordingly rather than leaving a `fix(...)` title on a change that no longer fixes anything.

The two `internal/job` cases come from PRs #781 and #782, which are superseded in the same way: probing `main` shows

```
validateQuerySort(&Query{Sort: "-"})      = invalid job query: unsupported sort field ""
buildStorageQuery(&Query{ID: "job-store-alpha"}).Filters = [{Field:id Op:eq Value:job-store-alpha}]
```

so both behaviours already hold and only the assertions were missing. I folded them in here and closed those PRs, so the coverage lives in one place instead of three.

## Testing

On Ubuntu 22.04 with go1.24.13, on top of `852d724d`:

```
$ gofmt -l internal/storage/store_test.go internal/job/storage_store_test.go   # no output
$ go vet  ./internal/storage/ ./internal/job/                                   # clean
$ go test -count=1 ./internal/storage/ ./internal/job/
ok  	github.com/ccfos/huatuo/internal/storage
ok  	github.com/ccfos/huatuo/internal/job
$ go test -count=1 ./internal/storage/ ./internal/job/ \
    -run 'TestNewStoreUsesSingleIndexSnapshot|TestValidateQuerySortRejectsLoneDescendingField|TestBuildStorageQueryFiltersByID' -v
--- PASS: TestNewStoreUsesSingleIndexSnapshot (0.00s)
--- PASS: TestValidateQuerySortRejectsLoneDescendingField (0.00s)
--- PASS: TestBuildStorageQueryFiltersByID (0.00s)
$ go build ./...                                                                # exit 0
```

Assisted-by: OpenAI:Codex (GPT-5)
Percentage of AI-generated code: 100%

### DIFF
--- a/internal/job/storage_store_test.go
+++ b/internal/job/storage_store_test.go
@@ -22,6 +22,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/ccfos/huatuo/internal/storage/driver"
 	"github.com/ccfos/huatuo/pkg/observation"
 	"github.com/ccfos/huatuo/pkg/profiling"
 )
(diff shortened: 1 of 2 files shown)


yyqdbngt:
The `Build, Lint and Vendor` job caught a real formatting problem in my previous revision:

```
$ golangci-lint run -v ./... --timeout=5m --config .golangci.yaml
$ git diff --exit-code
@@ -235,6 +235,7 @@ func jobIDs(jobs []*Job) []string {
 	return ids
 }
+
 func TestValidateQuerySortRejectsLoneDescendingField(t *testing.T) {
make: *** [Makefile:191: check] Error 1
```

I had checked the files with `gofmt`, which is satisfied without a blank line between top-level declarations; this repository also enables `gofumpt` in `.golangci.yaml` and in `Makefile`'s format target, and gofumpt requires one. Fixed, and the local check now runs the same formatters as CI:

```
$ gofumpt -l internal/storage/store_test.go internal/job/storage_store_test.go     # no output
$ goimports -l -local github.com/ccfos/huatuo <the two files>                       # no output
$ gofmt -l <the two files>                                                          # no output
$ golangci-lint run ./internal/storage/... ./internal/job/... --timeout=5m --config .golangci.yaml   # no findings
$ go test -count=1 ./internal/storage/ ./internal/job/
ok  	github.com/ccfos/huatuo/internal/storage
ok  	github.com/ccfos/huatuo/internal/job
```

PR #788, opened 2026-09-07
fix(metrics): skip zero disk iolatency buckets

### LINKED ISSUE #778: [Bug]: disk iolatency collector emits zero histogram buckets while container metrics skip them
### What happened?

`fetchBlkDiskIOlatency` emits `blkdisk_q2c` and `blkdisk_d2c` samples for every zone even when the bucket count is zero. `fetchContainerIOlatency` already skips zero buckets, so the disk path produces noisy empty series that are inconsistent with the container path.

### Expected behavior

Disk latency histogram buckets with a zero count should be omitted, matching the container latency collector.

### Relevant files

- `core/metrics/iolatency_collector.go`

yyqdbngt:
## Summary

Skip zero-count disk iolatency histogram buckets. The disk path now matches the container path, which already skips zero buckets, and both share one helper.

## Root cause

`fetchBlkDiskIOlatency` emitted `blkdisk_q2c` and `blkdisk_d2c` samples for every zone, including zones with a zero count, while `fetchContainerIOlatency` already skipped them. The two disk zone loops were copies of each other, so the skip is added once in `appendBlkdiskZoneMetrics`, which takes the formatted device instead of the whole 120-byte entry (the caller already builds that device for `blkdisk_freeze`).

## Validation

Ubuntu 22.04, kernel 6.8, as root, `-mod=vendor`, tree = current `main` + this patch:

```
--- PASS: TestIOLatencyUpdateReturnsContainerCollectionError (0.00s)
--- PASS: TestAppendBlkdiskZoneMetricsSkipsZeroBuckets (0.00s)
ok  github.com/ccfos/huatuo/core/metrics  0.016s

make build      → 0
sudo make unit  → 0
make check      → 0   (gofumpt + golangci-lint, 0 issues, no generated or formatting drift)
```

The regression test drives the helper with `zones[0] = 10` and `zones[2] = 5` and asserts that exactly two samples are produced, labelled `zone="0"` and `zone="2"`, so a zero-count zone no longer exports `blkdisk_q2c`/`blkdisk_d2c`.

The diff against `main` only adds that helper and its test; the existing `TestIOLatencyUpdateReturnsContainerCollectionError` coverage is preserved.

Fixes #778

Signed-off-by: yyqdbngt <btlqql@gmail.com>

### DIFF
--- a/core/metrics/iolatency_collector.go
+++ b/core/metrics/iolatency_collector.go
@@ -105,21 +105,8 @@ func (c *iolatencyTracing) fetchBlkDiskIOlatency(object bpf.BPF) ([]*metric.Data
 	for _, disk := range blkIOdata {
 		diskDev := fmt.Sprintf("%d:%d", disk.Major, disk.Minor)
 
-		for zone, cnt := range disk.Q2CZone {
-			metrics = append(metrics, metric.NewCounterData(
-				"blkdisk_q2c", float64(cnt),
-				"the disk q2c latency",
-				map[string]string{"disk": diskDev, "zone": strconv.Itoa(zone)},
-			))
-		}
-
-		for zone, cnt := range disk.D2CZone {
-			metrics = append(metrics, metric.NewCounterData(
-				"blkdisk_d2c", float64(cnt),
-				"the disk d2c latency",
-				map[string]string{"disk": diskDev, "zone": strconv.Itoa(zone)},
-			))
-		}
+		metrics = appendBlkdiskZoneMetrics(metrics, diskDev, disk.Q2CZone, "blkdisk_q2c", "the disk q2c latency")
+		metrics = appendBlkdiskZoneMetrics(metrics, diskDev, disk.D2CZone, "blkdisk_d2c", "the disk d2c latency")
 
 		metrics = append(metrics, metric.NewCounterData(
 			"blkdisk_freeze", float64(disk.FreezeNr),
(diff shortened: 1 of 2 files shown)


yyqdbngt:
Rebased onto current `main` (`e86a84b8`) and fixed the two problems of the previous revision.

## 1. `Build, Lint and Vendor` was red — `gocritic: hugeParam`

`core/metrics/iolatency_collector.go:122:2: hugeParam: disk is heavy (120 bytes); consider passing it by pointer (gocritic)`

The helper no longer receives the 120-byte entry; it receives the formatted device that the caller already builds for `blkdisk_freeze`:

```go
metrics = appendBlkdiskZoneMetrics(metrics, diskDev, disk.Q2CZone, "blkdisk_q2c", "the disk q2c latency")
```

## 2. The previous revision deleted an unrelated test

`TestIOLatencyUpdateReturnsContainerCollectionError` (the container error-propagation test added in `59057c85`) was dropped by the earlier revision — not intended. It is back, so the diff against `main` is now only the shared helper plus its regression test:

```
 core/metrics/iolatency_collector.go      | 35 +++++++++++++-------------
 core/metrics/iolatency_collector_test.go | 20 +++++++++++++
 2 files changed, 40 insertions(+), 15 deletions(-)
```

## Verification

Ubuntu 22.04, kernel 6.8, as root, `-mod=vendor`, tree = current `main` + this patch:

```
=== RUN   TestIOLatencyUpdateReturnsContainerCollectionError
--- PASS: TestIOLatencyUpdateReturnsContainerCollectionError (0.00s)
=== RUN   TestAppendBlkdiskZoneMetricsSkipsZeroBuckets
--- PASS: TestAppendBlkdiskZoneMetricsSkipsZeroBuckets (0.00s)
ok  github.com/ccfos/huatuo/core/metrics  0.016s

make build      → 0
sudo make unit  → 0
make check      → 0   (1139 issues scanned, 0 reported, no generated or formatting drift)
```

The regression test drives the helper with `zones[0] = 10`, `zones[2] = 5` and asserts that exactly two samples are produced, labelled `zone="0"` and `zone="2"` — a zero-count zone no longer exports `blkdisk_q2c`/`blkdisk_d2c`, which is what the container path already did.

中文摘要：这个 PR 上一版有两个问题，已一并修好——(1) CI 的 `Build, Lint and Vendor` 报 `gocritic hugeParam`，改为传已格式化好的设备名，不再复制 120 字节的结构体；(2) 上一版误删了 `TestIOLatencyUpdateReturnsContainerCollectionError`，已恢复。现在相对 main 的改动只有一个共享 helper 加一个回归测试，并在真机（Ubuntu 22.04 / kernel 6.8 / root）跑通 `make build`、`make unit`、`make check`。

PR #789, opened 2026-09-07
fix(metrics): track only iolatency containers with blkio CSS

### LINKED ISSUE #779: [Bug]: iolatency never adds containers whose blkio cgroup CSS appears after first discovery
### What happened?

`updateContainerBlkDisk` remembers every container returned by `pod.Containers()`, including containers that do not yet have a `SubsystemBlkIO` cgroup CSS. On the next refresh, those containers are considered already tracked and are skipped, so a container whose blkio cgroup appears after first discovery is never added to the BPF map.

### Expected behavior

Only containers that were actually added to the BPF map should be remembered, so newly available blkio cgroup CSS entries are added on a later refresh.

### Relevant files

- `core/metrics/iolatency_tracing.go`

yyqdbngt:
## Summary

Remember only containers that currently have a `SubsystemBlkIO` cgroup CSS as already tracked. Containers whose blkio cgroup appears after first discovery are now added on a later refresh instead of being skipped forever.

## Root cause

`updateContainerBlkDisk` assigned `latestContainers = containers` (all containers), so a container without `SubsystemBlkIO` at first discovery was considered already tracked and never re-evaluated for BPF map insertion.

## Validation

- `gofmt -w core/metrics/iolatency_tracing.go core/metrics/iolatency_tracing_test.go`
- `git diff --check`
- `go test ./core/metrics -run '^TestTrackContainersWithBlkioCSS$' -count=1` cannot run in this Windows checkout due pre-existing vendored build constraints (`internal/bpf/abi` and `prometheus/procfs/sysfs`); the regression test is intended to run on Linux CI.
- `GOOS=linux GOARCH=amd64 go test -c -mod=vendor ./core/metrics` is blocked by the same pre-existing `internal/bpf/abi` vendored-path error.

Fixes #779

### DIFF
--- a/core/metrics/iolatency_tracing.go
+++ b/core/metrics/iolatency_tracing.go
@@ -220,6 +220,16 @@ func (c *iolatencyTracing) updateContainerBlkDisk(b bpf.BPF) error {
 		}
 	}
 
-	c.latestContainers = containers
+	c.latestContainers = trackContainersWithBlkioCSS(containers)
 	return nil
+}
+
+func trackContainersWithBlkioCSS(containers map[string]*pod.Container) map[string]*pod.Container {
+	tracked := make(map[string]*pod.Container, len(containers))
+	for id, container := range containers {
+		if _, ok := container.CgroupCss[subsystem.SubsystemBlkIO]; ok {
+			tracked[id] = container
+		}
+	}
+	return tracked
 }
--- a/core/metrics/iolatency_tracing_test.go
+++ b/core/metrics/iolatency_tracing_test.go
@@ -1,0 +1,39 @@
+// Copyright 2026 The HuaTuo Authors
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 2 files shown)


yyqdbngt:
Rebased onto the current `main` (`852d724d`); this branch was 126 commits behind.

The rebase needed one fix that git did not flag: `core/metrics/iolatency_tracing_test.go` is a new file in this change and was written before the module rename, so it imported

```
	"huatuo-bamai/internal/cgroups/subsystem"
	"huatuo-bamai/internal/pod"
```

After `89b00992` ("refactor(module): migrate to github.com/ccfos/huatuo") the package tests do not compile with those paths. The rebase applied cleanly and the breakage was only visible when the package was actually built, so I have updated them to `github.com/ccfos/huatuo/...` in this revision. Without that fix this PR would have landed a test file that fails to build.

The earlier QEMU failure on this branch was also in the VM setup, not in the tests: `go install mvdan.cc/sh/v3/cmd/shfmt@latest` now needs Go 1.26 while the VM pins Go 1.24. `bf8c9bfa` pinned shfmt on `main` on 2026-09-10.

```
$ go build ./...                  # exit 0
$ go test ./core/metrics/         # ok  github.com/ccfos/huatuo/core/metrics
$ go vet  ./core/metrics/         # clean
$ gofmt -l core/metrics/iolatency_tracing.go core/metrics/iolatency_tracing_test.go   # no output
```

PR #794, opened 2026-09-08
fix(server): force-close server after shutdown deadline

### LINKED ISSUE #793: [Bug]: Server Shutdown leaves active connections running after deadline
## Problem

`Server.Shutdown(ctx)` in `internal/server/server.go` returns the caller's context error as soon as graceful shutdown exceeds the deadline, but it never force-closes the still-running `http.Server`. The method then marks the server as stopped and clears `activeExecution` even though the serving goroutine and active connections are still alive.

## Evidence

In `Shutdown`, the code calls `execution.shutdown(ctx)` and `execution.wait(ctx)` with the same context. When `ctx` is canceled or times out, both return the context error and the function records `serverStateStopped`. A handler blocked on `ctx.Request().Context().Done()` remains blocked because `http.Server.Shutdown` does not close active connections when the deadline is reached. The old serving goroutine therefore outlives the lifecycle state.

## Expected behavior

When graceful shutdown cannot complete before the context deadline, `Shutdown` should force-close the server (for example with `http.Server.Close`), wait for the serving goroutine to exit, and only then mark the server stopped. The returned error should include the context error, close error, and serve result.

## Test plan

Add a lifecycle regression test in `internal/server`: start a server with a handler that blocks until the request context is canceled, call `Shutdown` with an already-canceled context, and assert the handler is unblocked because the active connection is force-closed. Run `go test -mod=vendor ./internal/server`.

btlqql:
## Summary

Force-close the HTTP server when graceful shutdown exceeds the caller deadline, then wait for the serving goroutine to exit before marking the server stopped.

## Root cause

`Server.Shutdown(ctx)` used the same context for `http.Server.Shutdown` and `serveExecution.wait`. On deadline or cancellation, both returned the context error and the method cleared `activeExecution` and set `serverStateStopped` while active connections and the serving goroutine were still alive.

## Validation

- `gofmt -w internal/server/server.go internal/server/server_lifecycle_test.go`
- `git diff --check`
- `go test -mod=vendor ./internal/server`
- `go test -mod=vendor ./internal/server -run '^TestServerShutdownForceClosesActiveConnectionsOnDeadline$' -count=1 -v`
- `go test -race` was not available on this Windows host because race requires cgo.

Fixes #793

### DIFF
--- a/internal/server/server.go
+++ b/internal/server/server.go
@@ -172,10 +172,17 @@ func (s *Server) Shutdown(ctx context.Context) error {
 	s.mu.Unlock()
 
 	shutdownErr := execution.shutdown(ctx)
-	serveResult := execution.wait(ctx)
 	if shutdownErr != nil && ctx.Err() != nil {
+		// The graceful drain ran out of time. Force the remaining connections
+		// closed and wait for the serving goroutine without a deadline, so the
+		// server cannot stay in the stopping state with a live execution.
+		shutdownErr = errors.Join(shutdownErr, execution.httpServer.Close())
+		serveResult := execution.wait(context.Background())
+
+		s.finishExecution(execution)
 		return errors.Join(shutdownErr, serveResult)
 	}
+	serveResult := execution.wait(ctx)
 
 	s.finishExecution(execution)
 	return errors.Join(shutdownErr, serveResult)
--- a/internal/server/server_lifecycle_test.go
+++ b/internal/server/server_lifecycle_test.go
@@ -167,6 +167,78 @@ func TestServerRejectsOperationsWhileStopping(t *testing.T) {
 	}
 }
 
+func TestServerShutdownForceClosesActiveConnectionsOnDeadline(t *testing.T) {
+	srv := NewServer(nil)
+	handlerStarted := make(chan struct{})
+	releaseHandler := make(chan struct{})
+	handlerDone := make(chan struct{})
+	t.Cleanup(func() {
+		select {
+		case <-releaseHandler:
+		default:
+			close(releaseHandler)
+		}
+	})
+
+	srv.MustRegisterRoutes("", []Route{{
+		Method: http.MethodGet,
+		Path:   "/block",
+		Handler: func(ctx *Context) error {
+			close(handlerStarted)
+			select {
+			case <-releaseHandler:
+			case <-ctx.Request().Context().Done():
+			}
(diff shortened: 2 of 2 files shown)


btlqql:
Rebased onto current `main` (`852d724d`) — the branch is now `b01a5e1c` and the conflict in `internal/server/server.go` is resolved.

`Shutdown` gained a new early return on `main` while this PR was open:

```go
shutdownErr := execution.shutdown(ctx)
serveResult := execution.wait(ctx)
if shutdownErr != nil && ctx.Err() != nil {
	return errors.Join(shutdownErr, serveResult)
}
```

That path returns before `finishExecution`, so a shutdown that runs out of time still leaves the server in `serverStateStopping` with a live `activeExecution` — the same state leak this PR was opened for. The resolution keeps that guard and does the force-close inside it:

```go
shutdownErr := execution.shutdown(ctx)
if shutdownErr != nil && ctx.Err() != nil {
	// The graceful drain ran out of time. Force the remaining connections
	// closed and wait for the serving goroutine without a deadline, so the
	// server cannot stay in the stopping state with a live execution.
	shutdownErr = errors.Join(shutdownErr, execution.httpServer.Close())
	serveResult := execution.wait(context.Background())

	s.finishExecution(execution)
	return errors.Join(shutdownErr, serveResult)
}
serveResult := execution.wait(ctx)
```

Verification on this machine (`internal/server` runs here; the Linux-only packages do not):

```
# red: current main + the new test only
go test -count=1 -run TestServerShutdownForceClosesActiveConnectionsOnDeadline ./internal/server/
--- FAIL: TestServerShutdownForceClosesActiveConnectionsOnDeadline (1.00s)
    server_lifecycle_test.go:228: active connection was not force-closed after Shutdown deadline
FAIL	github.com/ccfos/huatuo/internal/server	1.080s

# green: with the change
go test -count=1 ./internal/server/ ./internal/config/
ok  	github.com/ccfos/huatuo/internal/server	0.114s
ok  	github.com/ccfos/huatuo/internal/config	0.037s
```

`gofmt` is clean. The fresh push should retrigger CI; the previous VM runs on this branch failed on the upstream toolchain fetch, not on the diff (`go: switching to go >= 1.26.0: module golang.org/toolchain: read "https://goproxy.cn/...": stream error: stream ID 59; INTERNAL_ERROR`).

Fixes #793

PR #724, opened 2026-09-05
fix(ci): pin mockery to Go-1.24-compatible version

unbridled-41:
## Problem / Evidence

The VM setup pins `GOLANG_VERSION="1.24.0"`, while its fallback mockery
install used `@latest`. The release Dockerfile has the same mismatch because
it builds from `golang:1.24` and also installed mockery at `@latest`.

`mockery@v2.53.7` requires Go 1.25 or later, so either install path fails when
the tool is not already present. The repository's development Dockerfile and
compiling documentation already use `v2.53.6`, which is compatible with Go
1.24.

Closes #721.

## Fix

Pin mockery to `v2.53.6` in both Go 1.24 environments:

- `.github/workflows/scripts/qemu-2-run-in-vm.sh`
- `Dockerfile`

Per maintainer review, shfmt remains at `@latest`; this PR no longer changes
it. The final diff is limited to the two mockery install lines.

## Verification

- `bash -n .github/workflows/scripts/qemu-2-run-in-vm.sh`: pass.
- Go 1.24 container install and `mockery --version`: pass, reports `v2.53.6`.
- `make check` in the repository development toolchain image: pass (mock
  generation, formatting, and golangci-lint).
- `docker build --network=host --target build -f Dockerfile .`: pass, including
  mock generation and all project binaries.
- GitHub CI: all required checks pass, including Build/Lint/Vendor and the
  Ubuntu 20.04, 22.04, 24.04, and 26.04 QEMU jobs. The Ubuntu 22.04 job passed
  on rerun after an unrelated physical-memory sampling mismatch in its first
  run.

## Risk

Low. The change only replaces two unbounded mockery installs with the same
version already used elsewhere in the repository. No Go source or generated
file changes.

PR #804, opened 2026-09-08
fix(storage): surface localfile directory stat errors

### LINKED ISSUE #802: [Bug]: localfile storage ignores non-ENOENT directory stat errors
**Problem**
`Storage.writerByName` calls `os.Stat` on the configured directory but only handles `os.IsNotExist`. Other stat errors (permission denied, invalid path, ENOTDIR) are silently discarded, and the method returns a lazy writer with a nil error. The actual failure is deferred until the first `Write`, or may be reported in a confusing place.

**Evidence**
`internal/storage/localfile/localfile.go`:
```go
if _, err := os.Stat(b.path); os.IsNotExist(err) {
    if mkdirErr := os.MkdirAll(b.path, 0o755); mkdirErr != nil {
        return nil, mkdirErr
    }
}
```
A non-ENOENT error reaches `newFileWriter` with no validation, so `writerByName` returns `(writer, nil)`.

**Expected behavior**
`writerByName` should return any non-ENOENT `os.Stat` error immediately, just as it already does for `MkdirAll` failures.

**Test plan**
- Add a regression test in `internal/storage/localfile` that creates a backend path whose stat returns a non-ENOENT error and asserts `writerByName` returns a non-nil error.
- Run `gofmt`, `git diff --check`, and `go test -mod=vendor ./internal/storage/localfile`.

yyqdbngt:
## Summary
Return non-ENOENT `os.Stat` errors from `localfile.Storage.writerByName`
instead of silently continuing to create a lazy file writer. The deferred
failure made permission-denied and invalid directory paths hard to diagnose.

## Evidence
`writerByName` only handled `os.IsNotExist`. Any other directory stat error
was discarded, so the method returned `(writer, nil)`.

## Regression test
Added `TestWriterByNameReturnsStatError`, which uses a path producing a
non-ENOENT stat error and asserts `writerByName` returns an error and no
writer. Verified failing before the fix and passing after.

## Testing
- `gofmt -w internal/storage/localfile/localfile.go internal/storage/localfile/localfile_test.go`
- `git diff --check`
- `go test -mod=vendor ./internal/storage/localfile -run TestWriterByNameReturnsStatError -count=1`

Fixes #802

### DIFF
--- a/internal/storage/localfile/localfile.go
+++ b/internal/storage/localfile/localfile.go
@@ -146,7 +146,10 @@ func (b *Storage) writerByName(name string) (io.Writer, error) {
 		return fileWriter, nil
 	}
 
-	if _, err := os.Stat(b.path); os.IsNotExist(err) {
+	if _, err := os.Stat(b.path); err != nil {
+		if !os.IsNotExist(err) {
+			return nil, err
+		}
 		if mkdirErr := os.MkdirAll(b.path, 0o755); mkdirErr != nil {
 			return nil, mkdirErr
 		}
--- a/internal/storage/localfile/localfile_test.go
+++ b/internal/storage/localfile/localfile_test.go
@@ -146,3 +146,22 @@ func TestBackendUnsupportedOperations(t *testing.T) {
 		t.Errorf("Backend.Terms() error = %v, want ErrUnsupported", err)
 	}
 }
+
+// TestWriterByNameReturnsStatError verifies that writerByName surfaces
+// non-ENOENT os.Stat errors instead of silently proceeding to create a lazy
+// file writer.
+func TestWriterByNameReturnsStatError(t *testing.T) {
+	invalidPath := string([]byte{'b', 'a', 'd', 0, 'p', 'a', 't', 'h'})
+	if _, statErr := os.Stat(invalidPath); statErr == nil || os.IsNotExist(statErr) {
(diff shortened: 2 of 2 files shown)


yyqdbngt:
Rebased onto the current `main` (`852d724d`); this branch was 126 commits behind.

Two things came out of doing that:

**The QEMU failure in the previous run was not this change.** The job died at
`.github/workflows/scripts/qemu-2-run-in-vm.sh` while preparing the VM, not in the tests:

```
+ which shfmt
+ go install mvdan.cc/sh/v3/cmd/shfmt@latest
go: mvdan.cc/sh/v3@v3.14.1 requires go >= 1.26.0 (running go 1.24.0)
go: switching to go >= 1.26.0: module golang.org/toolchain: read "https://goproxy.cn/...": stream error
##[error]Process completed with exit code 1.
```

`bf8c9bfa` ("fix(ci): pin shfmt version in QEMU VM") addressed that on `main` on 2026-09-10, and `main` had already pinned mockery in #724, so the rebase is what makes the checks meaningful again.

```
$ go build ./...                          # exit 0
$ go test ./internal/storage/localfile/   # ok  github.com/ccfos/huatuo/internal/storage/localfile
$ go vet   ./internal/storage/localfile/  # clean
$ gofmt -l internal/storage/localfile/localfile.go internal/storage/localfile/localfile_test.go   # no output
```

PR #807, opened 2026-09-08
fix(config): reject out-of-range Elasticsearch ports

### LINKED ISSUE #801: [Bug]: Elasticsearch config accepts out-of-range URL ports
## Problem

`ElasticsearchConfig.Validate` accepts an address whose port is outside the valid TCP port range. `http://127.0.0.1:99999` passes validation, but the address can never be dialed.

## Evidence

The validation loop calls `url.Parse` and checks only for a parse error, a non-empty host, and an http/https scheme. `net/url.Parse` does not reject an out-of-range numeric port. As a result, an invalid address is accepted as a complete Elasticsearch connection and later fails at connection time.

## Expected behavior

Validate each comma-separated address with `url.Parse` plus `net.SplitHostPort`/`strconv.Atoi` to ensure an explicit port is numeric and between 1 and 65535. IPv6 hosts without an explicit port should remain accepted when valid for HTTP/HTTPS.

## Test plan

Add regression cases in `internal/config/elasticsearch_test.go` for `http://127.0.0.1:99999` and a non-numeric port. Run `go test -mod=vendor ./internal/config`.

yyqdbngt:
`ElasticsearchConfig.Validate` previously accepted numeric ports outside the valid TCP range because `url.Parse` does not range-check them. Validate explicit ports with `strconv.Atoi` and reject values outside 1-65535. IPv6 hosts without an explicit port remain accepted.

Fixes #801

Signed-off-by: yyqdbngt <btlqql@gmail.com>

### DIFF
--- a/internal/config/elasticsearch.go
+++ b/internal/config/elasticsearch.go
@@ -18,6 +18,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
+	"strconv"
 	"strings"
 )
 
@@ -62,6 +63,12 @@ func (c ElasticsearchConfig) Validate() error {
 		if err != nil || parsed.Host == "" ||
 			(parsed.Scheme != "http" && parsed.Scheme != "https") {
 			return fmt.Errorf("invalid address %q", address)
+		}
+		if port := parsed.Port(); port != "" {
+			portNumber, err := strconv.Atoi(port)
+			if err != nil || portNumber < 1 || portNumber > 65535 {
+				return fmt.Errorf("invalid address %q", address)
+			}
 		}
 	}
 	return nil
--- a/internal/config/elasticsearch_test.go
+++ b/internal/config/elasticsearch_test.go
@@ -75,6 +75,38 @@ func TestElasticsearchConfigValidate(t *testing.T) {
 			wantEnabled: true,
 			wantError:   `invalid address "127.0.0.1:9200"`,
 		},
+		{
+			name: "out of range port",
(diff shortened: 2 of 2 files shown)


yyqdbngt:
Rebased onto the current `main` (`852d724d`); this branch was 126 commits behind, so the previous green run was from an older base.

The rebase itself was clean, but the checkout caught a real problem in the test cases added here: `ElasticsearchConfig.Validate` requires `Index` to be set when the connection fields are present, so the three new cases failed on `index is required` before ever reaching the port check:

```
--- FAIL: TestElasticsearchConfigValidate/out_of_range_port
--- FAIL: TestElasticsearchConfigValidate/non-numeric_port
--- FAIL: TestElasticsearchConfigValidate/ipv6_without_explicit_port
```

`Index` is now set in those three cases so they actually exercise the port validation.

```
$ gofmt -l internal/config/elasticsearch.go internal/config/elasticsearch_test.go   # no output
$ go build ./...                                                                    # exit 0
$ go test -count=1 ./internal/config/
ok  	github.com/ccfos/huatuo/internal/config
$ go test -count=1 ./internal/config/ -run TestElasticsearchConfigValidate -v
    --- PASS: TestElasticsearchConfigValidate/out_of_range_port (0.00s)
    --- PASS: TestElasticsearchConfigValidate/non-numeric_port (0.00s)
    --- PASS: TestElasticsearchConfigValidate/ipv6_without_explicit_port (0.00s)
$ go vet ./internal/config/                                                          # clean
```

PR #826, opened 2026-09-11
fix(metrics): retain cpu_stat series on sub-second scrapes

### LINKED ISSUE #825: [Bug]: cpu_stat drops every series when a scrape lands inside the 1s refresh window
### 这个问题是否已经存在？

### 部署环境

- HUATUO commit: `852d724d` (current `upstream/main`)
- Go: go1.24 linux/amd64
- Module: `core/metrics`

### 如何复现

1. Enable the `cpu_stat` collector with a cgroup that exposes `wait_sum` and throttle/burst counters.
2. Scrape `/metrics` twice with less than one second between scrapes
   (or use `scrape_interval: 1s`).
3. Observe that the second scrape omits every `cpu_stat` series for that
   container (`wait_sum_percent`, `nr_throttled`, `throttled_time`,
   `nr_bursts`, `burst_time`).

### 根因

`cpuStatCollector.updateDataCache` rate-limits cgroup reads to one per
second so `wait_sum` deltas have a meaningful window:

```go
if now.Sub(cpu.lastUpdate) < time.Second {
    return availability, nil // zero-value availability
}
```

`Update()` then emits each series only when the corresponding
`cpuStatAvailability` flag is true. Returning a zero-value availability
therefore suppresses the entire container for that scrape, even though
the previously computed values are still in the cache.

`cpu_util` already keeps and re-exports cached utilization on the same
sub-second guard; `cpu_stat` does not.

### 修复建议或方案

Persist the last successful `cpuStatAvailability` on the `cpuStat` life
resource and return it from the rate-limit path instead of a zero value.
A unit test can drive `updateDataCache` with a stub cgroup: first sample
stores availability, an immediate second call must retain it.

Acceptance:

```bash
go test ./core/metrics -run TestCPUStatUpdateDataCacheRetainsAvailabilityOnSubSecondScrape -count=1
```

### 其他

No open PR currently addresses this path. Related open issues #536 and
#466 cover uint64 underflow on counter reset (already guarded on main),
not the availability retention gap.

89799969:
## Summary

`cpuStatCollector.updateDataCache` rate-limits cgroup reads to one per second so `wait_sum` deltas have a meaningful window. The rate-limit path returned a zero-value `cpuStatAvailability`, so `Update()` skipped every `cpu_stat` series for that container. A scrape landing inside the window made `wait_sum_percent`, `nr_throttled`, `throttled_time`, `nr_bursts` and `burst_time` vanish even though the cached values were still valid.

This stores the last successful availability on the `cpuStat` life resource and returns it from the rate-limit path, matching `cpu_util` which already re-exports cached utilization under the same guard.

Fixes #825

## Changes

- `core/metrics/cpu_stat.go`: persist `cpuStatAvailability` on the cache and return it when the 1s window has not elapsed
- `core/metrics/cpu_stat_test.go`: stub cgroup + regression test that a sub-second call retains the previous availability

## Validation

- `gofmt -l` clean on both files; `git diff --check` clean
- `go test ./core/metrics -run TestCPUStatUpdateDataCacheRetainsAvailabilityOnSubSecondScrape` could not run in this Windows checkout: the package setup fails on pre-existing vendored build constraints (`internal/bpf/abi` with `-mod=vendor`, and `prometheus/procfs/sysfs`). The added test targets the pure cache path and is kept for Linux CI.
- `make check` / `make unit` not run for the same environment limitation.

## Notes for reviewers

DCO Signed-off-by present. No generated files touched.

### DIFF
--- a/core/metrics/cpu_stat.go
+++ b/core/metrics/cpu_stat.go
@@ -115,20 +115,20 @@ func newCPUStatSample(raw map[string]uint64, cpuTotal uint64, now time.Time) (cp
 	burstTime, burstTimeOK := durationNanoseconds(raw, "burst_time", "burst_usec")
 
 	return cpuStat{
-		nrThrottled:   nrThrottled,
-		throttledTime: throttledTime,
-		waitSum:       waitSum,
-		nrBursts:      nrBursts,
-		burstTime:     burstTime,
-		cpuTotal:      cpuTotal,
-		lastUpdate:    now,
-	}, cpuStatAvailability{
-		waitPercent:   waitSumOK,
-		nrThrottled:   nrThrottledOK,
-		throttledTime: throttledTimeOK,
-		nrBursts:      nrBurstsOK,
-		burstTime:     burstTimeOK,
-	}
+			nrThrottled:   nrThrottled,
+			throttledTime: throttledTime,
+			waitSum:       waitSum,
+			nrBursts:      nrBursts,
+			burstTime:     burstTime,
+			cpuTotal:      cpuTotal,
+			lastUpdate:    now,
+		}, cpuStatAvailability{
+			waitPercent:   waitSumOK,
+			nrThrottled:   nrThrottledOK,
+			throttledTime: throttledTimeOK,
+			nrBursts:      nrBurstsOK,
+			burstTime:     burstTimeOK,
+		}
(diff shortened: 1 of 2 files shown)


89799969:

PR #840, opened 2026-09-14
test(job): cover JobQuery.ID storage equality filter

### LINKED ISSUE #772: [Bug]: job List queries ignore the JobQuery.ID filter

89799969:
Fixes #772

## Problem

`buildStorageQuery` already translates a non-empty `JobQuery.ID` into an
`id` equality filter, but no test pinned that behavior. A regression would
silently widen List results to every row matching the remaining filters.

## Change

Test-only. No production code changes.

- `TestBuildStorageQueryIncludesIDFilter`: asserts a non-empty ID produces
  an `id eq` filter and an empty ID does not.
- `TestStorageStoreListFiltersByID`: List round-trip returns only the
  requested job when two jobs share the other filters.

## Validation

Windows cannot build this package (`internal/timeutil` needs
`unix.ClockGettime`). Ran:

- `GOOS=linux go test -c ./internal/job/` — compile OK
- `go vet ./internal/job/` — clean

Full `go test ./internal/job/...` is left to Linux CI.


89799969:

PR #841, opened 2026-09-14
fix(job): reject lone descending sort prefix

### LINKED ISSUE #771: [Bug]: Job List queries accept a lone "-" sort and fail only in the storage backend

89799969:
Fixes #771

## Problem

`querySort` strips a leading `-` from `JobQuery.Sort`. The lone value `-`
becomes an empty field. Validation previously rejected it only as
`unsupported sort field ""` via a map miss, which is easy to weaken if the
field set changes and is not an explicit contract.

## Change

Reject the empty-after-strip field explicitly with `ErrInvalidQuery` and
add a regression test covering both `validateQuerySort` and
`buildStorageQuery`.

## Validation

Windows cannot build this package (`internal/timeutil` needs
`unix.ClockGettime`). Ran:

- `GOOS=linux go test -c ./internal/job/` — compile OK
- `go vet ./internal/job/` — clean

Full `go test ./internal/job/...` is left to Linux CI.


89799969:

PR #846, opened 2026-09-15
fix(toolstream): finish expected session when client disconnects

### LINKED ISSUE #845: [Bug]: toolstream expected session leaks when the client disconnects without an end frame
## Problem

`transport.Server.handleConn` returns on EOF/decode error without notifying the session layer. `toolstream.Server.dispatch` only calls `finishSession` when `chunk.End` is true.

If a client process is SIGKILLed, crashes, or the socket drops before the end frame:

1. `expectedSession.done` is never closed;
2. `AwaitSession` blocks until context timeout instead of failing immediately;
3. the expectation entry stays in `sessions`, so a later `ExpectSession` for the same tool/task fails with "already expected".

`DocumentWriter` and the profiling pipeline rely on `ExpectSession`/`AwaitSession` pairing, so this is a contract gap rather than dead code.

## Reproduction

1. `ExpectSession("profiler", "job-1")`
2. Connect a client, send chunks, then close the socket without `End()`
3. `AwaitSession` hangs until ctx deadline

## Expected

Disconnect without an end frame should fail the expected session immediately with a clear error and free the expectation slot.

89799969:
## What

Finish an expected toolstream session when the transport connection ends without the protocol end frame.

## Why

`transport.handleConn` returned on EOF/decode errors without notifying the session layer. AwaitSession then blocked until ctx timeout and the expectation entry leaked, so a later ExpectSession for the same tool/task failed.

Fixes #845

## How

- Transport tracks whether the end frame was observed.
- `ServeWithDisconnect` invokes an optional callback after the connection ends.
- toolstream records `connection closed before end frame` and finishes the session when the end frame was missing.

## Tests

- `TestOnTransportDisconnectWithoutEndCompletesExpectedSession`
- `TestOnTransportDisconnectWithEndDoesNotFailSession`

`go test ./internal/toolstream/... -race` (Linux CI; Windows cannot build some vendor packages).


89799969:

PR #854, opened 2026-09-16
fix(server): map oversized request bodies to HTTP 413

### LINKED ISSUE #659: [Bug]: Oversized API request bodies are reported as HTTP 400
### 这个问题是否已经存在？

- [x] 我已经搜索过现有的问题

I searched open and closed Issues and pull requests using `request body too
large`, `oversized body`, `payload too large`, `MaxBytesReader`, `MaxBodyBytes`,
`HTTP 413`, `invalid_request`, and related endpoint terms. I also inspected the
current server and handler PRs. None covers preserving the body-size failure as
HTTP 413 across the JSON and protobuf APIs.

### 部署环境

```text
HUATUO commit: a1e74a8ef58499c98368e61fa892f168e989bb87
Branch tested: upstream/main
Architecture: linux/amd64
Go: 1.24.13
Test environment: WSL2 Linux on Windows
Affected components:
  internal/server/middleware.go
  internal/server/response
  cmd/huatuo-bamai/handlers
  cmd/huatuo-apiserver/handlers
```

#### Production call path

Every HUATUO HTTP server installs the body limit before routing:

```text
server.NewServer
  -> buildMiddlewareChain
  -> maxBodyBytesMiddleware
  -> http.MaxBytesReader
  -> route handler
  -> Context.ShouldBindJSON / Context.ShouldBindBodyWith
```

`http.MaxBytesReader` correctly stops reading after `Config.MaxBodyBytes` and
the binding operation returns an `*http.MaxBytesError`. The handlers then erase
that classification by converting every binding error to the generic
`invalid_request` response with HTTP 400.

Examples include:

```text
huatuo-bamai:
  POST /tasks
  PUT  /config
  POST /v1/events/watch

huatuo-apiserver:
  profiling create/patch endpoints
  trace create/patch endpoints
  protobuf profile query endpoints
```

#### Deterministic reproduction

1. Start the real internal HTTP server with `MaxBodyBytes: 64`.
2. Register the production `NewTaskHandler` routes under `/tasks`.
3. Send a syntactically valid task JSON document larger than 64 bytes.
4. Read the status and standard error envelope.

Observed output on the commit above:

```text
=== RUN   TestReproduceOversizedTaskBodyReturnsBadRequest
    body limit was enforced but reported as HTTP 400:
    {"error":{"code":"invalid_request","message":"http: request body too large"}}
--- PASS: TestReproduceOversizedTaskBodyReturnsBadRequest (0.04s)
PASS
```

The body was rejected before task creation, so the size guard itself works. The
incorrect behavior is the loss of the typed `*http.MaxBytesError` when the
handler constructs its response.

#### Actual result

```text
HTTP status: 400 Bad Request
error code: invalid_request
message: http: request body too large
```

A client cannot distinguish a malformed JSON/protobuf document from a request
that is valid but exceeds the configured server limit. API metrics also group
both outcomes under 400.

#### Expected result

```text
HTTP status: 413 Request Entity Too Large
error code: invalid_request
message: request body exceeds <configured-limit> bytes
```

Requirements:

- classify `*http.MaxBytesError` through wrapping with `errors.As`;
- preserve the standard HUATUO JSON error envelope;
- report the actual configured limit from the error;
- apply the classification consistently to JSON and protobuf endpoints in both
  daemons;
- retain current 400 behavior for malformed syntax, schema validation, and
  semantic request errors;
- work for both known `Content-Length` and streamed/chunked request bodies.

### 修复建议或方案

1. Add one response-layer binding-error classifier rather than duplicating
   `MaxBytesError` checks in every handler.
2. Return an `APIError` with HTTP 413 and `invalid_request` when the error chain
   contains `*http.MaxBytesError`.
3. Retain endpoint-specific fallback messages for non-size errors so the repair
   does not broaden the API compatibility surface.
4. Route every `ShouldBindJSON` and `ShouldBindBodyWith` failure through the
   classifier.
5. Add response-unit tests for direct and wrapped size errors, validator errors,
   fallback messages, and ordinary decode errors.
6. Add real HTTP regression coverage using a production route for both fixed
   `Content-Length` and chunked bodies, plus an under-limit malformed request
   control.

#### Root cause

The middleware retains the important error type:

```go
ctx.Request.Body = http.MaxBytesReader(ctx.Writer, ctx.Request.Body, limit)
```

But handlers currently flatten it, for example:

```go
if err := ctx.ShouldBindJSON(&req); err != nil {
    return response.ErrInvalidRequest.WithMessage(err.Error())
}
```

or write HTTP 400 directly. Once converted to `ErrInvalidRequest`, the central
response writer has no remaining evidence that the failure was a body-size
limit.

#### Impact

- clients may keep retrying an unchanged oversized request because 400 suggests
  malformed input rather than a configured size boundary;
- clients cannot adapt payload size from a stable, limit-bearing error;
- access logs and HTTP metrics misclassify capacity/guardrail rejections as
  syntax errors;
- JSON and protobuf endpoints do not expose a consistent HTTP contract even
  though they share the same body-limiting middleware.

The request is still bounded, so this is not a bypass of the memory guard. It is
a caller-visible protocol and observability defect on a normal production path.

#### Compatibility

No successful request behavior, body-size configuration, route, or response
envelope needs to change. Only requests that already fail because they exceed
`MaxBodyBytes` should move from HTTP 400 to HTTP 413. Other binding failures
should retain their existing status and endpoint-specific message behavior.

#### Acceptance criteria

- direct and wrapped `*http.MaxBytesError` values map to HTTP 413;
- the response uses code `invalid_request` and includes the byte limit;
- oversized JSON reaches no business operation and returns 413;
- oversized protobuf returns 413 instead of the generic protobuf 400;
- known-length and chunked bodies behave identically;
- malformed and validation-failing requests below the limit remain HTTP 400;
- all production body-binding call sites use the shared classifier;
- focused tests, race tests, vet, package lint, full lint, and `make check`
  pass.

89799969:

- Fixes #659

### Brief Description

`http.MaxBytesReader` correctly stops reading after `MaxBodyBytes`, but handlers flattened the typed `*http.MaxBytesError` into a generic `invalid_request` HTTP 400. Clients could not distinguish a malformed document from a valid request that exceeded the configured limit.

- Add `response.ClassifyBindingError` to map (including wrapped) `*http.MaxBytesError` to `ErrRequestTooLarge` with the byte limit in the message.
- Use it in the OpenAPI request validator and the protobuf profile-query reader.
- Malformed syntax, schema validation, and semantic failures keep HTTP 400.

Windows cannot build `internal/server` (`golang.org/x/unix`, `procfs`). Full `make check` / `make unit` left to Linux CI.

Local verification on this host:

```text
go test ./internal/server/response/ -count=1 -run TestClassifyBindingError
ok   github.com/ccfos/huatuo/internal/server/response
```

- `gofmt -l` clean on touched files
- `git diff --check` clean

Updated `openapi_test.go` expected message for `MaxBytesError{Limit: 10}` to `request body exceeds 10 bytes`.

Not run on this host: `make check`, `make unit`, race detector (Linux-only packages).

### Checklist

- [x] One logical change
- [x] Focused unit tests for the classifier


89799969:

PR #855, opened 2026-09-16
fix(storage): reject a nil backend configuration

### LINKED ISSUE #605: Storage backend construction panics on a nil configuration

89799969:

- Fixes #605

### Brief Description

`driver.NewBackend` dereferenced `cfg.Driver` before validating the configuration pointer. A nil backend config panicked instead of returning the same style of configuration error used for empty and unregistered drivers.

- Return a clear error when `cfg` is nil.
- Extend `TestNewBackend` with a nil-config table case (table now uses `*Config`).

```text
go test ./internal/storage/driver/ -count=1 -run TestNewBackend
ok  	github.com/ccfos/huatuo/internal/storage/driver
```

Local checks:

- `gofmt -l` clean on touched files
- `git diff --check` clean

Not run on this host: full `make check` / `make unit` (Linux packages).

### Checklist

- [x] One logical change
- [x] Focused unit test updated


89799969:

PR #856, opened 2026-09-16
fix(ascend): roll back the DCMI library reference on init failure

### LINKED ISSUE #624: [Bug]: Failed DCMI initialization leaks the library reference

89799969:

- Fixes #624

### Brief Description

`DcInit` called `libdcmi.load()` (which opens `libdcmi.so` and increments the reference count) and then `dcmi_init`. A native init failure returned immediately without releasing that reference, so the library stayed open with a positive count. A later retry took the already-loaded path, and shutdown no longer matched successful init calls.

On init failure, close the library and join a close error with the original init error.

This package depends on `purego`/`dl` and cannot be built on Windows (`undefined: purego.Dlopen`). Verification is left to Linux CI / an Ascend environment.

Local checks on this host:

- `gofmt -l` clean on `core/metrics/ascend/dcmi/init.go`
- `git diff --check` clean

Not run on this host: `go build ./core/metrics/ascend/dcmi/`, `make check`, `make unit`.

### Checklist

- [x] One logical change


89799969:

PR #862, opened 2026-09-17
fix(bpf): close rate-limit reader to nil and stop ethertype tracking at LoadMemShift

enkilee:
## Summary

Fix two P1 bugs found during static analysis of recent commits:

1. **`CloseEventPipe` doesn't nil out `r.reader`** — after closing
   the perf reader, the field retained the stale reference, blocking
   reopen with a spurious `errRateLimitEventPipeAlreadyOpen`.

2. **`findEthertypeComparisons` continues past `LoadMemShift`** —
   `LoadMemShift` overwrites register A (`A = P[off:X]`), so subsequent
   `JumpIf` instructions are not ethertype comparisons. Grouping it with
   A-preserving instructions (`StoreScratch`, `TAX`) caused false-positive
   ethertype marking and incorrect value rewriting in `retargetToL3`.

## Changes

- `internal/bpf/bpf_ratelimit.go`: set `r.reader = nil` before returning
  from `CloseEventPipe`, regardless of `Close()` error.
- `internal/pcapfilter/bpf_filter.go`: remove `cbpf.LoadMemShift` from
  the branch-follow case in `findEthertypeComparisons`.

## Why

- **Bug 1**: The rate limiter cannot be reopened after close in tests
  or hot-reload paths. Latent in production (single open/close cycle)
  but blocks test reuse and future restart scenarios.
- **Bug 2**: Complex pcap filters that use `LoadMemShift` (e.g.
  `tcp[tcpflags]`) between an ethertype load and a `JumpIf` with value
  `0x0800` or `0x86dd` get that value silently rewritten to `0x40` or
  `0x60`, producing an incorrect L3 filter that matches wrong packets.


internal/bpf/bpf_ratelimit.go
```diff
@@ -131,7 +131,9 @@ func (r *RateLimiter) CloseEventPipe() error {
 	if r.reader == nil {
 		return nil
 	}
-	if err := r.reader.Close(); err != nil {
+	err := r.reader.Close()
+	r.reader = nil
+	if err != nil {
 		return fmt.Errorf("%s: close rate-limit event pipe: %w", r.name, err)
 	}
```
hao022, line 136: 应该只有在成功执行close之后，再设置为nil

internal/pcapfilter/bpf_filter.go
```diff
@@ -465,7 +465,7 @@ func findEthertypeComparisons(insns []cbpf.Instruction) []bool {
 			if ins.Dst == cbpf.RegX {
 				pending = append(pending, index+1)
 			}
-		case cbpf.LoadMemShift, cbpf.StoreScratch, cbpf.TAX:
```
hao022, line 468: 修改这里的目的是什么？这个问题和上面的修改应该是两个独立的问题，如需修改需要分别提交，并说明现在的问题影响。

internal/bpf/bpf_ratelimit.go
enkilee: roger

internal/pcapfilter/bpf_filter.go
enkilee: already revert.

PR #891, opened 2026-09-20
fix(memsnap): recognize deleted runtime executables

### LINKED ISSUE #890: [Bug]: memsnap stops recognizing Python after its running executable is unlinked
### 这个问题是否已经存在？

### 部署环境

- Upstream commit: `5da3ac28aa4e2649c381f2d8bc10ef26e4b6f66a`.
- Ubuntu 22.04.5 LTS, WSL2, x86_64.
- Kernel: `6.6.114.1-microsoft-standard-WSL2`.
- Go: `go1.26.7 linux/amd64`, vendored dependencies.
- Interpreter: a temporary copy of `/usr/bin/python3.10`, which does not link to a shared `libpython3` library in this environment.
- The executable copy is created on Linux ext4 under `/tmp` and removed by the test. Do not use a Windows-mounted directory for this reproduction: reopening an unlinked executable behaves differently there.

### 如何复现

DetectLanguage` identifies the executable using the basename of `/proc/<pid>/exe`. When a still-running executable is unlinked, this link gains a ` (deleted)` suffix. `languageFromExecutable` passes that suffix into the Python executable-name regular expression and no longer recognizes the interpreter.

The fallback does not recover for the tested Ubuntu Python executable because it has no shared `libpython3` dependency. The same running process changes from `python` to `unknown` without a detection error. This prevents the collector from selecting its Python snapshot provider.

1. Save the attached test as `internal/memsnap/audit_deleted_python_test.go`.
2. On a Linux host with `/usr/bin/python3.10`, run:

```sh
go test -mod=vendor ./internal/memsnap -run '^TestAuditDeletedPythonExecutable$' -count=3 -v
```

The test starts an isolated Python interpreter, waits until it is ready, checks `DetectLanguage`, unlinks only the temporary executable copy, and checks the same process again. It terminates the child and removes its temporary directory on completion.

Observed in all three runs:

```text
before="python" after="unknown" error=<nil>
```

The process remains alive, and `/proc/<pid>/exe` points to `python3.10 (deleted)`.

Expected: the same live Python interpreter is still classified as `python` after unlinking its executable. This matters for long-running processes whose executable is replaced during an update.

[deleted_python_repro_test.go.txt](https://github.com/user-attachments/files/32437927/deleted_python_repro_test.go.txt)

### 修复建议或方案

Remove the procfs ` (deleted)` suffix before matching the executable basename. Keep the current runtime-name matching rules. Add regression tests for normal/deleted Python and Java names and unrelated executable names.

A focused fix and tests are ready on my fork, and I will submit the PR immediately after this issue.

### 其他

This is a real-process language-detection reproduction, not an end-to-end OOM snapshot test. Java basename matching has the same suffix issue, but its mapped-library fallback may still detect Java; the demonstrated end-to-end detection failure is Python.

SomaFon:
Fixes #890

## Summary

Strip the procfs ` (deleted)` suffix before matching runtime executable names.

A live Python process can stop being recognized after its executable is unlinked. The basename no longer matches the Python name pattern, and the dependency fallback cannot recover when the interpreter does not link to a shared libpython. The collector then receives `unknown` instead of selecting the Python provider.

This keeps the existing name matching rules and adds regression coverage for normal/deleted Python and Java names and unrelated executables.

## Tests

- The new executable-name test fails on the original code and passes after the fix.
- A real Python 3.10 process changes from `python` to `unknown` before the fix; after the fix it remains `python`. Repeated three times on ext4. The standalone reproduction is supplied separately, not committed as a Python-dependent unit test.
- `go test -mod=vendor -p 2 ./internal/memsnap/...` — passed.
- `git diff --check` — passed.

`make check APP_COMMIT=5da3ac2-dirty` was attempted but failed in the vendor prerequisite because module downloads were disabled (`GOPROXY=off`) in the isolated environment. Full repository checks and end-to-end OOM capture were not run.


internal/memsnap/language.go
```diff
@@ -120,6 +120,7 @@ func detectLanguage(ctx context.Context, exePath, mapsPath string) (Language, er
 }
 
 func languageFromExecutable(executable string) Language {
+	executable = strings.TrimSuffix(executable, " (deleted)")
```
hao022: 针对这个问题能够实现通用的检查函数名？至少覆盖如下场景，甚至profilter场景；此外，不需要提供单元测试。
internal/memsnapshot/language.go:101:           java = java || strings.HasSuffix(strings.TrimSuffix(mapping.Path, " (deleted)"), "/libjvm.so")
internal/memsnapshot/providers/golang/runtime_linux_test.go:26: path := "/bin/heap (deleted)"
internal/memsnapshot/providers/java/runtime_linux.go:107:               path := strings.TrimSuffix(mapping.Path, " (deleted)")
internal/memsnapshot/providers/python/runtime_linux.go:233:     executableTarget = strings.TrimSuffix(executableTarget, " (deleted)")
internal/memsnapshot/providers/python/runtime_linux.go:238:             path := strings.TrimSuffix(mapping.Path, " (deleted)")
internal/memsnapshot/providers/python/runtime_linux.go:290:             if !strings.HasSuffix(mapping.Path, " (deleted)") ||


internal/profiler/process/container.go:133:             resolvedExecutable = strings.TrimSuffix(resolvedExecutable, " (deleted)")
internal/profiler/process/container_test.go:95: writeTestProcess(t, root, testProcess{pid: 100, executable: "/usr/bin/java (deleted)", ppid: 1})

SomaFon:
@hao022 已按建议处理：分支已更新到当前 `main`，新增 `internal/procfs.TrimDeletedSuffix`，并统一用于语言识别、JVM/Python runtime module 处理以及 profiler 进程筛选，覆盖你列出的重复场景，避免各处单独处理 ` (deleted)`。按建议未新增单元测试。已验证 `internal/procfs`、`internal/memsnapshot` 及 Java/Python provider 测试通过；`internal/profiler/process` 在本地仍受仓库缺失 `internal/bpf/abi` 生成包影响，交由 CI 验证。

internal/memsnap/language.go
SomaFon: 已按建议处理并更新分支：新增 `internal/procfs.TrimDeletedSuffix`，统一用于语言识别、JVM/Python runtime module 处理和 profiler 进程筛选，覆盖评论中列出的重复处理位置。按建议未新增单元测试；相关 memsnapshot、Java/Python provider 测试已通过，profiler/process 仍交由 CI 验证。

PR #923, opened 2026-09-22
refactor(pod): share memory cgroup lifecycle across build tags

sbtger:
Memory threshold snapshots need lifecycle notifications with the internal didi backend as well as the default backend. Move the BPF listener and its dependencies into the existing shared cgroup_lifecycle.go, keeping local CSS cache initialization and updates behind default-only callbacks.

Remove the default CSS initialization test and combined dispatch benchmark from the shared lifecycle test file instead of relocating them. Keep the reader recovery test independent of the optional default CSS cache.


hao022:
@sbtger  Hi, 近期的提交对这块的修改很大，可以根据最新的代码，重新提交pr。

